PR #6, opened 2023-02-14
add falco as the ebpf engine implementation

rcohencyberarmor:
## Overview
falco imp as the ebpf engine

go test -v sniffer/pkg/ebpfeng
> Please provide instructions on how to test the changes made in this pull request
-->


internal/config/v1/config_data.go
```diff
@@ -81,13 +81,10 @@ func (c *ConfigData) IsFalcoEbpfEngine() bool {
...
 	if c.IsFalcoEbpfEngine() {
 		for i := range c.ServicesList {
-			switch v := c.ServicesList[i].Data.(type) {
-			case SnifferData:
-				for j := range v.FeatureList {
-					if v.FeatureList[j].Name == SNIFFER_SERVICE_RELEVANT_CVES {
-						falcoSyscallFilter = append(falcoSyscallFilter, []string{"open", "openat", "execve", "execveat"}...)
-					}
-				}
+			features := c.ServicesList[i].Data.(map[string]interface{})["features"]
```
matthyx, line 84: all these casts are really ugly, and they don't provide compile-time safety
can we do better?

pkg/ebpfeng/falco_sniffer_engine.go
```diff
@@ -0,0 +1,188 @@
...
+}
+
+func convertStringTimeToTimeOBJ(timestamp string) (*time.Time, error) {
+	dateAndTime := strings.Split(timestamp, "T")
+	date := strings.Split(dateAndTime[0], "-")
+	tm := strings.Split(dateAndTime[1], ":")
+
+	year, err := strconv.Atoi(date[0])
+	if err != nil {
+		logger.L().Error("", helpers.String("fail strconv %v", fmt.Sprintf("%v", err)))
```
matthyx, line 109: I think we can use `logger.L().Error("fail strconv", helpers.Error(err))`

matthyx (changes requested):

internal/config/config.go
```diff
@@ -15,18 +14,22 @@ const (
 )
 
 var c *Config
+var myContainerID string
+
+func init() {
+	myContainerID = "1111111"
```
matthyx, line 20: just curious, why do you need something not "" ?

internal/config/config.go
```diff
@@ -15,18 +14,22 @@ const (
...
 }
 
 func (cfg *Config) getConfigFilePath() (string, bool) {
 	return os.LookupEnv("SNIFFER_CONFIG")
 }
 
 func (cfg *Config) GetConfigurationData() (io.Reader, error) {
-	config := Config{}
-	c = &config
+	c = cfg
```
matthyx, line 32: why this global var?

internal/config/config.go
```diff
@@ -64,3 +66,19 @@ func IsFalcoEbpfEngine() bool {
 func GetSyscallFilter() []string {
 	return c.data.GetFalcoSyscallFilter()
 }
+
+func GetFalcoKernelObjPath() string {
```
matthyx, line 70: why a function?

internal/config/config_test.go
```diff
@@ -16,12 +17,27 @@ func TestConfig(t *testing.T) {
 	if err != nil {
 		t.Fatalf("GetConfigurationData failed with err %v", err)
 	}
-	err = cfg.ParseConfiguration(configData)
+	err = cfg.ParseConfiguration(&config.ConfigData{}, configData)
```
matthyx, line 20: would be nicer if ParseConfiguration would return the configdata object, instead of being a parameter... think of more functional programming

matthyx (changes requested):
I think it's much better this way, congrats!
What I don't like is that you're mixing methods (with receiver Config) and functions inside `config.go` so if you could solve that.

internal/config/config.go
rcohencyberarmor: I'll change it, no reason

matthyx (approved):
it looks much better like that, thanks Razi!

PR #7, opened 2023-02-16
accumulator imp

rcohencyberarmor:
## Overview

accumulator implementation - the accumulator is in charge of storing the ebpf data we are getiing from the ebpf engine + to stream ebpf data to register containers


pkg/accumulator/accumulator_test.go
```diff
@@ -0,0 +1,34 @@
+package accumulator
+
```
amirmalka, line 2: Any test fixtures/binaries should be placed under `testdata` directory. Go tooling knows to ignore anything placed under this directory. I'm referring to `kernel_obj_mock.o` & `userspace_app_mock`.
In addition, I don't see where it is being used.

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,221 @@
...
+
+func (acc *ContainerAccumulator) StopContainerAccumulator() {
+	acc.unregisterContainerAccumulator()
+}
+
+func GetCacheAccumulator() *CacheAccumulator {
+	return cacheAccumulator
+}
+
+func (acc *CacheAccumulator) AccumulatorByContainerID(aggregationData *[]evData.EventData, containerID string, containerStartTime interface{}) {
```
amirmalka, line 205: `containerStartTime` is not being used?

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,221 @@
...
+
+type ContainerAccumulator struct {
+	dataChannel chan evData.EventData
+	containerID string
+}
+
+var myContainerID string
+var cacheAccumulator *CacheAccumulator
+
+func CreateAccumulator() *CacheAccumulator {
```
amirmalka, line 44: If I understand correctly, the Accumulator is behaving as a Singleton? 
if so, you could simply have a single `GetAccumulator` method, which returns it if `cacheAccumulator != nil`, otherwise creates it.

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,221 @@
...
+	acc.unregisterContainerAccumulator()
+}
+
+func GetCacheAccumulator() *CacheAccumulator {
+	return cacheAccumulator
+}
+
+func (acc *CacheAccumulator) AccumulatorByContainerID(aggregationData *[]evData.EventData, containerID string, containerStartTime interface{}) {
+	for i := range acc.accumulatorData {
+		logger.L().Debug("", helpers.String("index ", fmt.Sprintf("%d:%v", i, acc.accumulatorData[i])))
```
amirmalka, line 207: what are you logging here?

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+	for i := range acc.accumulatorData {
+		for j := range acc.accumulatorData[i][containerID] {
+			acc.syncReaderWriterAccumulatorData.Lock()
+			*aggregationData = append(*aggregationData, acc.accumulatorData[i][containerID][j])
+			acc.syncReaderWriterAccumulatorData.Unlock()
+		}
+	}
+	logger.L().Debug("data ", helpers.String("", fmt.Sprintf("%v", aggregationData)))
```
amirmalka, line 216: what are you logging here? any context?

pkg/accumulator/accumulator_test.go
```diff
@@ -0,0 +1,34 @@
...
+		t.Fatalf("ParseConfiguration failed with err %v", err)
+	}
+
+	cacheAccumulatorErrorChan := make(chan error)
+	acc := CreateAccumulator()
+	err = acc.StartAccumulator(cacheAccumulatorErrorChan)
+	if err != nil {
+		t.Fatalf("StartAccumulator failed with err %v", err)
+	}
+	time.Sleep(5 * time.Second)
```
amirmalka, line 33: the test ends with a sleep but  what is it checking?

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,221 @@
...
+	errChan <- acc.ebpfEngine.GetEbpfEngineError()
+}
+
+func (acc *CacheAccumulator) StartAccumulator(errChan chan error) error {
+	falcoEbpfEngine := ebpfeng.CreateFalcoEbpfEngine(config.GetConfigurationConfigContext().GetSyscallFilter(), false, false, "")
+	acc.ebpfEngine = falcoEbpfEngine
+
+	err := acc.ebpfEngine.StartEbpfEngine()
+	if err != nil {
+		logger.L().Error("fail to create ebpf engine")
```
amirmalka, line 171: if you are logging the error here, you should also include the original error in the log

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,221 @@
...
+func (acc *CacheAccumulator) getEbpfEngineData() {
+	acc.ebpfEngine.GetData(acc.mainDataChannel)
+}
+
+func (acc *CacheAccumulator) getEbpfEngineError(errChan chan error) {
+	errChan <- acc.ebpfEngine.GetEbpfEngineError()
+}
+
+func (acc *CacheAccumulator) StartAccumulator(errChan chan error) error {
+	falcoEbpfEngine := ebpfeng.CreateFalcoEbpfEngine(config.GetConfigurationConfigContext().GetSyscallFilter(), false, false, "")
```
amirmalka, line 166: are the last two arguments configurable? (why hardcoding `false, false, ""` )

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,221 @@
...
+
+func (acc *CacheAccumulator) accumulateEbpfEngineData() {
+	for {
+		event := <-acc.mainDataChannel
+		logger.L().Debug("metadataAcc ", helpers.String("", fmt.Sprintf("%v", event)))
+		if strings.Contains(event.GetEventContainerID(), myContainerID) {
+			continue
+		}
+		if event != nil {
+			if event.GetEventCMD() == "drop event occurred\n" {
```
amirmalka, line 142: use const for drop event string

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,221 @@
...
+				acc.removeAllStreamedContainers(event)
+			} else {
+				index := acc.findIndexByTimestamp(event)
+				if index == -1 {
+					// logger.L().Debug("metadataAcc %v\n", metadataAcc)
```
amirmalka, line 147: remove comments

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,221 @@
...
+		firstKey := acc.firstMapKeysOfAccumulatorData[i]
+		if event.GetEventTimestamp().Sub((acc.accumulatorData[i])[firstKey][0].GetEventTimestamp()) < time.Second {
+			return i
+		}
+	}
+	index := acc.findIndexByTimestampWhenAccumulatorDataIsFull(event)
+	if index != -1 {
+		return index
+	}
+	// logger.L().Debug("findIndexByTimestamp: failed to find index, sniffer data will not saved")
```
amirmalka, line 104: remove comment

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,221 @@
...
+	m := make(map[string][]evData.EventData)
+	m[event.GetEventContainerID()] = slice
+	acc.accumulatorData[index] = m
+	acc.firstMapKeysOfAccumulatorData[index] = event.GetEventContainerID()
+}
+
+func (acc *CacheAccumulator) findIndexByTimestampWhenAccumulatorDataIsFull(event *evData.EventData) int {
+	index := 0
+	minTimestamp := acc.accumulatorData[0][acc.firstMapKeysOfAccumulatorData[0]][0].GetEventTimestamp()
+	for i := range acc.accumulatorData {
```
amirmalka, line 76: You could simplify by starting your loop at 1.

```suggestion
	for i := 1; i < len(acc.accumulatorData); i++ {
```

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,221 @@
...
+		if len(acc.accumulatorData[i]) == 0 {
+			acc.createNewMap(event, i)
+			return i
+		}
+		firstKey := acc.firstMapKeysOfAccumulatorData[i]
+		if event.GetEventTimestamp().Sub((acc.accumulatorData[i])[firstKey][0].GetEventTimestamp()) < time.Second {
+			return i
+		}
+	}
+	index := acc.findIndexByTimestampWhenAccumulatorDataIsFull(event)
```
amirmalka, line 100: by the way `findIndexByTimestampWhenAccumulatorDataIsFull` there is no possibility to get `index = -1` ...

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,221 @@
...
+	slice := make([]evData.EventData, 0)
+	m := make(map[string][]evData.EventData)
+	m[event.GetEventContainerID()] = slice
+	acc.accumulatorData[index] = m
+	acc.firstMapKeysOfAccumulatorData[index] = event.GetEventContainerID()
+}
+
+func (acc *CacheAccumulator) findIndexByTimestampWhenAccumulatorDataIsFull(event *evData.EventData) int {
+	index := 0
+	minTimestamp := acc.accumulatorData[0][acc.firstMapKeysOfAccumulatorData[0]][0].GetEventTimestamp()
```
amirmalka, line 75: I see multiple index "0" access without any length checks... how can you be sure you access within the bounds.

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,221 @@
...
+		logger.L().Debug("metadataAcc ", helpers.String("", fmt.Sprintf("%v", event)))
+		if strings.Contains(event.GetEventContainerID(), myContainerID) {
+			continue
+		}
+		if event != nil {
+			if event.GetEventCMD() == "drop event occurred\n" {
+				acc.removeAllStreamedContainers(event)
+			} else {
+				index := acc.findIndexByTimestamp(event)
+				if index == -1 {
```
amirmalka, line 146: `findIndexByTimestamp` never returns -1

pkg/accumulator/accumulator_test.go
```diff
@@ -0,0 +1,34 @@
+package accumulator
+
+import (
+	"os"
+	"sniffer/pkg/config"
+	v1 "sniffer/pkg/config/v1"
+	"testing"
+	"time"
+)
+
```
amirmalka, line 10: Please add unit tests for the accumulator functions.

amirmalka (changes requested):

pkg/accumulator/accumulator.go
rcohencyberarmor: all the last arguments is configurable to the ebpf engine based on falco.

pkg/accumulator/accumulator.go
rcohencyberarmor: I saw a benchmark test that there is no different between those 2 options, there is another reason to this change.
https://pierspowlesland.wordpress.com/2016/08/25/golang-slice-iteration-performance/

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+	minTimestamp, err := acc.getFirstTimestamp()
+	if err != nil {
+		logger.L().Warning("findIndexByTimestampWhenAccumulatorDataIsFull fail to find the place to insert the event, fail with error", helpers.Error(err))
+		return -1
+	}
+	for i := range acc.accumulatorData {
+		if i == 0 {
+			continue
+		}
+		if acc.accumulatorData[i][acc.firstMapKeysOfAccumulatorData[i]][0].GetEventTimestamp().Before(minTimestamp) {
```
amirmalka, line 120: don't you need to use the `syncReaderWriterAccumulatorData`?

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	return acc.accumulatorData[0][acc.firstMapKeysOfAccumulatorData[0]][0].GetEventTimestamp(), nil
+}
+
+func (acc *Accumulator) findIndexByTimestampWhenAccumulatorDataIsFull(event *evData.EventData) int {
+	index := 0
+	minTimestamp, err := acc.getFirstTimestamp()
+	if err != nil {
+		logger.L().Warning("findIndexByTimestampWhenAccumulatorDataIsFull fail to find the place to insert the event, fail with error", helpers.Error(err))
+		return -1
```
amirmalka, line 114: by doing so, you are forcing the caller to know that "-1" is an error. 
why not return `(int, error)` if we can get into an error flow?

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+			accumulatorDataPerContainer: make(map[string]chan evData.EventData),
+		},
+	}
+
+	return accumulatorSingelton
+}
+
+func GetAccumulator() *Accumulator {
+	if accumulatorSingelton == nil {
+		lock.Lock()
```
amirmalka, line 66: please have a meaningful name for this ("lock" is just too generic).

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+		defer lock.Unlock()
+		if accumulatorSingelton == nil {
```
amirmalka, line 68: why are you checking twice for nil?

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+			logger.L().Debug("Creating accumulatorSingelton now.")
+			accumulatorSingelton = createAccumulator()
+		} else {
+			logger.L().Debug("accumulatorSingelton already created.")
```
amirmalka, line 72: I don't think these logs are necessary...

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+	dataChannel chan evData.EventData
+	containerID string
+}
+
+var myContainerID string
+var accumulatorSingelton *Accumulator
+var lock = &sync.Mutex{}
+
+func createAccumulator() *Accumulator {
+	myContainerID = "111111"
```
amirmalka, line 50: what is this myContainerID?

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+
+	logger "github.com/kubescape/go-logger"
+	"github.com/kubescape/go-logger/helpers"
+)
+
+const (
+	ACCUMULATOR_SIZE = 10
+)
+
+const (
```
amirmalka, line 21: combine const declaration

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+	DROP_EVENT_OCCURRED = "drop event occurred\n"
+)
+
+type containersAccumulator struct {
+	accumulatorDataPerContainer map[string]chan evData.EventData
```
amirmalka, line 26: the name of this map is not clear -  I assume the key is the containerID?
if so, maybe a better name is containerIdToAccumuationData?  or xxxToYYY ...

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+	accumulatorData                 []map[string][]evData.EventData
+	syncReaderWriterAccumulatorData sync.Mutex
+	firstMapKeysOfAccumulatorData   []string
+	cacheAccumulatorSize            int
+	mainDataChannel                 chan *evData.EventData
+	containersData                  containersAccumulator
+	ebpfEngine                      ebpfeng.EbpfEngineClient
+}
+
+type ContainerAccumulator struct {
```
amirmalka, line 40: It is confusing that we have two types here - one `containersAccumulator` and the other `ContainerAccumulator` ...

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+	dataChannel chan evData.EventData
+	containerID string
+}
+
+var myContainerID string
+var accumulatorSingelton *Accumulator
```
amirmalka, line 46: typo.. should be `Singleton`. 
Anyways, it can be simplified by calling it with some shorter name like `instance` (it's pretty obvious that it's a singleton)

pkg/accumulator/accumulator_interface.go
```diff
@@ -2,7 +2,7 @@ package accumulator
```
amirmalka, line 2: a better naming convention for this file is to call it `interface.go` (it is already under the accumulator directory and package.

pkg/accumulator/accumulator_test.go
amirmalka: ?

pkg/accumulator/accumulator_test.go
amirmalka: please create a `testdata` directory under the `config` directory and move both files there. The convention is that the testdata directory is in the same directory of its consumer.

pkg/config/config.go
```diff
@@ -13,12 +13,16 @@ const (
 	CILIUM_EBPF_ENGINE = "cilium"
 )
 
+const (
+	SNIFFER_CONFIG = "SNIFFER_CONFIG"
```
amirmalka, line 17: `SNIFFER_CONFIG_ENV_VAR` might be a more meaningful name.

Also, please combine const declarations

amirmalka (dismissed):

pkg/accumulator/accumulator.go
amirmalka: just so that you don't check in every iteration if i == 0.

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+var lock = &sync.Mutex{}
+
+func createAccumulator() *Accumulator {
```
dwertent, line 49:
```suggestion
func NewAccumulator() *Accumulator {
```

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+		containersData: containersAccumulator{
+			accumulatorDataPerContainer: make(map[string]chan evData.EventData),
+		},
+	}
+
+	return accumulatorSingelton
+}
+
+func GetAccumulator() *Accumulator {
+	if accumulatorSingelton == nil {
```
dwertent, line 65:
```suggestion
	if accumulatorSingelton != nil {
	    return accumulatorSingelton
	}
```
[Required]: This is the correct way

pkg/accumulator/accumulator.go
dwertent, line 68:
```suggestion
		
```

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+		lock.Lock()
+		defer lock.Unlock()
+		if accumulatorSingelton == nil {
+			logger.L().Debug("Creating accumulatorSingelton now.")
+			accumulatorSingelton = createAccumulator()
+		} else {
+			logger.L().Debug("accumulatorSingelton already created.")
+		}
+	} else {
```
dwertent, line 74:
```suggestion
```

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+
+func (acc *Accumulator) getFirstTimestamp() (time.Time, error) {
+	if len(acc.accumulatorData) < 1 {
+		return time.Time{}, fmt.Errorf("getFirstTimestamp failed the slice accumulatorData has no members")
+	}
+	if len(acc.firstMapKeysOfAccumulatorData) < 1 {
+		return time.Time{}, fmt.Errorf("getFirstTimestamp failed the slice firstMapKeysOfAccumulatorData has no members")
+	}
+	if len(acc.accumulatorData[0][acc.firstMapKeysOfAccumulatorData[0]]) < 1 {
+		return time.Time{}, fmt.Errorf("getFirstTimestamp failed the slice acc.accumulatorData[0][acc.firstMapKeysOfAccumulatorData[0] has no members")
```
dwertent, line 104: [Required]: This should not be an error message, the error should tell the logic that went wrong and not a name of a variable.

pkg/accumulator/accumulator.go
dwertent, line 68: [Required]: You did not review this code before opening the PR. Please do not do this again!

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+		}
+	}
+	acc.createNewMap(event, index)
+	return index
+}
+
+func (acc *Accumulator) findIndexByTimestamp(event *evData.EventData) int {
+	for i := range acc.accumulatorData {
+		if len(acc.accumulatorData[i]) == 0 {
+			acc.createNewMap(event, i)
```
dwertent, line 132: [Question]: What? Why are you "creating things" in a "find" function?

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+			return i
+		}
+		firstKey := acc.firstMapKeysOfAccumulatorData[i]
```
dwertent, line 135: [Required]: **DO NOT ACCESS A SLICE BEFORE CHECKING IT'S LENGTH!!!!!!!!!!!!!!!!!!**
I do not want to see such code again!

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func (acc *Accumulator) removeAllStreamedContainers(event *evData.EventData) {
+	acc.containersData.registerMutex.Lock()
+	if len(acc.containersData.accumulatorDataPerContainer) > 0 {
+		for contID := range acc.containersData.accumulatorDataPerContainer {
+			acc.containersData.accumulatorDataPerContainer[contID] <- *event
+		}
+	}
+	acc.containersData.registerMutex.Unlock()
```
dwertent, line 150: [TODO]: move this to the top and add `defer`

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func (acc *Accumulator) addEventToCacheAccumulator(event *evData.EventData, index int) {
```
dwertent, line 153:
```suggestion
func (acc *Accumulator) addEventToCache(event *evData.EventData, index int) {
```

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+	acc.syncReaderWriterAccumulatorData.Lock()
+	a := acc.accumulatorData[index]
+	a[event.GetEventContainerID()] = append(a[event.GetEventContainerID()], *event)
+	acc.accumulatorData[index][event.GetEventContainerID()] = append(acc.accumulatorData[index][event.GetEventContainerID()], *event)
```
dwertent, line 157: Do you mean: 
```suggestion
	acc.accumulatorData[index][event.GetEventContainerID()] = a[event.GetEventContainerID()]
```

What am I missing here?

pkg/accumulator/accumulator.go
dwertent, line 153: [Required]: Why is there no unitest for this function?

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+	if containerAccumulatorChan, exist := acc.containersData.accumulatorDataPerContainer[event.GetEventContainerID()]; exist {
+		containerAccumulatorChan <- *event
+	}
+	acc.containersData.registerMutex.Unlock()
+}
+
+func (acc *Accumulator) accumulateEbpfEngineData() {
+	for {
+		event := <-acc.mainDataChannel
+		logger.L().Debug("metadataAcc ", helpers.String("", fmt.Sprintf("%v", event)))
```
dwertent, line 172: [Question]: How will this event look like?

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+func (acc *Accumulator) streamEventToRegisterContainer(event *evData.EventData, index int) {
+	acc.containersData.registerMutex.Lock()
+	if containerAccumulatorChan, exist := acc.containersData.accumulatorDataPerContainer[event.GetEventContainerID()]; exist {
+		containerAccumulatorChan <- *event
+	}
+	acc.containersData.registerMutex.Unlock()
+}
+
+func (acc *Accumulator) accumulateEbpfEngineData() {
+	for {
```
dwertent, line 170: [Required]: Add a comment explaining the flow of this function

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	go acc.accumulateEbpfEngineData()
+	go acc.getEbpfEngineData()
+	go acc.getEbpfEngineError(errChan)
+	return nil
+}
+
+func (acc *ContainerAccumulator) registerContainerAccumulator() {
+	accumulatorSingelton.containersData.registerMutex.Lock()
```
dwertent, line 216: [Required]: Use `defer`

dwertent (changes requested):

pkg/accumulator/accumulator.go
rcohencyberarmor: no, we have only problem when we are writing and reading to a golang map, we can have multiple readers, and also this is in the same thread of the writer.

pkg/accumulator/accumulator.go
```diff
@@ -0,0 +1,256 @@
...
+
+func (acc *ContainerAccumulator) StartContainerAccumulator() {
+	acc.registerContainerAccumulator()
+}
+
+func (acc *ContainerAccumulator) StopContainerAccumulator() {
+	acc.unregisterContainerAccumulator()
+}
+
+func GetCacheAccumulator() *Accumulator {
```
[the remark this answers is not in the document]
rcohencyberarmor: It will be called later

pkg/accumulator/accumulator.go
rcohencyberarmor: this is how implementing a singleton, please see this link:
https://refactoring.guru/design-patterns/singleton/go/example

pkg/accumulator/accumulator.go
rcohencyberarmor: this is how implementing a singleton, please see this link:
https://refactoring.guru/design-patterns/singleton/go/example

pkg/accumulator/accumulator.go
rcohencyberarmor: this is how implementing a singleton, please see this link:
https://refactoring.guru/design-patterns/singleton/go/example

pkg/config/config.go
```diff
@@ -9,16 +9,19 @@ import (
 )
 
 const (
-	FALCO_EBPF_ENGINE  = "falco"
-	CILIUM_EBPF_ENGINE = "cilium"
+	FALCO_EBPF_ENGINE      = "falco"
```
dwertent, line 12:
```suggestion
	EBPFEngineFalco      = "falco"
```
[Required]: Please follow the [Go convention for naming variables](https://google.github.io/styleguide/go/decisions#constant-names).

pkg/config/config.go
```diff
@@ -9,16 +9,19 @@ import (
...
+	CILIUM_EBPF_ENGINE     = "cilium"
+	SNIFFER_CONFIG_ENV_VAR = "SNIFFER_CONFIG_ENV_VAR"
 )
 
+const ()
```
dwertent, line 17: [Required]: Remove this

pkg/config/v1/config_data_falco_mock.go
```diff
@@ -12,9 +16,9 @@ func (c *ConfigDataFalcoMock) GetFalcoSyscallFilter() []string {
 }
 
 func (c *ConfigDataFalcoMock) GetFalcoKernelObjPath() string {
-	return "./../../resources/ebpf/mock_falco_ebpf_engine/kernel_obj_mock.o"
+	return "./../config/testdata/mock_falco_ebpf_engine/kernel_obj_mock.o"
```
dwertent, line 19: [Required]: You should use dynamic relative paths when testing

pkg/event_data_storage/accumulator.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+func CreateContainerAccumulator(containerID string, dataChannel chan evData.EventData) *ContainerAccumulator {
+	return &ContainerAccumulator{
+		dataChannel: dataChannel,
+		containerID: containerID,
+	}
+}
+
+func (acc *Accumulator) createNewSlotInIndex(event *evData.EventData, index int) {
```
dwertent, line 80:
```suggestion
func (acc *Accumulator) createNewSlot(event *evData.EventData, index int) {
```
But still, what does "create new slot" mean? You should call it: "add id", "insert id", etc. so it is clear what this function do from the application perspective and not how the function implements it.

dwertent (changes requested):

pkg/config/config.go
```diff
@@ -9,16 +9,17 @@ import (
 )
 
 const (
-	FALCO_EBPF_ENGINE  = "falco"
-	CILIUM_EBPF_ENGINE = "cilium"
+	EBPFEngineFalco        = "falco"
+	EBPFEngineCilium       = "cilium"
+	SNIFFER_CONFIG_ENV_VAR = "SNIFFER_CONFIG_ENV_VAR"
```
matthyx, line 14: SNIFFER_CONFIG ? Probably need to change from SNIFFER also

pkg/config/config_test.go
```diff
@@ -2,14 +2,18 @@ package config
...
 	"testing"
 )
 
 func TestConfig(t *testing.T) {
-	err := os.Setenv("SNIFFER_CONFIG", "../../configuration/ConfigurationFile.json")
+	configPath := path.Join(currentDir(), "..", "..", "configuration", "ConfigurationFile.json")
+	err := os.Setenv(SNIFFER_CONFIG_ENV_VAR, configPath)
 	if err != nil {
-		t.Fatalf("failed to set env SNIFFER_CONFIG with err %v", err)
+		t.Fatalf("failed to set env SNIFFER_CONFIG_ENV_VAR with err %v", err)
```
matthyx, line 16: use variable here

pkg/config/config_test.go
```diff
@@ -41,3 +45,9 @@ func TestConfig(t *testing.T) {
 		t.Fatalf("GetEbpfEngineLoaderPath failed")
 	}
 }
+
+func currentDir() string {
+	_, filename, _, _ := runtime.Caller(1)
```
matthyx, line 50: https://pkg.go.dev/os#Getwd ?

pkg/config/v1/config_data_falco_mock.go
```diff
@@ -12,9 +22,15 @@ func (c *ConfigDataFalcoMock) GetFalcoSyscallFilter() []string {
...
-	return "./../../resources/ebpf/mock_falco_ebpf_engine/kernel_obj_mock.o"
+	return path.Join(currentDir(), "..", "testdata", "mock_falco_ebpf_engine", "kernel_obj_mock.o")
 }
 
 func (c *ConfigDataFalcoMock) GetEbpfEngineLoaderPath() string {
-	return "./../../resources/ebpf/mock_falco_ebpf_engine/userspace_app_mock"
+	return path.Join(currentDir(), "..", "testdata", "mock_falco_ebpf_engine", "userspace_app_mock")
+}
+
+func currentDir() string {
```
matthyx, line 32: should probably move this to a tools or util package

pkg/config/config_test.go
rcohencyberarmor: I have copy it from the kubescape. I prefer to use in the same method that other kubescape projects use

pkg/config/config.go
```diff
@@ -9,16 +9,17 @@ import (
 )
 
 const (
-	FALCO_EBPF_ENGINE  = "falco"
-	CILIUM_EBPF_ENGINE = "cilium"
+	EBPFEngineFalco  = "falco"
```
matthyx, line 12: I think you have the same const in `config_data.go`

pkg/config/config.go
```diff
@@ -9,16 +9,17 @@ import (
...
+	EBPFEngineCilium = "cilium"
+	CONFIG_ENV_VAR   = "CONFIG_ENV_VAR"
```
matthyx, line 14: please look everywhere to use variables with camel case instead of snake case

pkg/event_data_storage/accumulator.go
```diff
@@ -0,0 +1,265 @@
...
+
+type containersEventStreamer struct {
+	streamDataChannelForContainerID map[string]chan evData.EventData
+	registerMutex                   sync.Mutex
+}
+
+type Accumulator struct {
+	data                            []map[string][]evData.EventData
+	startStatus                     bool
+	syncReaderWriterData            *sync.RWMutex
```
matthyx, line 30: I don't see often pointer to mutexes, this is normally used in specific cases - if so document why you need pointer here.

pkg/event_data_storage/accumulator.go
```diff
@@ -0,0 +1,265 @@
...
+	"sniffer/pkg/ebpfeng"
+	evData "sniffer/pkg/ebpfev/v1"
+
+	logger "github.com/kubescape/go-logger"
+	"github.com/kubescape/go-logger/helpers"
+)
+
+const (
+	AccumulatorSize     = 10
+	DROP_EVENT_OCCURRED = "drop event occurred\n"
```
matthyx, line 19: camel case (like everywhere else)

matthyx (changes requested):
getting better, although it still looks like C++ code
please run `go mod tidy` as you have unused modules

matthyx (approved):

dwertent (approved):

PR #10, opened 2023-03-02
container handling implementation

rcohencyberarmor:
## Overview
container handler implementation
## Additional Information

> Any additional information that may be useful for reviewers to know 
-->


pkg/config/config.go
```diff
@@ -67,3 +67,27 @@ func (cfg *Config) GetFalcoKernelObjPath() string {
 func (cfg *Config) GetEbpfEngineLoaderPath() string {
 	return cfg.data.GetEbpfEngineLoaderPath()
 }
+
+func (cfg *Config) GetUpdateDataPeriod() int {
```
matthyx, line 71: I wonder if we should not directly return a duration here, I know you cannot unmarshall a duration, but it's not the client's responsibility to know he should convert int into seconds. WDYT?

pkg/config/config.go
```diff
@@ -67,3 +67,27 @@ func (cfg *Config) GetFalcoKernelObjPath() string {
...
+	return cfg.data.GetUpdateDataPeriod()
+}
+
+func (cfg *Config) GetSniffingMaxTimes() int {
```
matthyx, line 75: same here

matthyx (commented):
small nit regarding duration, otherwise let's move on and merge - this PR is too big

pkg/config/config.go
rcohencyberarmor: so do you think it is better that we change the return here time.Second instead?

matthyx: return directly a time.duration that you can use for the ticker

pkg/conthandler/container_main_handler.go
```diff
@@ -120,7 +120,7 @@ func (ch *ContainerHandler) startTimer(watchedContainer watchedContainerData, co
 }
 
 func createTicker() *time.Ticker {
-	return time.NewTicker(time.Duration(config.GetConfigurationConfigContext().GetUpdateDataPeriod()) * time.Second)
+	return time.NewTicker(time.Duration(config.GetConfigurationConfigContext().GetUpdateDataPeriod()))
```
matthyx, line 123: do you still need to convert to time.duration here?

rcohencyberarmor: forgot to push this file

pkg/conthandler/container_main_handler.go
```diff
@@ -1,5 +1,214 @@
...
+	}
+	watchedContainer := containerDataInterface.(watchedContainerData)
+
+	err := watchedContainer.containerAggregator.StartAggregate(watchedContainer.syncChannel[StepEventAggregator])
+	if err != nil {
+		return
+	}
+
+	now := time.Now()
+	configStopTime := time.Duration(config.GetConfigurationConfigContext().GetSniffingMaxTimes())
```
matthyx, line 140: `time.Duration` conversion is not needed here

pkg/conthandler/v1/container_handler_data.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/kubescape/k8s-interface/workloadinterface"
+)
+
+const (
+	CONTAINER_RUNNING = "Running"
```
matthyx, line 12: snake case

pkg/storageclient/storage_client_mock.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+type StorageHttpClientMock struct {
+}
+
+const (
+	NGINX = "nginx"
+)
+
+var nginx_SBOM_spdx_bytes []byte
```
matthyx, line 17: snake case

matthyx (approved):
let's merge

PR #34, opened 2023-03-23
store filtered SBOMs in the right storage API

rcohencyberarmor:

### DIFF
--- a/pkg/sbom/v1/sbom_spdx_storage_format.go
+++ b/pkg/sbom/v1/sbom_spdx_storage_format.go
@@ -25,7 +25,7 @@ const (
 
 type SBOMData struct {
 	spdxData                              spdxv1beta1.SBOMSPDXv2p3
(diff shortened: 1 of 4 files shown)


pkg/sbom/v1/sbom_spdx_storage_format.go
```diff
@@ -50,7 +50,9 @@ func (sbom *SBOMData) StoreSBOM(sbomData any) error {
 	for i := range sbom.spdxData.Spec.SPDX.Files {
```
dwertent, line 50: If any of the inner files are empty, you should quit and stop sniffing.
But make sure you are not accessing here pointers without checking if they are nil.

dwertent:
@rcohencyberarmor Please see my comment above

pkg/sbom/v1/sbom_spdx_storage_format.go
rcohencyberarmor: I don't think that there is a SBOM with no file in it.

PR #44, opened 2023-03-28
change labels and annotations

rcohencyberarmor:

### DIFF
--- a/pkg/sbom/v1/sbom_spdx_storage_format.go
+++ b/pkg/sbom/v1/sbom_spdx_storage_format.go
@@ -4,19 +4,30 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/armosec/utils-k8s-go/wlid"
+	"github.com/kubescape/go-logger"
+	"github.com/kubescape/go-logger/helpers"
 	instanceidhandler "github.com/kubescape/k8s-interface/instanceidhandler"
 	spdxv1beta1 "github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/validation"
 )
 
 const (
 	// CreatorType should be one of "Person", "Organization", or "Tool"
-	Organization              = "Organization"
-	Tool                      = "Tool"
-	Person                    = "Person"
-	KubescapeOrganizationName = "Kubescape"
-	KubescapeNodeAgentName    = "KubescapeNodeAgent"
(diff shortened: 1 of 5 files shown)


matthyx (approved):

pkg/sbom/v1/sbom_spdx_storage_format.go
```diff
@@ -128,14 +139,38 @@ func (sbom *SBOMData) StoreFilteredSBOMName(name string) {
...
 	}
 	sbom.filteredSpdxData.ObjectMeta.SetLabels(labels)
+
+	annotations := make(map[string]string)
+	if imageID != "" {
+		annotations[imageIDKeyForAnnotations] = imageID
+	}
+	instanceIDString := instanceID.GetStringFormatted()
+	if instanceIDString != "" {
+		annotations[instanceIDKeyForAnnotations] = instanceIDString
```
vladklokun, line 171: issue: label value is inconsistent with Kubevuln

According to the constants definition:
```go
const (
	labelPrefix                 = "kubescape.io"
	Organization                = "Organization"
	Tool                        = "Tool"
	Person                      = "Person"
	KubescapeOrganizationName   = "Kubescape"
	KubescapeNodeAgentName      = "KubescapeNodeAgent"
	RelationshipContainType     = "CONTAINS"
	namespaceLabelKey           = "/workload-namespace"
	kindLabelKey                = "/workload-kind"
	nameLabelKey                = "/workload-name"
	containerNameLabelKey       = "/workload-conatienr-name"
	imageIDKeyForAnnotations    = "imageID"
	instanceIDKeyForAnnotations = "instanceID"
)
```

So `instanceIDKeyForAnnotations` is `instanceID`. However, this is inconsistent with keys that [Kubevuln uses](https://github.com/kubescape/kubevuln/blob/e6f88e14e4a894503915473cb5e056581d8dc6a7/core/domain/sbom.go#L9) to store such annotations and values that we agreed with David.

The Instance ID annotation should be:
```
"kubescape.io/instance-id": "instanceID-value"
```

vladklokun (changes requested):

pkg/sbom/v1/sbom_spdx_storage_format.go
rcohencyberarmor: @matthyx it is ok by you that I'll add the instance id annotations key to the instanceidhandler package under the k8s-interface repo, so we will both take it from there?

matthyx: ok

matthyx (approved):

vladklokun (approved):

PR #46, opened 2023-04-18
disable internal check of SBOM exist

rcohencyberarmor:


dwertent (approved):
We should re-visit this

PR #110, opened 2023-07-31
reduce allocations when adding files

matthyx:
## Overview
This is with full_path enabled for trace_open and trace_openat:
image

```
GOMEMLIMIT=250MiB
resources:
  limits:
    cpu: 1500m
    memory: 600Mi
  requests:
    cpu: 500m
    memory: 600Mi                                                                                                                                                                
```


pkg/config/config.go
```diff
@@ -32,9 +32,11 @@ func LoadClusterData(path string) (ClusterData, error) {
...
 
 type Config struct {
-	EnableRelevancy  bool          `mapstructure:"relevantCVEServiceEnabled"`
-	MaxSniffingTime  time.Duration `mapstructure:"maxSniffingTimePerContainer"`
-	UpdateDataPeriod time.Duration `mapstructure:"updateDataPeriod"`
+	EnableFullPathTracing bool          `mapstructure:"fullPathTracingEnabled"`
+	EnableRelevancy       bool          `mapstructure:"relevantCVEServiceEnabled"`
+	InitialDelay          time.Duration `mapstructure:"initialDelay"`
+	MaxSniffingTime       time.Duration `mapstructure:"maxSniffingTimePerContainer"`
+	UpdateDataPeriod      time.Duration `mapstructure:"updateDataPeriod"`
```
dwertent, line 39: We should add this to the helm chart as well

dwertent: Also, we should have default values (if missing from config)

dwertent (approved):
Well done!
Please see my comments :)

pkg/config/config.go
matthyx: like in line https://github.com/kubescape/node-agent/pull/110/files#diff-a3d824da3c42420cd5cbb0a4a2c0e7b5bfddd819652788a0596d195dc6e31fa5R48 ?

matthyx: what we should modify in the helm chart is the `updateDataPeriod` as well as the resources...

PR #112, opened 2023-08-06
use fullPath if not empty

dwertent:


slashben (approved):
This is a wonderfullpath change

PR #126, opened 2023-09-13
ensure no missing info before handleRelevancy

matthyx:
## PR Type:
Bug fix

___
## PR Description:
This PR addresses an issue where short-lived containers, such as host-scanner, could attempt to generate a filteredSBOM on exit with missing instanceID. The solution ensures that all necessary image information (ImageID, ImageTag, InstanceID, and Wlid) is present before handling relevancy.

___
## PR Main Files Walkthrough:
`pkg/relevancymanager/v1/relevancy_manager.go`: The function 'ensureImageID' has been renamed to 'ensureImageInfo' and its condition has been expanded to check for the presence of ImageID, ImageTag, InstanceID, and Wlid. This function is called in 'monitorContainer' method before handling relevancy.

___
## User Description:
with short lived containers, such as host-scanner, we could try to generate a filteredSBOM on exit with a missing instanceID
```
{"level":"warn","ts":"2023-09-12T17:29:05Z","msg":"failed to filter SBOM","container ID":"e095f32808abc9b8417b3c0d6035a727d50e17984a6d5347d36a74edc1ff687b","k8s workload":"kubescape/host-scanner-gwlth/host-sensor","error":"instance id is nil"}
```

### DIFF
--- a/pkg/relevancymanager/v1/relevancy_manager.go
+++ b/pkg/relevancymanager/v1/relevancy_manager.go
@@ -73,8 +73,8 @@ func (rm *RelevancyManager) deleteResources(watchedContainer *utils.WatchedConta
 	_ = rm.fileHandler.RemoveBucket(watchedContainer.ContainerID)
 }
 
-func (rm *RelevancyManager) ensureImageID(container *containercollection.Container, watchedContainer *utils.WatchedContainerData) {
-	if watchedContainer.ImageID == "" {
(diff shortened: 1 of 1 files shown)


dwertent (approved):
How is it possible you are missing info? What is the use-case? (this should never happen...)

matthyx:
> How is it possible you are missing info? What is the use-case? (this should never happen...)

apparently it does... since I saw this log - I guess more checks don't harm

PR #153, opened 2023-12-01
don't rely on ebpf event for dns resolution

matthyx:


pkg/dnsmanager/dns_manager.go
```diff
@@ -18,13 +17,23 @@ type DNSManager struct {
...
+func CreateDNSManager() *DNSManager {
 	return &DNSManager{}
 }
 
 func (dm *DNSManager) ProcessDNSEvent(dnsEvent tracerdnstype.Event) {
-	for _, address := range dnsEvent.Addresses {
-		dm.addressToDomainMap.Set(address, dnsEvent.DNSName)
+	if dnsEvent.NumAnswers > 0 {
+		// FIXME use dnsEvent.Nameserver to resolve the DNS name
+		addresses, err := net.LookupIP(dnsEvent.DNSName)
```
Daniel-GrunbergerCA, line 27: we want to specifically use the ig dns resolution...

pkg/networkmanager/network_neighbors.go
```diff
@@ -54,7 +54,7 @@ func generateNetworkNeighborsNameFromWlid(parentWlid string) string {
 
 func generateNetworkNeighborsLabels(workload k8sinterface.IWorkload) map[string]string {
 	return map[string]string{
-		instanceidhandlerV1.ApiVersionMetadataKey:      workload.GetApiVersion(),
+		//instanceidhandlerV1.ApiVersionMetadataKey:      workload.GetApiVersion(), FIXME invalid label, move to annotations
```
Daniel-GrunbergerCA, line 57: is this a fix for all our crds?

pkg/dnsmanager/dns_manager.go
Daniel-GrunbergerCA: why did we update the ig version?

matthyx: For a fix on arm processors... But are you sure it works with the older IG?

Daniel-GrunbergerCA: the uncrompessed issue doesn't bother us since it's only inside the cluster

Daniel-GrunbergerCA: yes I am sure. I will try to figure out what happened on the last version

pkg/networkmanager/network_neighbors.go
matthyx: This should be a discussion between David and Eran

pkg/dnsmanager/dns_manager.go
matthyx: We should still have a workaround when there's no address

matthyx:
@Daniel-GrunbergerCA I think this is ready to review, you can try the image quay.io/matthiasb_1/node-agent:fanotify

Daniel-GrunbergerCA (approved):

PR #267, opened 2024-04-17
Fix application profile patching

amirmalka:

### DIFF
--- a/tests/component_test.go
+++ b/tests/component_test.go
@@ -4,11 +4,21 @@
 package tests
 
 import (
+	"context"
+	"encoding/json"
 	"node-agent/pkg/utils"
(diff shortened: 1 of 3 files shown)


tests/component_test.go
```diff
@@ -397,3 +407,77 @@ func Test_07_RuleBindingApplyTest(t *testing.T) {
 //func Test_08_BasicAlertTestExistingProfile(t *testing.T) {
 //
 //}
+
+func TestApplicationProfilePatching_AlwaysFail(t *testing.T) {
```
amirmalka, line 411: This test needs to be fixed (we expect an error at the end, but we shouldn't get one)

dwertent (approved):

PR #281, opened 2024-05-03
add runtime path validation with real open file

matthyx:


matthyx:
it doesn't fail on the wrong `runc` path, waiting for IG guidance

PR #283, opened 2024-05-13
Changing alert behavior on whitelisted processes

slashben:
## Overview

This PR changes behavior on two rules: 
* R1001 - Exec not in base image
* R1004 - Exec in mounted path

Both of the rules generated alerts on whitelisted processes. This causes many false positives for users who have good application profiles. 

The change removes alerts for whitelisted processes, though this should be refined in the future when the rule information becomes available in the application profile therefore a full comparison of behavior can be done.

cc: @amitschendel

### DIFF
--- a/pkg/ruleengine/v1/helpers.go
+++ b/pkg/ruleengine/v1/helpers.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"node-agent/pkg/objectcache"
 	"path/filepath"
+	"slices"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


amitschendel:
Initially we didn't made those rules application profile dependent because of 2 things:
1. We want this alert to fire even if we don't have an application profile, we don't have a mechanism like this in node-agent today.
2. In my perspective, even if this is the behavior of the application, it's not a "good" practice, therefore I am okay with alerting and the user will need to choose between fixing the image or ignoring the alert.

PR #285, opened 2024-05-13
Improvement/changing r1001 r1004 behavior

slashben:
## Overview

This PR changes behavior on two rules:

R1001 - Exec not in base image
R1004 - Exec in mounted path
Both of the rules generated alerts on whitelisted processes. This causes many false positives for users who have good application profiles.

The change removes alerts for whitelisted processes, though this should be refined in the future when the rule information becomes available in the application profile therefore a full comparison of behavior can be done.

Related PR #283

### DIFF
--- a/pkg/ruleengine/v1/r1001_exec_binary_not_in_base_image.go
+++ b/pkg/ruleengine/v1/r1001_exec_binary_not_in_base_image.go
@@ -62,7 +62,7 @@ func (rule *R1001ExecBinaryNotInBaseImage) ProcessEvent(eventType utils.EventTyp
 
 	if execEvent.UpperLayer {
 
-		if whiteListed, err := isExecEventWhitelisted(execEvent, objectCache, false); whiteListed || err != nil {
+		if whiteListed, err := isExecEventWhitelisted(execEvent, objectCache, false); whiteListed && err != nil {
 			return nil
 		}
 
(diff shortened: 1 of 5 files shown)


matthyx:
@slashben can you fix TestR1004ExecFromMount ?

pkg/ruleengine/v1/r1001_exec_binary_not_in_base_image.go
```diff
@@ -50,7 +50,7 @@ func (rule *R1001ExecBinaryNotInBaseImage) ID() string {
 func (rule *R1001ExecBinaryNotInBaseImage) DeleteRule() {
 }
 
-func (rule *R1001ExecBinaryNotInBaseImage) ProcessEvent(eventType utils.EventType, event interface{}, objCache objectcache.ObjectCache) ruleengine.RuleFailure {
+func (rule *R1001ExecBinaryNotInBaseImage) ProcessEvent(eventType utils.EventType, event interface{}, objectCache objectcache.ObjectCache) ruleengine.RuleFailure {
```
matthyx, line 53: I like consistent naming...

matthyx (approved):

pkg/ruleengine/v1/r1001_exec_binary_not_in_base_image.go
slashben: Copy paste from here:
https://github.com/kubescape/node-agent/blob/544d7c9020608a75cc9c58eda32934b709ec49a7/pkg/ruleengine/v1/r0001_unexpected_process_launched.go#L68

PR #287, opened 2024-05-16
use socket to reach clamd

matthyx:


tests/chart/values.yaml
```diff
@@ -139,9 +142,9 @@ nodeAgent:
 clamav:
   name: clamav
   image:
-    repository: quay.io/armosec/klamav
-    tag: beta5
-    pullPolicy: Always
+    repository: clamav/clamav-debian
```
amitschendel, line 145: Please wrap this image using the [script](https://github.com/armosec/kubecop/tree/main/resources/clamav) because we filter out signatures that isn't for Linux

pkg/malwaremanager/v1/clamav/open.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/kubescape/go-logger/helpers"
+)
+
+func (c *ClamAVClient) handleOpenEvent(event *types.Event, containerPid uint32) malwaremanager.MalwareResult {
+	if event == nil {
+		return nil
+	}
+
+	// discard if it is an open for writing event
+	if slices.Contains(event.Flags, "O_WRONLY") {
```
amitschendel, line 26: I think there is a possibility to check it with bit wise operations. It will be much more efficient.

amitschendel (changes requested):

amitschendel (approved):

PR #290, opened 2024-05-21
bump clamav image tag, build for ARM

amirmalka:


amirmalka:
@matthyx FYI
looks like bumping the clamAV image is not working...

matthyx:
> @matthyx FYI looks like bumping the clamAV image is not working...

why are we so brittle :(

amitschendel (approved):

PR #296, opened 2024-05-27
use all statuses, add fallback annotation for app

matthyx:
fixing nits from previous PR


pkg/containerwatcher/v1/ig_k8sclient.go
```diff
@@ -48,15 +45,15 @@ func (k *IGK8sClientMock) GetNonRunningContainers(pod *corev1.Pod) []string {
 
 // GetRunningContainers returns a list of the containers of a given Pod that are running.
```
matthyx, line 46: we don't even check if the status is running here @dwertent

pkg/containerwatcher/v1/ig_k8sclient.go
```diff
@@ -48,15 +45,15 @@ func (k *IGK8sClientMock) GetNonRunningContainers(pod *corev1.Pod) []string {
...
+	var containers []containercollection.Container
 
 	labels := map[string]string{}
 	for k, v := range pod.ObjectMeta.Labels {
 		labels[k] = v
 	}
 
-	containerStatuses := append([]v1.Container{}, pod.Spec.InitContainers...)
-	containerStatuses = append(containerStatuses, pod.Spec.Containers...)
+	containerStatuses := slices.Concat(pod.Spec.InitContainers,
```
dwertent, line 55: Very fancy 😉

dwertent (approved):

PR #306, opened 2024-06-05
Adding a check if the container is still being monitord before readin…

amitschendel:
…g from perf buffer


matthyx (dismissed):

pkg/containerwatcher/v1/container_watcher_private.go
```diff
@@ -31,7 +31,7 @@ func (ch *IGContainerWatcher) containerCallback(notif containercollection.PubSub
 		ch.timeBasedContainers.Add(notif.Container.Runtime.ContainerID)
 	}
 
-	tracingContext := context.Background()
+	tracingContext := make(chan struct{})
```
matthyx, line 34: maybe a better name?

matthyx (approved):

PR #308, opened 2024-06-05
Ignoring coredns pod

amitschendel:


matthyx:
we already exclude `kube-system` pods

PR #311, opened 2024-06-17
Removing debug logs from object file

amitschendel:


matthyx (approved):
so it's just the binary file that was compiled from the C ?

PR #318, opened 2024-07-02
Enabling custom sniffing time for pods (via label)

slashben:
## Overview

This small feature code enables users to define per pod the length of the recording time of application profiles and not just rely on the global settings in the node-agent configuration "maxLearningPeriod".

The length of the learning period for a given pod's containers will be taken from the label `kubescape.io/max-sniffing-time` in the same format as the Helm chart's `maxLearningPeriod` value. If the label does not exist then the global configuration will be used as normal.

### DIFF
--- a/pkg/containerwatcher/v1/container_watcher_private.go
+++ b/pkg/containerwatcher/v1/container_watcher_private.go
@@ -16,6 +16,10 @@ import (
 	"github.com/kubescape/node-agent/pkg/utils"
 )
 
+const (
+	MaxSniffingTimeLabel = "kubescape.io/max-sniffing-time"
+)
+
 func (ch *IGContainerWatcher) containerCallback(notif containercollection.PubSubEvent) {
 
 	// do not trace the node-agent pod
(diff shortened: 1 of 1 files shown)


pkg/containerwatcher/v1/container_watcher_private.go
```diff
@@ -34,7 +38,18 @@ func (ch *IGContainerWatcher) containerCallback(notif containercollection.PubSub
...
 		logger.L().Info("start monitor on container", helpers.String("container ID", notif.Container.Runtime.ContainerID), helpers.String("k8s workload", k8sContainerID))
-		time.AfterFunc(ch.cfg.MaxSniffingTime, func() {
+
+		// Check if Pod has a label of max sniffing time
+		sniffingTime := ch.cfg.MaxSniffingTime
+		if podLabelMaxSniffingTime, ok := notif.Container.K8s.PodLabels[MaxSniffingTimeLabel]; ok {
+			if duration, err := time.ParseDuration(podLabelMaxSniffingTime); err == nil {
+				sniffingTime = duration
+			} else {
+				logger.L().Error("parsing sniffing time in label", helpers.Error(err))
```
matthyx, line 48: nit: I would have added that we default to `ch.cfg.MaxSniffingTime` in that case, and maybe only warn

matthyx (approved):

PR #326, opened 2024-07-11
Adding java exception on ld_preload rule

amitschendel:


pkg/ruleengine/v1/r1011_ld_preload_hook.go
```diff
@@ -20,6 +20,7 @@ const (
 	R1011ID         = "R1011"
 	R1011Name       = "LD_PRELOAD Hook"
 	LD_PRELOAD_FILE = "/etc/ld.so.preload"
+	JAVA_COMM       = "java"
```
matthyx, line 23: are we sure it's the only name?

matthyx (approved):

pkg/ruleengine/v1/r1011_ld_preload_hook.go
amitschendel: I think, this is what we saw for now, if we will encounter more we will add them.

PR #338, opened 2024-08-13
Feature/rules improvements

amitschendel:


slashben (approved):

matthyx (approved):

pkg/ebpf/gadgets/ssh/tracer/bpf/ssh.bpf.c
```diff
@@ -0,0 +1,108 @@
...
+            event->dst_port = bpf_ntohs(tcph.dest);
+
+            event->timestamp = bpf_ktime_get_boot_ns();
+        }
+        __u64 skb_len = skb->len;
+        bpf_perf_event_output(skb, &events, skb_len << 32 | BPF_F_CURRENT_CPU, event, sizeof(struct event));
+    }
+
+    return 0;
+}
```
slashben, line 105: @amitschendel , if I understand correctly, this will fire an event for every SSH frame. I am not sure we need event per packet, maybe per connection.

slashben: I take the above back. The `SSH_` header is only there in the handshake messages (one that client sends to server, the second when the server sends to the client)

Still, will we get two alerts here?

amitschendel: Correct, this is why I am filtering out the response in the rule.

PR #341, opened 2024-08-18
Improving eBPF support checks

amitschendel:

### DIFF
--- a/internal/validator/ebpf/verifier.go
+++ b/internal/validator/ebpf/verifier.go
@@ -1,0 +1,100 @@
+package ebpf
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/kubescape/go-logger"
+	"github.com/kubescape/go-logger/helpers"
+)
+
+// VerifyEbpf checks if the node-agent ebpf features can run on the current system.
+func VerifyEbpf() error {
+	if err := checkBTFSupport(); err != nil {
+		return err
+	}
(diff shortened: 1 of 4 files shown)


internal/validator/ebpf/verifier.go
```diff
@@ -0,0 +1,100 @@
...
+	cmd := exec.Command("sh", "-c", "grep -q CONFIG_DEBUG_INFO_BTF=y /boot/config-$(uname -r)")
+	if err := cmd.Run(); err == nil {
+		return nil // BTF is enabled in kernel config
+	}
+
+	return errors.New("BTF support not detected")
+}
+
+// CheckEBPFSupport checks for eBPF support.
+func checkEBPFSupport() error {
```
matthyx, line 57: maybe rename to checkKernelVersion?

matthyx (dismissed):
so instead of checking <5.4 we check <4.4 with BTF?

amitschendel:
> so instead of checking <5.4 we check <4.4 with BTF?

We are checking for BTF which is mandatory for our sensor and ig.
Regarding the kernel version, checking 5.4 is because since 5.4 there is BTF/CO-RE but some kernels "lies" with their version so it's not valuable to check it. 
Since kernel 4.4 there is eBPF so we can check that just in case but again, it's not too much valuable but it's more for us to have more debugging information, this is why we don't exist on the error but log it.

What do you think? @matthyx

matthyx:
Fine go ahead

PR #359, opened 2024-09-02
add jitter to all sniffing tickers

matthyx:


pkg/containerwatcher/v1/container_watcher_private.go
```diff
@@ -44,7 +44,7 @@ func (ch *IGContainerWatcher) containerCallback(notif containercollection.PubSub
 		logger.L().Info("start monitor on container", helpers.String("container ID", notif.Container.Runtime.ContainerID), helpers.String("k8s workload", k8sContainerID))
 
 		// Check if Pod has a label of max sniffing time
-		sniffingTime := ch.cfg.MaxSniffingTime
+		sniffingTime := utils.AddJitter(ch.cfg.MaxSniffingTime, ch.cfg.MaxJitterPercentage)
```
amitschendel, line 47: This is for us to play so let's not apply jitter here.

amitschendel (changes requested):

amitschendel (approved):

PR #362, opened 2024-09-05
remove dynamic client calls for known types

matthyx:


pkg/watcher/cooldownqueue/cooldownqueue.go
```diff
@@ -40,11 +41,10 @@ func NewCooldownQueue() *CooldownQueue {
 
 // makeEventKey creates a unique key for an event from a watcher
 func makeEventKey(e watch.Event) string {
-	object, ok := e.Object.(*unstructured.Unstructured)
-	if !ok {
-		return ""
-	}
-	return strings.Join([]string{object.GroupVersionKind().Group, object.GroupVersionKind().Version, object.GetKind(), object.GetNamespace(), object.GetName()}, "/")
+	object := e.Object.(runtime.Object)
```
matthyx, line 44: we'll get a panic if the assertion fails (which is better than returning "")

pkg/rulebindingmanager/cache/cache.go
```diff
@@ -84,19 +84,13 @@ func (c *RBCache) AddNotifier(n *chan rulebindingmanager.RuleBindingNotify) {
...
 	var rbs []rulebindingmanager.RuleBindingNotify
 
-	switch obj.GetKind() {
-	case "Pod":
-		pod, err := unstructuredToPod(obj)
-		if err != nil {
-			logger.L().Error("failed to convert unstructured to pod", helpers.Error(err))
-			return
-		}
+	if pod, ok := obj.(*corev1.Pod); ok {
```
amitschendel, line 90: Do you want to keep the switch case which is a bit more elegant?
You can check the kind, see [here](https://pkg.go.dev/k8s.io/apimachinery/pkg/runtime#Object).

pkg/rulebindingmanager/types/v1/types.go
```diff
@@ -52,3 +53,62 @@ func (r *RuntimeAlertRuleBindingRule) Equal(other *RuntimeAlertRuleBindingRule)
...
+
+func (r *RuntimeAlertRuleBindingRule) DeepCopy() RuntimeAlertRuleBindingRule {
+	if r == nil {
+		return RuntimeAlertRuleBindingRule{}
+	}
+	out := *r
+	if r.Parameters != nil {
+		out.Parameters = make(map[string]interface{}, len(r.Parameters))
+		for key, val := range r.Parameters {
+			out.Parameters[key] = val
```
amitschendel, line 106: Are we okay with this shallow copy? we don't need deep copy as well?

pkg/rulebindingmanager/types/v1/types.go
```diff
@@ -52,3 +53,62 @@ func (r *RuntimeAlertRuleBindingRule) Equal(other *RuntimeAlertRuleBindingRule)
...
+		}
+	}
+	if r.RuleTags != nil {
+		out.RuleTags = make([]string, len(r.RuleTags))
+		copy(out.RuleTags, r.RuleTags)
+	}
+	return out
+}
```
amitschendel, line 114: Add new line at the end of a file.

pkg/watcher/cooldownqueue/cooldownqueue.go
amitschendel: Do we really want to panic in production (?)

amitschendel (approved):

pkg/watcher/cooldownqueue/cooldownqueue.go
matthyx: yes, but we are catching the missing resource in other places before reaching that

pkg/rulebindingmanager/cache/cache.go
matthyx: I agree with the switch case, however I didn't find a way to use type inference which is what we want here

pkg/rulebindingmanager/types/v1/types.go
matthyx: I'm removing these changes, I don't need them since we're forced to use dynamic client for the RuntimeAlertRuleBinding

PR #365, opened 2024-09-11
Feature/dynamic opens

afek854:
## Overview

Note: Maybe in sensitive file we don't want it.


pkg/ruleengine/v1/helpers.go
```diff
@@ -153,3 +153,32 @@ func interfaceToStringSlice(val interface{}) ([]string, bool) {
 	}
 	return nil, false
 }
+
+func CompareDynamic(dynamicPath, regularPath string) (string, bool) {
+	const dynamic = "<dynamic>"
```
matthyx, line 158: this should use the same const as the storage, otherwise we might get into trouble

slashben (dismissed):

go.mod
```diff
@@ -22,7 +22,7 @@ require (
 	github.com/kubescape/backend v0.0.20
 	github.com/kubescape/go-logger v0.0.22
 	github.com/kubescape/k8s-interface v0.0.170
-	github.com/kubescape/storage v0.0.111
+	github.com/kubescape/storage v0.0.112-0.20240912075751-9a554eb26977
```
matthyx, line 25: use the release pls

matthyx (changes requested):
update storage

pkg/ruleengine/v1/helpers.go
afek854: DONE

go.mod
afek854: I need to merge https://github.com/kubescape/storage/pull/148
before

PR #367, opened 2024-09-12
add random delay when fetching full AP and NN

matthyx:
I'm not sure it's effective, but that's a mitigation we talked about


amitschendel (approved):

matthyx:
@amitschendel reworked the logic here

amitschendel:
@matthyx 
Did you see improvements with this?

matthyx:
We have to soak it in a big cluster... The one created with 20 nodes and the script from Boris wasn't big enough to show a difference

PR #371, opened 2024-09-19
Feature/cooldown

amitschendel:


matthyx:
I think we don't need this anymore

PR #377, opened 2024-09-24
remove relevancy manager

matthyx:


matthyx:
should not forget to update RBAC

amitschendel (commented):
Looks good!
What with the config? don't you want to remove it from the config as well?

amitschendel (approved):
Nice!

PR #388, opened 2024-10-08
Xmr mining all options

sarieren:


sarieren:
detect xmr mining

matthyx:
@amitschendel can you have a look?

PR #389, opened 2024-10-22
User generated application profiles/NN

amitschendel:


amitschendel:
cc: @rotemamsa @slashben @jnathangreeg 
Once merged and released we can move with the feature.

matthyx:
> cc: @rotemamsa @slashben @jnathangreeg Once merged and released we can move with the feature. Currently waiting for CR from @matthyx .

looking at it now

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
```diff
@@ -213,57 +293,104 @@ func (ap *ApplicationProfileCacheImpl) removeContainer(containerID string) {
...
+func (ap *ApplicationProfileCacheImpl) performMerge(normalProfile, userManagedProfile *v1beta1.ApplicationProfile) *v1beta1.ApplicationProfile {
+	mergedProfile := normalProfile.DeepCopy()
+
+	// Merge spec
+	mergedProfile.Spec.Containers = ap.mergeContainers(mergedProfile.Spec.Containers, userManagedProfile.Spec.Containers)
+	mergedProfile.Spec.InitContainers = ap.mergeContainers(mergedProfile.Spec.InitContainers, userManagedProfile.Spec.InitContainers)
+	mergedProfile.Spec.EphemeralContainers = ap.mergeContainers(mergedProfile.Spec.EphemeralContainers, userManagedProfile.Spec.EphemeralContainers)
+
+	// Remove the user-managed annotation
+	delete(mergedProfile.Annotations, "kubescape.io/managed-by")
```
matthyx, line 329: this is not necessary, since `mergeProfile` is the `normalProfile`

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
```diff
@@ -47,34 +48,113 @@ func newApplicationProfileState(ap *v1beta1.ApplicationProfile) applicationProfi
...
+		storageClient:       storageClient,
+		containerToSlug:     maps.SafeMap[string, string]{},
+		slugToAppProfile:    maps.SafeMap[string, *v1beta1.ApplicationProfile]{},
+		slugToContainers:    maps.SafeMap[string, mapset.Set[string]]{},
+		slugToState:         maps.SafeMap[string, applicationProfileState]{},
+		allProfiles:         mapset.NewSet[string](),
+		userManagedProfiles: maps.SafeMap[string, *v1beta1.ApplicationProfile]{},
 	}
 
+	return cache
```
matthyx, line 75: I think the compiler will revert that change anyways

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
```diff
@@ -47,34 +48,113 @@ func newApplicationProfileState(ap *v1beta1.ApplicationProfile) applicationProfi
...
 }
 
 // ------------------ objectcache.ApplicationProfileCache methods -----------------------
 
+func (ap *ApplicationProfileCacheImpl) handleUserManagedProfile(appProfile *v1beta1.ApplicationProfile) {
+	baseProfileName := strings.TrimPrefix(appProfile.GetName(), "ug-")
+	baseProfileUniqueName := objectcache.UniqueName(appProfile.GetNamespace(), baseProfileName)
+
+	// Get the full user managed profile from the storage
+	fullAP, err := ap.getApplicationProfile(appProfile.GetNamespace(), appProfile.GetName())
```
matthyx, line 85: maybe rename to `userManagedProfile` like elsewere

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
```diff
@@ -213,57 +293,104 @@ func (ap *ApplicationProfileCacheImpl) removeContainer(containerID string) {
...
+
+	return mergedProfile
+}
+
+func (ap *ApplicationProfileCacheImpl) mergeContainers(normalContainers, userManagedContainers []v1beta1.ApplicationProfileContainer) []v1beta1.ApplicationProfileContainer {
+	// Create a map to store containers by name
+	containerMap := make(map[string]int) // map name to index in slice
+
+	// Store indices of normal containers
+	for i := range normalContainers {
```
matthyx, line 339: we don't need that... the order of containers should be kept as in the Pod for both profiles (otherwise I will complain to @jnathangreeg )
maybe something strange could appear if we have ephemeral containers added in either profile (always appended at the right place, leaving empty slices if needed)

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
```diff
@@ -213,57 +293,104 @@ func (ap *ApplicationProfileCacheImpl) removeContainer(containerID string) {
...
+
+	// Remove the user-managed annotation
+	delete(mergedProfile.Annotations, "kubescape.io/managed-by")
+
+	return mergedProfile
+}
+
+func (ap *ApplicationProfileCacheImpl) mergeContainers(normalContainers, userManagedContainers []v1beta1.ApplicationProfileContainer) []v1beta1.ApplicationProfileContainer {
+	// Create a map to store containers by name
+	containerMap := make(map[string]int) // map name to index in slice
```
matthyx, line 336: make to max(len(normalContainers), len(userManagedContainers))

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
```diff
@@ -213,57 +293,104 @@ func (ap *ApplicationProfileCacheImpl) removeContainer(containerID string) {
...
+		} else {
+			normalContainers = append(normalContainers, userContainer)
+		}
+	}
+
+	return normalContainers
+}
+
+func (ap *ApplicationProfileCacheImpl) mergeContainer(normalContainer, userContainer *v1beta1.ApplicationProfileContainer) {
+	normalContainer.Capabilities = append(normalContainer.Capabilities, userContainer.Capabilities...)
```
matthyx, line 357: by design we should not have duplicates... but keep in mind we could (should?) rework later with Sets

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
```diff
@@ -213,57 +293,104 @@ func (ap *ApplicationProfileCacheImpl) removeContainer(containerID string) {
...
 func (ap *ApplicationProfileCacheImpl) deleteApplicationProfile(obj runtime.Object) {
-	apName := objectcache.MetaUniqueName(obj.(metav1.Object))
-	ap.slugToAppProfile.Delete(apName)
-	ap.slugToState.Delete(apName)
-	ap.allProfiles.Remove(apName)
+	appProfile := obj.(*v1beta1.ApplicationProfile)
+	apName := objectcache.MetaUniqueName(appProfile)
 
-	logger.L().Info("deleted application profile from cache", helpers.String("uniqueSlug", apName))
+	isUserManaged := appProfile.Annotations != nil &&
```
matthyx, line 368: please extract this into a method

pkg/objectcache/networkneighborhoodcache/networkneighborhoodcache.go
```diff
@@ -78,6 +81,215 @@ func (nn *NetworkNeighborhoodCacheImpl) GetNetworkNeighborhood(containerID strin
 	return nil
 }
 
+func (nn *NetworkNeighborhoodCacheImpl) handleUserManagedNN(netNeighborhood *v1beta1.NetworkNeighborhood) {
```
matthyx, line 84: I assume this one is copy/paste from AP (if not keep it similar so we can verify we have the same handling between AP and NN)

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
amitschendel: done

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
amitschendel: Yes this is because I had more logic there that I removed.

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
amitschendel: done

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
amitschendel: done

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
amitschendel: I thought about it but for now let's not handle this case.

PR #399, opened 2024-11-06
Feature/policy rules

afek854:


afek854:
Need to update storage in go.mod before

pkg/containerwatcher/v1/container_watcher.go
```diff
@@ -321,8 +326,11 @@ func CreateIGContainerWatcher(cfg config.Config, applicationProfileManager appli
 		if event.K8s.ContainerName == "" {
 			return
 		}
+		k8sContainerID := utils.CreateK8sContainerID(event.K8s.Namespace, event.K8s.PodName, event.K8s.ContainerName)
```
matthyx, line 329: we usually compute k8sContainerID inside the callbacks (see the managers) to avoid computing it if we disable the feature

pkg/eventreporters/rulepolicy/rule_policy.go
```diff
@@ -0,0 +1,27 @@
...
+
+func NewRulePolicyReporter(ruleManager rulemanager.RuleManagerClient, applicationProfileManager applicationprofilemanager.ApplicationProfileManagerClient) *RulePolicyReporter {
+	return &RulePolicyReporter{
+		ruleManager:               ruleManager,
+		applicationProfileManager: applicationProfileManager,
+	}
+}
+
+func (rpm *RulePolicyReporter) ReportEvent(eventType utils.EventType, event utils.K8sEvent, k8sContainerID string, allowedProcess string) {
+	rulesIds := rpm.ruleManager.EvaluateRulesForEvent(eventType, event)
```
matthyx, line 22: compute k8sContainerID here

pkg/utils/applicationprofile.go
```diff
@@ -73,10 +73,20 @@ func CreateCapabilitiesPatchOperations(capabilities, syscalls []string, execs ma
 		})
 	}
 
+	// add rule policies
+	if len(rulePolicies) != 0 {
+		rulePoliciesPath := fmt.Sprintf("/spec/%s/%d/policybyruleid/-", containerType, containerIndex)
+		profileOperations = append(profileOperations, PatchOperation{
+			Op:    "add",
```
matthyx, line 80: did you test if the key already exists?

pkg/utils/applicationprofile.go
```diff
@@ -138,3 +162,24 @@ func GetApplicationProfileContainer(object *v1beta1.ApplicationProfile, containe
 	}
 	return nil
 }
+
+func MergePolicy(policy1, policy2 v1beta1.RulePolicy) v1beta1.RulePolicy {
```
matthyx, line 166: I am sure you can do better than this...

matthyx (changes requested):
few comments.. at least use the released storage

pkg/containerwatcher/v1/container_watcher.go
afek854: I see that we do it in `container_watcher`  for almost every tracer.
It will require another interface to do it otherwise and I did not want to do it

go.mod
```diff
@@ -262,3 +262,5 @@ require (
 replace github.com/vishvananda/netns => github.com/inspektor-gadget/netns v0.0.5-0.20230524185006-155d84c555d6
 
 replace github.com/goradd/maps => github.com/matthyx/maps v0.0.0-20241029072232-2f5d83d608a7
+
+replace github.com/kubescape/storage => github.com/kubescape/storage v0.0.128-0.20241104095737-2142149b0f10
```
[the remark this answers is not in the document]
afek854: DONE

pkg/utils/applicationprofile.go
afek854: DONE

pkg/utils/applicationprofile.go
```diff
@@ -163,21 +163,21 @@ func GetApplicationProfileContainer(object *v1beta1.ApplicationProfile, containe
...
+		processes.Add(process)
 	}
-	for _, process := range policy2.AllowedProcesses {
-		processMap[process] = struct{}{}
+	for _, process := range secondary.AllowedProcesses {
+		processes.Add(process)
 	}
 
-	for process := range processMap {
+	for process := range processes.Iter() {
```
matthyx, line 180: I wonder if we want to sort AllowedProcesses... like we do with Syscalls https://github.com/kubescape/storage/blob/ebcd04e7b23cd48120e8a429eec8dc148a802cca/pkg/registry/file/applicationprofile_processor.go#L97

matthyx: this is to avoid sending too many new versions with the synchronizer if the content doesn't change, but just with a different order

afek854: In storage or node-agent?

matthyx: here, the reason why we prune and sort the syscalls list in storage is that this list can be appended by different node-agents
actually maybe it's the case here too...

afek854: Not sure I understand how the two cases are similar.
Rule Policies is container level and they suppose to occur rarely

pkg/applicationprofilemanager/v1/applicationprofile_manager.go
```diff
@@ -211,7 +223,7 @@ func (am *ApplicationProfileManager) monitorContainer(ctx context.Context, conta
 	}
 }
 
-func (am *ApplicationProfileManager) saveProfile(ctx context.Context, watchedContainer *utils.WatchedContainerData, namespace string) {
+func (am *ApplicationProfileManager) saveProfile(ctx context.Context, watchedContainer *utils.WatchedContainerData, namespace string, initalizeOperations []utils.PatchOperation) {
```
matthyx, line 226: this is a bit ugly, but let's leave it like that for the moment

matthyx (approved):
well done, not an easy feat !

PR #400, opened 2024-11-06
Feature/metadata

amitschendel:


go.mod
```diff
@@ -60,9 +58,9 @@ require (
 	github.com/acobaugh/osrelease v0.1.0 // indirect
 	github.com/ajstarks/svgo v0.0.0-20211024235047-1546f124cd8b // indirect
 	github.com/anchore/go-logger v0.0.0-20240217160628-ee28a485904f // indirect
-	github.com/anchore/packageurl-go v0.1.1-0.20240312213626-055233e539b4 // indirect
-	github.com/anchore/stereoscope v0.0.3-0.20240501181043-2e9894674185 // indirect
-	github.com/anchore/syft v1.4.1 // indirect
+	github.com/anchore/packageurl-go v0.1.1-0.20240507183024-848e011fc24f // indirect
+	github.com/anchore/stereoscope v0.0.3 // indirect
+	github.com/anchore/syft v1.13.0 // indirect
```
slashben, line 63: I am not 100% sure that we can upgrade `syft` version, Matthias?

main.go
```diff
@@ -263,7 +273,7 @@ func main() {
 	var malwareManager malwaremanager.MalwareManagerClient
 	if cfg.EnableMalwareDetection {
 		// create exporter
-		exporter := exporters.InitExporters(cfg.Exporters, clusterData.ClusterName, nodeName)
+		exporter := exporters.InitExporters(cfg.Exporters, clusterData.ClusterName, nodeName, cloudMetadata)
```
slashben, line 276: Food for thought: What are we not adding to the data in a way that allows Rules to access it? In theory we could make rules that take it into account

slashben (commented):
Strange, I thought that you must talk to the metadata service to get this data. If not, I am happy :)

main.go
amitschendel: In theory maybe but I don't think there is something there that is relevant for rules.

go.mod
amitschendel: @matthyx ?

matthyx: but that upgrade will come with the local SBOM PR

matthyx: ok, I'm OK with that since it's indirect at the moment

matthyx (approved):

PR #402, opened 2024-11-07
Adding cache for dns

amitschendel:


matthyx:
should we leverage `"istio.io/pkg/cache"` like in the cool down queue?

matthyx (approved):

PR #404, opened 2024-11-08
Bil

matthyx:


pkg/applicationprofilemanager/v1/applicationprofile_manager.go
```diff
@@ -243,7 +246,7 @@ func (am *ApplicationProfileManager) saveProfile(ctx context.Context, watchedCon
 	// sleep for container index second to desynchronize the profiles saving
 	time.Sleep(time.Duration(watchedContainer.ContainerIndex) * time.Second)
 
-	if droppedEvents := am.droppedEvents.Get(watchedContainer.K8sContainerID); droppedEvents {
+	if am.droppedEventsContainers.ContainsOne(watchedContainer.K8sContainerID) {
```
amitschendel, line 249: Searching in a map is faster why using a slice?

pkg/applicationprofilemanager/v1/applicationprofile_manager.go
```diff
@@ -648,11 +633,10 @@ func (am *ApplicationProfileManager) ContainerCallback(notif containercollection
 		if am.watchedContainerChannels.Has(notif.Container.Runtime.ContainerID) {
 			return
 		}
-		am.savedCapabilities.Set(k8sContainerID, mapset.NewSet[string]())
-		am.droppedEvents.Set(k8sContainerID, false)
-		am.savedEndpoints.Set(k8sContainerID, new(maps.SafeMap[string, *v1beta1.HTTPEndpoint]))
-		am.savedExecs.Set(k8sContainerID, new(maps.SafeMap[string, []string]))
-		am.savedOpens.Set(k8sContainerID, new(maps.SafeMap[string, mapset.Set[string]]))
+		am.savedCapabilities.Set(k8sContainerID, cache.NewTTL(5*am.cfg.UpdateDataPeriod, am.cfg.UpdateDataPeriod))
```
amitschendel, line 636: The cache should be as long as the learning period no?

amitschendel: Maybe we need a logic instead of expiring cache that when the learning period is over we delete these maps.

pkg/applicationprofilemanager/v1/applicationprofile_manager.go
matthyx: This is a set not a slice

pkg/applicationprofilemanager/v1/applicationprofile_manager.go
matthyx: We already have this logic, see the method where I added the missing endpoint maps. The idea here is to avoid keeping in memory files that we access once and we already sent to storage. The whole point of these saved maps is to avoid sending the same patches every 10 minutes to storage, at the expense of more memory

matthyx: Initially we didn't have them, but David feared we're putting too much pressure on storage

amitschendel (approved):

PR #406, opened 2024-11-13
Fixing socket path

amitschendel:


matthyx:
@amitschendel I think we can close this?

PR #410, opened 2024-11-21
Improving rule r0006

amitschendel:

### DIFF
--- a/pkg/ruleengine/v1/r0006_unexpected_service_account_token_access.go
+++ b/pkg/ruleengine/v1/r0006_unexpected_service_account_token_access.go
@@ -11,7 +11,6 @@ import (
 
 	apitypes "github.com/armosec/armoapi-go/armotypes"
 	traceropentype "github.com/inspektor-gadget/inspektor-gadget/pkg/gadgets/trace/open/types"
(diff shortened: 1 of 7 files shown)


pkg/ruleengine/v1/r0006_unexpected_service_account_token_access.go
```diff
@@ -42,15 +41,68 @@ var R0006UnexpectedServiceAccountTokenAccessRuleDescriptor = ruleengine.RuleDesc
...
+		if strings.HasPrefix(path, prefix) {
+			return prefix
+		}
+	}
+	return ""
+}
+
+// normalizeTokenPath removes timestamp directories from the path while maintaining
+// the essential structure. Optimized for minimal allocations.
+func normalizeTokenPath(path string) string {
```
matthyx, line 62: do you have a time constraint on this? because splitting and joining strings are really wasteful
you can check how much fun we had with indexing strings in https://github.com/kubescape/storage/blob/2b75b33130a55255df7575dc473a49858e3e325d/pkg/registry/file/dynamicpathdetector/analyzer.go#L137

matthyx (dismissed):
just check the comment regarding performance

PR #432, opened 2024-12-11
Adding support for non unix prefixes

amitschendel:


matthyx (approved):

ggerrit:
Tested the change on an Openshift 4.16 cluster: worked like a charm.

PR #437, opened 2024-12-17
add IG config for default socketpaths

matthyx:


build/Dockerfile
```diff
@@ -9,9 +9,10 @@ RUN --mount=target=. \
     --mount=type=cache,target=/go/pkg \
     GOOS=$TARGETOS GOARCH=$TARGETARCH go build -o /out/node-agent .
 
-FROM gcr.io/distroless/static-debian12:latest
+FROM gcr.io/distroless/static-debian12:nonroot
```
matthyx, line 12: like in the private one, we don't need root (apparently)

matthyx: I think it's better to keep root, as we have `runAsUser: 0`

amitschendel (approved):

PR #444, opened 2024-12-21
Improving the way we fetch the api server address

amitschendel:


matthyx (approved):
Do we have to set this in the chart?

amitschendel:
> Do we have to set this in the chart?

No Kubernetes sets this for every container.
[More information](https://stackoverflow.com/questions/45759205/what-environment-variables-are-created-in-kubernetes-by-default).

PR #458, opened 2025-01-15
Feature/pod informer

amitschendel:


pkg/containerwatcher/v1/container_watcher_private.go
```diff
@@ -111,14 +111,12 @@ func (ch *IGContainerWatcher) startContainerCollection(ctx context.Context) erro
 		// Enrich events with Linux namespaces information, it is needed for per container filtering
 		containercollection.WithLinuxNamespaceEnrichment(),
 
-		// Get containers created with container runtimes
-		containercollection.WithContainerRuntimeEnrichment(ch.runtime),
```
matthyx, line 115: this is the one that should set the correct imageID and imageTag

matthyx (approved):

PR #460, opened 2025-01-16
Unify container watcher api calls

amitschendel:

### DIFF
--- a/pkg/applicationprofilemanager/v1/applicationprofile_manager_test.go
+++ b/pkg/applicationprofilemanager/v1/applicationprofile_manager_test.go
@@ -3,8 +3,6 @@ package applicationprofilemanager
 import (
 	"context"
 	"encoding/json"
-	"net/http"
-	"net/url"
 	"sort"
 	"strings"
 	"testing"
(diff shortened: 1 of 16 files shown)


pkg/objectcache/k8scache/k8scache.go
```diff
@@ -71,6 +71,16 @@ func (k *K8sObjectCacheImpl) GetPodStatus(namespace, podName string) *corev1.Pod
 	return nil
 }
 
+// GetPod returns the pod for the given namespace and pod name, if not found returns nil
+func (k *K8sObjectCacheImpl) GetPod(namespace, podName string) *corev1.Pod {
+	p := podKey(namespace, podName)
+	if k.pods.Has(p) {
```
matthyx, line 77: TOCTOU?

amitschendel: Like all of our code

pkg/applicationprofilemanager/v1/applicationprofile_manager_test.go
```diff
@@ -3,29 +3,82 @@ package applicationprofilemanager
...
 	"github.com/kubescape/node-agent/pkg/seccompmanager"
 	"github.com/kubescape/node-agent/pkg/storage"
+	"github.com/kubescape/node-agent/pkg/utils"
 	"github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1"
 	"github.com/kubescape/storage/pkg/registry/file/dynamicpathdetector"
 	"github.com/stretchr/testify/assert"
 	istiocache "istio.io/pkg/cache"
 )
 
+func ensureInstanceID(container *containercollection.Container, watchedContainer *utils.WatchedContainerData, k8sclient *k8sclient.K8sClientMock, clusterName string) error {
```
matthyx, line 35: why don't you use the new methods like setSharedWatchedContainerData here?

pkg/containerwatcher/v1/container_watcher_private.go
```diff
@@ -69,7 +76,93 @@ func (ch *IGContainerWatcher) containerCallback(notif containercollection.PubSub
...
+	// fill container type, index and names
+	if watchedContainer.ContainerType == utils.Unknown {
+		if err := watchedContainer.SetContainerInfo(pod, container.K8s.ContainerName); err != nil {
+			return nil, fmt.Errorf("failed to set container info: %w", err)
+		}
+	}
+	// get pod template hash
+	watchedContainer.TemplateHash, _ = pod.GetPodLabel("pod-template-hash")
+	// find parentWlid
+	kind, name, err := ch.k8sClient.CalculateWorkloadParentRecursive(pod)
```
matthyx, line 132: ok so this is level 1 optimization, in a later PR we will rewrite this with calling workload.GetOwnerReferences() once and use the slice in the k8sapi shit

matthyx (approved):
just not sure why you moved the old method in the test, I would have expected to use the new ones

pkg/applicationprofilemanager/v1/applicationprofile_manager_test.go
amitschendel: I am using but I don't have the logic of filling the SharedData because it happens in the ContainerWatcher.

PR #462, opened 2025-01-20
using ContainerStartedAt to detect prerunning containers

matthyx:
wondering what you think of this


pkg/containerwatcher/v1/container_watcher_private.go
```diff
@@ -427,14 +392,6 @@ func (ch *IGContainerWatcher) unregisterContainer(container *containercollection
 	logger.L().Debug("IGContainerWatcher - stopping to monitor on container", helpers.String("container ID", container.Runtime.ContainerID), helpers.String("namespace", container.K8s.Namespace), helpers.String("PodName", container.K8s.PodName), helpers.String("ContainerName", container.K8s.ContainerName))
 
 	ch.containerCollection.RemoveContainer(container.Runtime.ContainerID)
-
-	// TODO: I dont think we need the following code ->
```
matthyx, line 431: @amitschendel we have to test this

pkg/applicationprofilemanager/v1/applicationprofile_manager.go
```diff
@@ -607,6 +609,7 @@ func (am *ApplicationProfileManager) startApplicationProfiling(ctx context.Conte
 		SeccompProfilePath:     am.k8sObjectCache.GetSharedContainerData(container.Runtime.ContainerID).SeccompProfilePath,
 		ContainerType:          am.k8sObjectCache.GetSharedContainerData(container.Runtime.ContainerID).ContainerType,
 		ContainerIndex:         am.k8sObjectCache.GetSharedContainerData(container.Runtime.ContainerID).ContainerIndex,
+		PreRunningContainer:    preRunning,
```
amitschendel, line 612: Why not setting this also on the ContainerWatcher in the SharedContainerData instead of doing it in each manager?

pkg/networkmanager/v2/network_manager.go
```diff
@@ -66,7 +66,7 @@ func CreateNetworkManager(ctx context.Context, cfg config.Config, clusterName st
 		trackedContainers:       mapset.NewSet[string](),
 		removedContainers:       mapset.NewSet[string](),
 		droppedEventsContainers: mapset.NewSet[string](),
-		preRunningContainerIDs:  preRunningContainerIDs,
+		startTime:               time.Now(),
```
amitschendel, line 69: Same comment as the application profile manager

amitschendel (changes requested):

pkg/containerwatcher/v1/container_watcher_private.go
amitschendel: I think we don't need it.

amitschendel (approved):

PR #466, opened 2025-01-24
use cri runtime for enrichment to get pod labels

matthyx:


pkg/utils/utils.go
```diff
@@ -259,9 +260,10 @@ func (watchedContainer *WatchedContainerData) SetContainerInfo(wl workloadinterf
 	checkContainers := func(containers iter.Seq2[int, v1.Container], containerStatuses []v1.ContainerStatus, containerType ContainerType) error {
 		var containersInfo []ContainerInfo
 		for i, c := range containers {
+			normalizedImageName := NormalizeImageName(c.Image) // FIXME missing tag !!!!
```
amitschendel, line 263: Is this comment still relevant? @matthyx

amitschendel (approved):

pkg/containerwatcher/v1/container_watcher_private.go
```diff
@@ -104,7 +104,7 @@ func (ch *IGContainerWatcher) getSharedWatchedContainerData(container *container
 	watchedContainer := utils.WatchedContainerData{
 		ContainerID: container.Runtime.ContainerID,
 		ImageID:     container.Runtime.ContainerImageDigest,
-		ImageTag:    container.Runtime.ContainerImageName,
+		// we get ImageTag from the pod spec for consistency between different runtimes
```
amitschendel, line 107: I don't see you fill it, is it only used for the sbom manager?

pkg/utils/utils.go
matthyx: indeed, we can remove it

pkg/containerwatcher/v1/container_watcher_private.go
matthyx: we fill it in https://github.com/kubescape/node-agent/pull/466/files#diff-81ddbadfb415ccbb9c7af84f11668d1aa5e53c34025bf86d4702f16b4e42f045R275

PR #525, opened 2025-04-08
split ContainerCallback into sync and async parts

matthyx:


amitschendel:
@matthyx How about instead of creating a worker pool in each manager, which is something that is hard to maintain during development, we will create one worker pool in the container watcher that takes a list of callbacks and streams the container notifications as today?

matthyx:
> @matthyx How about instead of creating a worker pool in each manager, which is something that is hard to maintain during development, we will create one worker pool in the container watcher that takes a list of callbacks and streams the container notifications as today?

To be discussed, as we'll lose the parallel processing, so one callback type could slow down the others

amitschendel:
> > @matthyx How about instead of creating a worker pool in each manager, which is something that is hard to maintain during development, we will create one worker pool in the container watcher that takes a list of callbacks and streams the container notifications as today?
> 
> To be discussed, as we'll lose the parallel processing, so one callback type could slow down the others

Why? you can queue the callbacks in go routines.

matthyx:
ok, let me think and propose another approach

matthyx:
do you like it better https://github.com/kubescape/node-agent/pull/533 ?

matthyx:
superseded by #533

PR #532, opened 2025-04-14
only process events from known pods

matthyx:


pkg/rulemanager/v1/rule_manager.go
```diff
@@ -207,7 +207,7 @@ func (rm *RuleManager) ContainerCallback(notif containercollection.PubSubEvent)
 		go rm.startRuleManager(notif.Container, k8sContainerID)
 	case containercollection.EventTypeRemoveContainer:
 		rm.trackedContainers.Remove(k8sContainerID)
-		rm.podToWlid.Delete(utils.CreateK8sPodID(notif.Container.K8s.Namespace, notif.Container.K8s.PodName))
+		//rm.podToWlid.Delete(utils.CreateK8sPodID(notif.Container.K8s.Namespace, notif.Container.K8s.PodName)) // TODO: check if we want to remove the pod from the map
```
matthyx, line 210: @afek854 I think we should not remove the pod here, because we might still have some events to process (race)
Also, we're extrapolating the pod from just one container (I'm sure we have other places like that).

matthyx: but then... we'll slowly leak memory (not sure it's a big deal)

amitschendel: Hmm it's a scary leak don't you think? maybe schedule a go routine of 10 minutes and than delete?

matthyx: Good idea, I'll do that

amitschendel (approved):

PR #533, opened 2025-04-15
split ContainerCallback into sync and async parts

matthyx:
@amitschendel better like that?


amitschendel:
Yes!
Is there a way to scale the worker pool dynamically instead of having to bump it manually?

amitschendel:
This is better, sorry for being annoying but can you put a comment above the callback list that if there is more then 10 we should change the configuration?

matthyx:
> This is better, sorry for being annoying but can you put a comment above the callback list that if there is more then 10 we should change the configuration?

done

amitschendel (approved):

PR #545, opened 2025-04-29
fixed rule manager race condition

afek854:
When the node-agent crash while in learning period, it does not update the AP status. (It got stuck as Ready).
- To reproduce, learning period (2m) -> crash -> status not changing

Explanation:
```
Container Restart Sequence:
--------------------------

Time      ContainerCallback          monitorContainer         deleteResources         saveProfile
----      ----------------           ---------------          --------------          -----------
t0        AddContainer               
t1                                   Start monitoring
t2                                   SetStatus(Initializing)
t3                                                                                   First save (Initializing)
t4        RemoveContainer
t5                                   ← Gets error signal
t6                                   SetStatus(Completed)
t7                                                                                   Final save (Completed)
t8                                   Exits                     
t9                                                            Deletes resources
t10       AddContainer (new)      
t11                                  Start monitoring (new)
t12                                  SetStatus(Initializing)
t13                                                                                  First save (Initializing)
t14       [RACE CONDITION]: If t9 (deleteResources) happens after t10-t13, it wipes out the tracking data for the new container
t15                                  Ticker fires, but tracking data is gone
t16                                  Cannot transition to Ready
```


pkg/applicationprofilemanager/v1/applicationprofile_manager.go
```diff
@@ -708,6 +720,10 @@ func (am *ApplicationProfileManager) ContainerCallback(notif containercollection
 		if am.watchedContainerChannels.Has(notif.Container.Runtime.ContainerID) {
 			return
 		}
+		// Set a new version for this container (current timestamp)
+		containerVersion := time.Now().UnixNano()
```
amitschendel, line 724: Use the container start time from the Runtime metadata

matthyx:
good idea, let's continue in this direction

PR #549, opened 2025-05-04
Fix ap updates

afek854:


pkg/storage/v1/applicationprofile.go
```diff
@@ -49,6 +49,14 @@ func (sc Storage) PatchApplicationProfile(name, namespace string, operations []u
 	return nil
 }
 
+func (sc Storage) ListApplicationProfiles(namespace string) (*v1beta1.ApplicationProfileList, error) {
```
matthyx, line 52: do we want to use a pager here? (in general I mean, since we can have a lot of APs in the cluster)

matthyx:
superseded by #551

PR #560, opened 2025-05-13
Adding initial code

amitschendel:


pkg/ruleengine/v1/r0002_unexpected_file_access.go
```diff
@@ -161,6 +162,14 @@ func (rule *R0002UnexpectedFileAccess) ProcessEvent(eventType utils.EventType, e
...
 	}
 
+	profileMetadata := &apitypes.ProfileMetadata{
+		Status:             ap.GetAnnotations()[helpersv1.StatusMetadataKey],
+		Completion:         ap.GetAnnotations()[helpersv1.CompletionMetadataKey],
+		Name:               ap.Name,
+		Type:               apitypes.ApplicationProfile,
+		IsProfileDependent: true,
+	}
+
```
slashben, line 172: Wouldn't it make sense to do it as part of a function the pre-populate `BaseRuntimeAlert` for every rule? This would reduce the line of code

PR #564, opened 2025-05-18
Refactor profiles cache to be more stable and simple

amitschendel:


pkg/objectcache/applicationprofilecache/applicationprofilecache.go
```diff
@@ -55,91 +36,252 @@ type ContainerCallStackIndex struct {
...
-		slugToAppProfile:    maps.SafeMap[string, *v1beta1.ApplicationProfile]{},
-		slugToContainers:    maps.SafeMap[string, mapset.Set[string]]{},
-		slugToState:         maps.SafeMap[string, applicationProfileState]{},
-		allProfiles:         mapset.NewSet[string](),
-		userManagedProfiles: maps.SafeMap[string, *v1beta1.ApplicationProfile]{},
-		containerCallStacks: maps.SafeMap[string, *ContainerCallStackIndex]{},
-		containerToName:     maps.SafeMap[string, string]{},
+// NewApplicationProfileCache creates a new application profile cache with periodic updates
+func NewApplicationProfileCache(cfg config.Config, storageClient versioned.SpdxV1beta1Interface, k8sObjectCache objectcache.K8sObjectCache) *ApplicationProfileCacheImpl {
+	updateInterval := utils.AddJitter(cfg.ProfilesCacheRefreshRate, 10) // Add 10% jitter to avoid high load on the storage
```
slashben, line 53: It would be much better to add the jitter to the global frequency (setting it per node-agent instance), but in between loops! In this case the the agent will have a 10% global frequency difference over time, I am not sure you wanted this. (think about it, if a node agent runs for 30 days, it might did 10% less updates you meant it to do)

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
```diff
@@ -55,91 +36,252 @@ type ContainerCallStackIndex struct {
...
 
-	// Store the user-managed profile temporarily
-	ap.userManagedProfiles.Set(baseProfileUniqueName, appProfile)
+	for {
+		select {
+		case <-ticker.C:
+			apc.updateAllProfiles(ctx)
+		case <-ctx.Done():
+			return
+		}
```
slashben, line 85: I don't know ticker. Is it running independently from this function? I explain why I am asking. If the `updateInterval` is set to 10s, and the `updateAllProfiles` takes more than 10s (for some reason the API server slow), can bad things happen? 
Usually, it is better to make this kind of tasks is a way that they cannot crowd.

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
```diff
@@ -55,91 +36,252 @@ type ContainerCallStackIndex struct {
...
+			// Update the profile in the cache
+			if existingProfile, exists := apc.workloadIDToProfile.Load(workloadID); exists {
+				// If the profile already exists and it's complete/completed, continue to the next one
+				if existingProfile.Annotations[helpersv1.CompletionMetadataKey] == helpersv1.Complete {
+					continue
+				}
+
+				// If the new profile is not complete and we already have a completed/partial one, skip it
+				if profile.Annotations[helpersv1.CompletionMetadataKey] != helpersv1.Complete {
+					continue
```
slashben, line 149: I am confused here. We check at line 141 that we already have a completed profile, but we do not check if it was partial or complete. In theory, we prevent the update of any partially completed profile if we have a completed profile in the cache. Was this the purpose?

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
```diff
@@ -55,91 +36,252 @@ type ContainerCallStackIndex struct {
...
-	slugToContainers    maps.SafeMap[string, mapset.Set[string]]          // cache the containerIDs that belong to the application profile, this will enable removing from cache AP without pods
-	slugToState         maps.SafeMap[string, applicationProfileState]     // cache the containerID to slug mapping, this will enable a quick lookup of the application profile
-	storageClient       versioned.SpdxV1beta1Interface
-	allProfiles         mapset.Set[string] // cache all the application profiles that are ready. this will enable removing from cache AP without pods that are running on the same node
-	userManagedProfiles maps.SafeMap[string, *v1beta1.ApplicationProfile]
-	containerCallStacks maps.SafeMap[string, *ContainerCallStackIndex] // cache the containerID to call stack search tree mapping
-	containerToName     maps.SafeMap[string, string]                   // cache the containerID to container name mapping
+	cfg                            config.Config
+	workloadIDToProfile            maps.SafeMap[string, *v1beta1.ApplicationProfile]
+	containerIDToInfo              maps.SafeMap[string, *ContainerInfo]
```
slashben, line 41: You need to create a struct that includes v1beta1.ApplicationProfile instead of directly referencing it here. The struct can hold info about the current status of the application profile, like error states, learning period (ready/initializing). It is important for the reasons I have told you. You need to update this status in the periodic updates.

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
```diff
@@ -55,91 +36,252 @@ type ContainerCallStackIndex struct {
...
+
+	// Create a unique tracking key for this user profile
+	profileKey := fmt.Sprintf("%s/%s", profile.Namespace, normalizedProfileName)
+
+	// Check if we've already processed this exact version of the user-managed profile
+	if storedIdentifier, exists := apc.profileToUserManagedIdentifier.Load(profileKey); exists &&
+		storedIdentifier == userManagedProfileUniqueIdentifier {
+		return
+	}
+
```
slashben, line 199: Can we have more than one user-managed profile for a given workload? I don't know if it works.

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
```diff
@@ -55,91 +36,252 @@ type ContainerCallStackIndex struct {
...
+// NewApplicationProfileCache creates a new application profile cache with periodic updates
+func NewApplicationProfileCache(cfg config.Config, storageClient versioned.SpdxV1beta1Interface, k8sObjectCache objectcache.K8sObjectCache) *ApplicationProfileCacheImpl {
+	updateInterval := utils.AddJitter(cfg.ProfilesCacheRefreshRate, 10) // Add 10% jitter to avoid high load on the storage
+
+	apc := &ApplicationProfileCacheImpl{
+		cfg:                       cfg,
+		workloadIDToProfile:       maps.SafeMap[string, *v1beta1.ApplicationProfile]{},
+		containerIDToInfo:         maps.SafeMap[string, *ContainerInfo]{},
+		namespaceToContainers:     maps.SafeMap[string, mapset.Set[string]]{},
+		containerToCallStackIndex: maps.SafeMap[string, *ContainerCallStackIndex]{},
```
slashben, line 60: SafeMap is not optimal for our use case. Our map is barely changing (low frequency writes) and it is used for reading many times (per eBPF event). Using `sync.Map` is better. Even better is a copy on write map which avoids using mutexes on read:
```
type COWMap struct {
    mu    sync.Mutex
    m     atomic.Value // holds a map[string]interface{}
}

func NewCOWMap() *COWMap {
    cm := &COWMap{}
    cm.m.Store(make(map[string]interface{}))
    return cm
}

// Get requires no locking - extremely fast for reads
func (cm *COWMap) Get(key string) (interface{}, bool) {
    m := cm.m.Load().(map[string]interface{})
    val, ok := m[key]
    return val, ok
}

// Set creates a complete copy of the map for writing
func (cm *COWMap) Set(key string, val interface{}) {
    cm.mu.Lock()
    defer cm.mu.Unlock()
    
    m := cm.m.Load().(map[string]interface{})
    newMap := make(map[string]interface{}, len(m)+1)
    
    // Copy all data from the current map
    for k, v := range m {
        newMap[k] = v
    }
    
    // Update the value
    newMap[key] = val
    
    // Atomically replace the map
    cm.m.Store(newMap)
}
```

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
amitschendel: Done

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
amitschendel: Yes, if we hold in cache partial/completed I want to update if we get complete/completed. But if I have complete/completed I don't want to update.

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
amitschendel: We can't have more than one.

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
```diff
@@ -163,249 +352,157 @@ func (ap *ApplicationProfileCacheImpl) indexContainerCallStacks(containerID, con
...
-		if ap.slugToAppProfile.Has(apName) {
-			ap.slugToAppProfile.Delete(apName)
-			ap.allProfiles.Remove(apName)
-		}
-		return
+// ContainerCallback handles container lifecycle events
+func (apc *ApplicationProfileCacheImpl) ContainerCallback(notif containercollection.PubSubEvent) {
+	switch notif.Type {
+	case containercollection.EventTypeAddContainer:
+		go func() {
```
matthyx, line 369: now that we have the async calls, I don't think you need the goroutine here

pkg/utils/utils.go
```diff
@@ -162,6 +162,17 @@ func AddJitter(duration time.Duration, maxJitterPercentage int) time.Duration {
 	return duration * time.Duration(jitter)
 }
 
+// Jitter returns a random duration
+func Jitter(duration *time.Duration, maxJitterPercentage float64) {
```
matthyx, line 166: I don't see why we have so many ways of dealing with durations... AddJitter, Jitter, RandomDuration, ...

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
```diff
@@ -163,249 +352,157 @@ func (ap *ApplicationProfileCacheImpl) indexContainerCallStacks(containerID, con
...
 
-func (ap *ApplicationProfileCacheImpl) GetCallStackSearchTree(containerID string) *callstackcache.CallStackSearchTree {
-	if index := ap.containerCallStacks.Get(containerID); index != nil {
-		return index.searchTree
+	// Create container info
+	containerInfo := &ContainerInfo{
+		ContainerID: containerID,
+		WorkloadID:  workloadID,
+		Namespace:   namespace,
+		Name:        container.Runtime.ContainerName,
```
matthyx, line 405: funny that you don't need this in the NN cache

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
```diff
@@ -55,91 +37,296 @@ type ContainerCallStackIndex struct {
...
+			// Set the flag and release the lock
+			apc.updateInProgress = true
+			apc.updateMutex.Unlock()
 
-	// If we have the base profile cached, fetch a fresh copy and merge.
-	// If the base profile is not cached yet, the merge will be attempted when it's added.
-	if ap.slugToAppProfile.Has(baseProfileUniqueName) {
-		// Fetch fresh base profile from cluster
-		freshBaseProfile, err := ap.getApplicationProfile(appProfile.GetNamespace(), baseProfileName)
+			// Run the update in a goroutine
```
matthyx, line 100: why in a goroutine?

matthyx: this is maybe why we are seeing an increase in memory when we run every 5s

pkg/objectcache/applicationprofilecache/applicationprofilecache_test.go
```diff
@@ -1,937 +0,0 @@
-package applicationprofilecache
```
matthyx, line 1: too bad we no longer have test coverage...

matthyx (commented):
Few questions to answer first.
I see why you haven't merged the 2 cache implementation, however I think we can still do it in a clever way to make sure we don't have bugs that we fix in one and not the other.
It would also be nice to have test coverage...

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
amitschendel: So we can track long/stuck profile updates.

pkg/utils/utils.go
amitschendel: The other functions contains bugs :(

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
amitschendel: You are probably right, but, I think it's a good habit to not have blocking things in the ContainerCallback.

amitschendel: Removed.

matthyx (approved):
let's merge like that, I'll open a new  PR for the DRY version

PR #569, opened 2025-05-25
Adding validation on template hash to verify we don't take old profiles

amitschendel:

### DIFF
--- a/pkg/objectcache/applicationprofilecache/applicationprofilecache.go
+++ b/pkg/objectcache/applicationprofilecache/applicationprofilecache.go
@@ -25,10 +25,11 @@ import (
 
 // ContainerInfo holds container metadata we need for application profile mapping
 type ContainerInfo struct {
-	ContainerID string
-	WorkloadID  string
-	Namespace   string
-	Name        string
+	ContainerID          string
+	WorkloadID           string
+	InstanceTemplateHash string
+	Namespace            string
+	Name                 string
 }
 
 // ContainerCallStackIndex maintains call stack search trees for a container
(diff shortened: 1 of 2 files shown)


pkg/objectcache/applicationprofilecache/applicationprofilecache.go
```diff
@@ -115,10 +116,25 @@ func (apc *ApplicationProfileCacheImpl) periodicUpdate(ctx context.Context) {
...
+			continue // Skip to next namespace
+		}
+
 		// Skip empty namespaces
 		if containerSet.Cardinality() == 0 {
-			return true
+			continue
 		}
 
 		// Get profiles list for this namespace
```
matthyx, line 140: instead of calling per namespace, I would use the [pager](https://github.com/kubernetes/client-go/blob/master/tools/pager/pager.go) as it is made for that and it will avoid bigger or smaller list calls by setting the page size

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
```diff
@@ -160,9 +176,10 @@ func (apc *ApplicationProfileCacheImpl) updateAllProfiles(ctx context.Context) {
 
 			// Check if this workload ID is used by any container in this namespace
 			workloadIDInUse := false
-			for containerID := range containerSet.Iter() {
+			for _, containerID := range containerSet.ToSlice() {
 				if containerInfo, exists := apc.containerIDToInfo.Load(containerID); exists &&
```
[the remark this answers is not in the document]
afek854: It causes deadlock

pkg/objectcache/applicationprofilecache/applicationprofilecache.go
matthyx: ok now I understand that we use NS as shards to minimize the issue of the linear lookups on containerSet

matthyx (approved):

PR #603, opened 2025-08-06
removed batch from process queue

afek854:


matthyx:
@afek854 do we still need that?

PR #654, opened 2025-10-09
Adding initial eBPF refactor

amitschendel:


pkg/utils/datasource_event.go
```diff
@@ -45,8 +78,20 @@ func (e *DatasourceEvent) GetArgs() []string {
 	switch e.EventType {
 	case ExecveEventType:
 		args, _ := e.Datasource.GetField("args").String(e.Data)
-		return strings.Split(args, " ")
+		return strings.Split(args, igconsts.ArgsSeparator)
```
matthyx: well spotted... I should have known this would happen after taking my PR into IG

PR #670, opened 2025-12-04
chore: Update Dockerfile

pfarikrispy:
## Overview
- don't use Ubuntu and install python, use python images
- update python to 3.14 (latest)
- cleanup after apt-get
- install Debian pkg for `netcat` which is `netcat-traditional`
- reorder RUN and COPY lines
- end with non root user `clamav`

## Additional Information

Not sure if clamav is still used nor what for. Wanted to show some improvements are possible.
clamav latest is 1.5.1 so a rebuilt will bring many current updates

not sure if python-pip needs to be installed in python image. This would further improve build time since it installs even less that changes less often

### DIFF
--- a/clamav/Dockerfile
+++ b/clamav/Dockerfile
@@ -1,16 +1,22 @@
-ARG CLAMAV_VERSION
-
-FROM --platform=$BUILDPLATFORM ubuntu:23.10 as builder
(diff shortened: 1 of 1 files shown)


matthyx:
it's deprecated... but I'll let @amitschendel have a look later

PR #690, opened 2026-01-20
fix a bug where failed expressions would recompile on every event

YakirOren:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of expression evaluation by caching failed compilations to avoid repeated work and noisy errors.
  * Added safeguards so failed or missing expressions are skipped safely and return empty results instead of causing failures.
  * Improved logging for compilation/evaluation issues to aid diagnosis without affecting runtime behavior.


matthyx (changes requested):
please check https://github.com/kubescape/node-agent/pull/690/changes#r2709290176

PR #716, opened 2026-02-10
optimize CEL rule evaluation performance

YakirOren:
## Summary
- Reorder CreateEventFromRequest to check packet direction before expensive HTTP parsing
- Eliminate full StructEvent copy in MakeHttpEvent by mutating in-place
- Extract isPrivateIP helper, removing inline IIFE and redundant net.ParseIP closures
- Replace map-based GetPacketDirection with switch statement

## Summary by CodeRabbit

* **New Features**
  * Added parse.basename() for extracting filenames.
  * New pooled, context-based evaluation API for faster, lower-allocation rule and expression evaluation; single-call wrappers retained.

* **Refactor**
  * Pre-computed expression indexing by event type and single reused context per event to speed evaluations.
  * Improved HTTP event / payload access with allocation-free wrappers.

* **Bug Fixes**
  * Avoided caching of runtime evaluation errors; preserved nil-on-compilation-failure behavior.

* **Tests**
  * Added basename edge-case tests.

* **Chores**
  * Ensure rules are initialized when registered or updated.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -177,7 +177,7 @@ func main() {
 	if cfg.EnablePrometheusExporter {
 		prometheusExporter = metricprometheus.NewPrometheusMetric()
 	} else {
-		prometheusExporter = metricsmanager.NewMetricsMock()
+		prometheusExporter = metricsmanager.NewMetricsNoop()
 	}
 
 	// Create watchers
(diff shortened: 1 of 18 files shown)


matthyx (changes requested):
did you run a Bench to see the improvements? maybe we should commit it too...

PR #784, opened 2026-04-21
add process-name exclude prefilter options

YakirOren:
## Summary by CodeRabbit

* **New Features**
  * Added process and parent-process exclusion filters to pre-filters that require both process name and executable path to match (paths are normalized); exec events now include parent process context so filters can use it.
* **Tests**
  * Added/expanded tests covering parsing, normalization, invalid-entry handling, and runtime skip behavior for the new exclusion settings.

### DIFF
--- a/pkg/rulemanager/prefilter/prefilter.go
+++ b/pkg/rulemanager/prefilter/prefilter.go
@@ -70,15 +70,20 @@ func methodToBit(method string) MethodMask {
 }
 
 // EventFields holds event data extracted once per event for pre-filtering.
-// Passed by value (stack-allocated, ~28 bytes) — extracted once before the
-// rule loop, reused across all rules.
+// Passed by value, stack-allocated, extracted once before the rule loop and
+// reused across all rules.
 type EventFields struct {
-	Path         string     // file/exec path (empty if not applicable)
-	DstPort      uint16     // destination port from network/SSH event
-	Dir          Direction  // pre-computed from HTTP direction string
-	MethodBit    MethodMask // pre-computed from HTTP method string
-	PortEligible bool       // true for SSH/network events (port filter applies)
-	Extracted    bool       // true after extractEventFields has run
+	Path string // file path (open) or exe path (exec); empty otherwise
+	// Exec-only fields; empty for other event types.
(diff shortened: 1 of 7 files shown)


pkg/rulemanager/prefilter/prefilter.go
```diff
@@ -176,16 +197,53 @@ func ParseWithDefaults(ruleState map[string]any, bindingParams map[string]any) *
...
+	}
+	m := make(map[processKey]struct{}, len(entries))
+	dropped := 0
+	for _, entry := range entries {
+		if entry.Name == "" || entry.Path == "" {
+			dropped++
+			continue
+		}
+		m[processKey{Name: entry.Name, Path: entry.Path}] = struct{}{}
+	}
```
matthyx, line 234: **Suggestion: normalize the config-side `Path` to match the event side.**

The event-side `ParentExePath` / `ExePath` are run through `utils.NormalizePath` before being placed in `EventFields` (see `pkg/utils/struct_event.go` and `pkg/utils/datasource_event.go`). The map lookup in `ShouldSkip` is exact-match, so any operator-supplied path that doesn't already match what `NormalizePath` produces will silently never fire — e.g. a trailing slash (`/usr/bin/dockerd/`), a missing leading `/`, or a bare `1234/exe` that the event layer would rewrite to `/proc/1234/exe`.

This mirrors what `trimTrailingSlash` does for `IgnorePrefixes` / `IncludePrefixes`, but `NormalizePath` is stricter and matches what the runtime actually compares against.

```suggestion
	for _, entry := range entries {
		if entry.Name == "" || entry.Path == "" {
			dropped++
			continue
		}
		m[processKey{Name: entry.Name, Path: utils.NormalizePath(entry.Path)}] = struct{}{}
	}
```

(Requires adding `"github.com/kubescape/node-agent/pkg/utils"` to the imports.)

Worth adding a unit test for the normalization, e.g. config `/usr/bin/dockerd/` → matches event `/usr/bin/dockerd`.

matthyx (commented):
One suggestion on the prefilter parsing: normalize operator-supplied paths to match what the event layer produces, so subtle config mismatches (trailing slash, missing leading slash) don't silently fail to match.

matthyx (approved):

PR #785, opened 2026-04-21
add parse.basename CEL function

YakirOren:


pkg/rulemanager/cel/libraries/parse/parse.go
```diff
@@ -25,3 +27,15 @@ func (l *parseLibrary) getExecPath(args ref.Val, comm ref.Val) ref.Val {
...
+	s, ok := path.Value().(string)
+	if !ok {
+		return types.MaybeNoSuchOverloadErr(path)
+	}
+	idx := strings.LastIndex(s, "/")
+	if idx == -1 {
+		return types.String(s)
+	}
+	return types.String(s[idx+1:])
+}
```
matthyx, line 41: Consider using `path.Base` from the stdlib instead of hand-rolling the scan:

```suggestion
func (l *parseLibrary) basename(p ref.Val) ref.Val {
	s, ok := p.Value().(string)
	if !ok {
		return types.MaybeNoSuchOverloadErr(p)
	}
	return types.String(path.Base(s))
}
```

(and swap the `strings` import for `path`)

**Behavioral differences vs. the current implementation:**

| Input        | This PR | `path.Base` |
|--------------|---------|-------------|
| `/usr/bin/nmap` | `nmap`  | `nmap`  |
| `nmap`       | `nmap`  | `nmap`  |
| `/nmap`      | `nmap`  | `nmap`  |
| `/usr/bin/`  | `""`    | `bin`   |
| `""`         | `""`    | `.`     |
| `///`        | `""`    | `/`     |

`path.Base` strips trailing slashes before taking the last segment, which matches POSIX `basename(1)` semantics that most rule authors would expect. The current impl returns an empty string for any path ending in `/`, which could silently bypass a rule like `parse.basename(event.path) == "nmap"` if the path happens to have a trailing slash.

That said — if the empty-string-on-trailing-slash behavior is intentional (e.g., to distinguish file paths from directory-like strings in your rules), this is fine as-is and worth a one-line comment explaining why.

What semantics do you want here?

PR #792, opened 2026-04-25
add first of the component tests for the new storage

entlein:
## Overview
Adds the first of the component tests to verify the OpensWildcard (of new storage) actually works e2e

put an [x] in the box to get it checked 

- [ ] My code follows the style guidelines of this project
- [ ] I have commented on my code, particularly in hard-to-understand areas
- [ ] I have performed a self-review of my code
- [ ] If it is a core feature, I have added thorough tests.

It only contains tests
 

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test suite for ApplicationProfile Opens path matching validation, including exact-match, wildcard segment, and multi-segment depth matching scenarios.
  * Introduced test deployment resources for nginx and curl workloads configured with user-defined ApplicationProfiles.
  * Validates absolute path handling requirements and accounts for profile propagation delays in alert retrieval.

### DIFF
--- a/tests/component_test.go
+++ b/tests/component_test.go
@@ -11,6 +11,7 @@ import (
 	"slices"
 	"sort"
 	"strconv"
+	"strings"
 	"testing"
 	"time"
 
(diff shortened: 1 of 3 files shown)


matthyx (commented):
Excluding the upcoming `*` open-path wildcard support as requested; adding the remaining inline suggestions.

tests/component_test.go
```diff
@@ -1569,3 +1570,380 @@ func Test_24_ProcessTreeDepthTest(t *testing.T) {
...
+		require.NoError(t, wl.WaitForReady(80), "workload not ready in ns %s", ns.Name)
+
+		// Wait for node-agent to load the user-defined profile into cache.
+		time.Sleep(10 * time.Second)
+		return wl
+	}
+
+	// triggerAndGetAlerts execs cat on the given path, then polls for alerts
+	// up to 60s to avoid race conditions with alert propagation.
+	triggerAndGetAlerts := func(t *testing.T, wl *testutils.TestWorkload, filePath string) []testutils.Alert {
```
matthyx, line 1671: This helper is shared by both the "expect alert" and "expect no alert" cases, but right now the polling only waits for `GetAlerts()` to become reachable. In practice that can return immediately with an empty list, so the positive cases still depend on the fixed `10 * time.Second` sleep below.

I think this is less flaky if `triggerAndGetAlerts` takes an `expectAlert bool` and polls for the actual condition in the positive cases:

```go
require.Eventually(t, func() bool {
	alerts, err = testutils.GetAlerts(wl.Namespace)
	return err == nil && hasAlert(alerts)
}, 60*time.Second, 5*time.Second, "expected R0002 alert in ns %s", wl.Namespace)
```

For the no-alert cases, keep the stabilization sleep/final fetch path so late alerts are still caught.

tests/component_test.go
```diff
@@ -1569,3 +1570,380 @@ func Test_24_ProcessTreeDepthTest(t *testing.T) {
...
+		addResult("exact_path_match", profilePath, filePath, false, !got,
+			fmt.Sprintf("got %d alerts, expected none for cat", len(alerts)))
+	})
+
+	// 1b-2: Exact path — profile has a DIFFERENT file => alert.
+	t.Run("exact_path_mismatch", func(t *testing.T) {
+		profilePath := "/etc/nginx/nginx.conf"
+		filePath := "/etc/hostname"
+		wl := deployWithProfile(t, []v1beta1.OpenCalls{
+			{Path: profilePath, Flags: []string{"O_RDONLY"}},
```
matthyx, line 1766: `cat` also opens `/etc/ld.so.cache` on exec. Since this case is looking for _any_ R0002 from `cat`, omitting the loader cache here can make the subtest pass for the wrong reason even if `/etc/hostname` matching regresses.

```suggestion
			{Path: profilePath, Flags: []string{"O_RDONLY"}},
			{Path: "/etc/ld.so.cache", Flags: []string{"O_RDONLY", "O_CLOEXEC"}}, // dynamic linker opens this on every exec
```

tests/component_test.go
```diff
@@ -1569,3 +1570,380 @@ func Test_24_ProcessTreeDepthTest(t *testing.T) {
...
+		allowedFiles := []string{
+			"/etc/hosts",           // covered by /etc/*
+			"/etc/resolv.conf",     // covered by /etc/*
+			"/etc/ssl/openssl.cnf", // exact match
+		}
+		for _, f := range allowedFiles {
+			stdout, stderr, err := wl.ExecIntoPod([]string{"cat", f}, "curl")
+			if err != nil {
+				t.Logf("exec 'cat %s' failed: %v (stdout=%q stderr=%q)", f, err, stdout, stderr)
+			}
```
matthyx, line 1924: If one of these `cat` commands fails, the subtest still passes as long as no alert arrives, which makes this branch vacuous. I'd fail fast here so the "allowed opens" assertion only runs after the opens were actually exercised.

```suggestion
			stdout, stderr, err := wl.ExecIntoPod([]string{"cat", f}, "curl")
			require.NoErrorf(t, err, "exec 'cat %s' in container curl failed (stdout=%q stderr=%q)", f, stdout, stderr)
```

entlein:
I think, this is outdated

PR #799, opened 2026-04-28
implement Rule-Aware Profile Projection

matthyx:
## Summary by CodeRabbit

* **New Features**
  * Profile projection: compact projected container profiles and projection spec compilation/install with immediate reprojection on spec changes
  * Rules can declare which profile fields they require
  * Config toggle to enable detailed projection/memory-savings metrics

* **Behavior Changes**
  * CEL helpers now consult projected profiles; exec/open/network checks focus on path/address (args/flags/port-protocol matching de-scoped)

* **Chores**
  * Added projection and memory-savings metrics plus no-op/mock implementations

* **Tests**
  * Many tests updated/added to cover projection, compilation, application, cache, and concurrency behavior


matthyx:
@coderabbitai review

pkg/metricsmanager/prometheus/prometheus.go
```diff
@@ -70,6 +70,29 @@ type PrometheusMetric struct {
...
+	cpHelperCallCounter                   *prometheus.CounterVec
+	cpProjectionUndeclaredRulesListGauge  *prometheus.GaugeVec
+
+	// Memory-savings metrics — detailed
+	cpProfileRawSizeHistogram         prometheus.Histogram
+	cpProfileProjectedSizeHistogram   prometheus.Histogram
+	cpProfileEntriesRawHistogram      *prometheus.HistogramVec
+	cpProfileEntriesRetainedHistogram *prometheus.HistogramVec
+	cpProfileRetentionRatioHistogram  *prometheus.HistogramVec
+
```
slashben, line 95: @matthyx aren't this "too many"? These metrics are not for free (memory perspective and we should take a second look if all of these worth the trade off

slashben (approved):

PR #806, opened 2026-05-15
allowing CIDRs, wildcards and Plural in IP and DNS

entlein:
## Overview
network-wildcards: DNS leading/trailing/mid + CIDR + IPAddresses + "*" sentinel

## Additional Information

CEL `networkneighborhood` helpers now query
projected fields with wildcard/CIDR/DNS-wildcard support and the v0.0.2
spec semantics (port/protocol checks degrade to address-only in v1
when only a CIDR is declared).

  - DNS leading-`*` (RFC 4592 single-label wildcard)
  - DNS trailing-`*` (subdomain-prefix wildcard)
  - DNS mid-`⋯` (interior dynamic segment)
  - IPv4/IPv6 literals and CIDR ranges
  - `*` any-IP sentinel
  - `IPAddresses` list (deprecated single-IP field) preserved on the
    decode path

## How to Test
20 declarative fixture YAMLs under `tests/resources/network-wildcards/`
pin the wildcard contract end-to-end:

  - Literal IPv4/IPv6
  - CIDR IPv4/IPv6
  - `*` any-IP and `*` as CIDR
  - Mixed IP list
  - Deprecated `IPAddresses` field path
  - DNS literal, leading wildcard, mid-`⋯`, trailing-`*`
  - Trailing-dot normalisation
  - Recursive `*` rejection
  - Egress-and-ingress, egress-none
  - Realistic Stripe API scenario
  - Cluster DNS via mid-`⋯`
  - Port+protocol with CIDR
  - Multi-container mixed wildcards

 Has sister PR in storage

## Release Notes

* **New Features**
  * Added wildcard and CIDR matching support for network traffic profiles, enabling more flexible IP address and DNS name definitions
  * Enabled "Unexpected Egress Network Traffic" rule detection by default

* **Tests**
  * Added comprehensive test coverage for network neighborhood matching scenarios
  * Added test fixtures validating IP/DNS wildcard and CIDR matching behavior

* **Chores**
  * Updated dependencies to latest versions

PR #808, opened 2026-05-15
profile-compaction: CollapseConfig CRD + projection overlay + user-ma…

entlein:
here things got confusing during the rebase, this is the sister PR to storage and need the signature PR (from matthyx)
 

## Summary by CodeRabbit

* **New Features**
  * Added signed profile support for application, network, seccomp, and rule objects, including a new command-line tool for signing, verifying, and extracting signatures.
  * Introduced tamper alerts for modified signed profiles and optional signature verification for synced rules.
  * Added support for user-defined network references in workload metadata.

* **Bug Fixes**
  * Improved cache handling so invalid overlays are ignored before being applied.
  * Strengthened checksum tracking so overlay changes are detected reliably.

* **Documentation**
  * Added guidance for profile signing and tamper detection.


entlein:
@coderabbitai review

Collective re-review request after applying your atomic-split advisory + rabbit-feedback fixes across the full SBOB upstream PR set. Each PR is now scoped to a single tier per your dependency map:

Tier 0 (opens):
- kubescape/storage#323 (path-wildcards core: anchored trailing-*, per-port key, WildcardIdentifier + CollapseConfig API — squash per your "merge first / or squash together" note)
- kubescape/node-agent#811 (was_path_opened* CEL helpers — NEW)

Tier 2 (execs):
- kubescape/storage#322 (CompareExecArgs)
- kubescape/storage#325 (CollapseConfig CRD + processors)
- kubescape/node-agent#805 (resolveExecPath + 3-arg parse)
- kubescape/node-agent#807 (R0040 args consumer)
- kubescape/node-agent#812 (ExecsByPath composite-key surface — NEW)

Tier 3 (network):
- kubescape/storage#324 (networkmatch package)
- kubescape/node-agent#806 (network wildcards CEL)

Tier 4 (tamper/sign):
- kubescape/node-agent#808 (projection overlay + tamper + non-blocking fanout + NN resource)
- kubescape/node-agent#809 (cosign sign/verify + tamper detection)

Umbrellas:
- kubescape/storage#326
- kubescape/node-agent#810

Please confirm atomicity per your earlier advisory and flag any remaining cross-PR coupling we missed.

pkg/objectcache/containerprofilecache/tamper_alert.go
```diff
@@ -0,0 +1,190 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/armosec/armoapi-go/armotypes"
+	"github.com/kubescape/go-logger"
+	"github.com/kubescape/go-logger/helpers"
+	"github.com/kubescape/node-agent/pkg/exporters"
+	"github.com/kubescape/node-agent/pkg/rulemanager/types"
+	"github.com/kubescape/node-agent/pkg/signature"
+	"github.com/kubescape/node-agent/pkg/signature/profiles"
```
matthyx, line 28: `go test ./...` fails on this new file because `github.com/kubescape/node-agent/pkg/signature` and `.../pkg/signature/profiles` do not exist in the checked-in module (`pkg/**/signature/**` is empty here). As written, this breaks package setup for every importer of `containerprofilecache`, so the PR will not build until these imports point at a real package or the code is wired through an existing dependency.

pkg/objectcache/containerprofilecache/projection_apply.go
```diff
@@ -191,6 +260,9 @@ func extractEgressAddresses(cp *v1beta1.ContainerProfile) []string {
 		if n.IPAddress != "" {
 			addrs = append(addrs, n.IPAddress)
 		}
+		// v0.0.2 IPAddresses[] — list form supporting CIDRs and '*' sentinel.
+		// Same semantics as the deprecated singular IPAddress, just plural.
+		addrs = append(addrs, n.IPAddresses...)
```
matthyx, line 229: The current `github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1.NetworkNeighbor` type only exposes `IPAddress` (singular) — `go doc ...NetworkNeighbor` has no `IPAddresses` field. That makes this append, and the matching ingress append below, a compile error against the current dependency, so this needs to stay on the singular field unless the storage API is updated first.

pkg/objectcache/v1/mock.go
```diff
@@ -175,55 +193,73 @@ func (r *RuleObjectCacheMock) GetProjectedContainerProfile(containerID string) *
...
+	// routed to Patterns rather than Values so the runtime CEL helpers can
+	// pass them through networkmatch on the cache-miss path.
 	if !specInstalled || spec.EgressAddresses.InUse || spec.EgressDomains.InUse {
 		for _, n := range cp.Spec.Egress {
-			if (!specInstalled || spec.EgressAddresses.InUse) && n.IPAddress != "" {
-				if pcp.EgressAddresses.Values == nil {
-					pcp.EgressAddresses.All = true
-					pcp.EgressAddresses.Values = make(map[string]struct{})
+			if !specInstalled || spec.EgressAddresses.InUse {
+				addrs := make([]string, 0, len(n.IPAddresses)+1)
```
matthyx: Same blocker in the mock path: `NetworkNeighbor` does not have an `IPAddresses` field, so `len(n.IPAddresses)` and the append here fail to compile (same issue repeats in the analogous ingress block below). `go test ./pkg/objectcache/v1` is currently red because of this.

matthyx (changes requested):
I found three merge-blocking build issues and left inline notes on each failing spot.

pkg/objectcache/containerprofilecache/tamper_alert.go
```diff
@@ -0,0 +1,190 @@
...
+	// drop a valid overlay because of a transient operational failure.
+	if !errors.Is(err, signature.ErrSignatureMismatch) {
+		logger.L().Warning("user-defined ApplicationProfile signature verification operational error (NOT tamper)",
+			helpers.String("profile", profile.Name),
+			helpers.String("namespace", profile.Namespace),
+			helpers.String("wlid", wlid),
+			helpers.Error(err))
+		// Honour strict-mode: refuse to load on any verification failure,
+		// but do NOT touch the dedup map or emit R1016.
+		return !c.cfg.EnableSignatureVerification
```
matthyx, line 94: `c.cfg.EnableSignatureVerification` still does not exist on `config.Config`, so this file has four hard compiler errors at the strict-mode return sites (for example here, and again at lines 108/136/146). Please either add/plumb that config field in this PR or remove the strict-mode branches so the tamper path matches currently available config.

pkg/signature/stub.go
```diff
@@ -0,0 +1,27 @@
...
+// errors.Is-checks. Must remain non-nil with this exact message —
+// tamper_alert_test.go pins both invariants.
+var ErrSignatureMismatch = errors.New("signature verification failed")
+
+// Signable is the marker interface adapters in pkg/signature/profiles
+// satisfy. Empty in the stub; the real package decorates it with
+// signature-payload accessors.
+type Signable interface{}
+
+func IsSigned(_ Signable) bool                    { return false }
```
matthyx: These stubs still leave the new tamper-detection path effectively disabled: `IsSigned` always returns false, `VerifyObjectAllowUntrusted` always returns nil, and the adapters in `pkg/signature/profiles/stub.go` return nil. That means the runtime path never verifies anything and the new tamper tests are not exercising the intended behavior. This needs the real signature implementation (or the feature/tests need to stay out of this PR).

matthyx (changes requested):
I re-checked this PR and I’m still seeing three merge blockers:

- `pkg/objectcache/containerprofilecache/tamper_alert.go` references `c.cfg.EnableSignatureVerification`, but `pkg/config/config.go` still has no such field. LSP flags four compiler errors at the strict-mode returns.
- `cmd/main.go` creates both the runtime alert exporter and the container profile cache, but never calls `cpc.SetTamperAlertExporter(exporter)`. As written, `emitTamperAlert()` is a no-op in production, so the new R1016 path never actually emits alerts end-to-end.
- `pkg/signature/stub.go` / `pkg/signature/profiles/stub.go` still short-circuit the feature (`IsSigned == false`, `VerifyObjectAllowUntrusted == nil`, adapters return `nil`), so the new tamper-detection behavior is effectively dormant and the added tamper tests are not validating real verification behavior.

entlein:
@coderabbitai review

pkg/signature/stub.go
entlein: Yes, those are just to make the PR "atomic" they need to be deleted after merge. I havent yet found a way to completely seperate the PRs 808 and 809

matthyx (commented):
Checked again.

The earlier import/IPAddresses blockers are addressed now: the PR has local `pkg/signature` stubs, and the storage replace brings `NetworkNeighbor.IPAddresses`.

I’m still seeing three blockers before merge:

- `pkg/objectcache/containerprofilecache/tamper_alert.go` still reads `c.cfg.EnableSignatureVerification` at the strict-mode return sites, but `pkg/config/config.go` still has no such field on `config.Config`.
- `cmd/main.go` creates both the runtime alert exporter and the container profile cache, but never calls `cpc.SetTamperAlertExporter(exporter)`, so the new R1016 tamper path still never emits alerts end-to-end.
- `pkg/signature/stub.go` and `pkg/signature/profiles/stub.go` still short-circuit the feature (`IsSigned == false`, `VerifyObjectAllowUntrusted == nil`, adapters return `nil`), so the tamper-detection path remains dormant and the added tamper tests are not validating a real verification flow.

entlein:
Component tests are green end-to-end for the tamper-alert wiring in `78f20392`: Test_01 (basic alert), Test_29 (signed application profile), Test_30 (tampered signed profile still loads without enforcement), and Test_31 (tamper detection fires R1016) all pass.

Root cause that `78f20392` addresses: `SetTamperAlertExporter` was never called in `cmd/main.go`, so `tamperAlertExporter` stayed `nil` and `emitTamperAlert` returned early — tampering was detected and logged, but the R1016 alert was silently dropped. Wiring the exporter once the container-profile cache is built emits R1016, while enforcement stays gated by `enableSignatureVerification` (default off). That separation is what lets Test_30 (tampered profile still loads) and Test_31 (R1016 fires) both pass: detection is independent of enforcement.

entlein:
@coderabbitai review

entlein:
@coderabbitai review

PR #822, opened 2026-05-26
feat: add container meta-context for shared ECS/Kubernetes rules

matthyx:
## Summary

This PR introduces a `container` meta-context so that rules can be written once and fire on **any containerized workload** — both Kubernetes pods and ECS/standalone containers — without duplicating tags.

### What changed

- **`pkg/contextdetection/types.go`**: Added `Container EventSourceContext = "container"` alongside the existing `Kubernetes`, `Host`, and `Standalone` constants.

- **`pkg/rulemanager/rulepolicy.go`**: Updated `RuleAppliesToContext` to treat `context:container` as a meta-context that matches when the agent is running in either a `kubernetes` or a `container` context.

### How the meta-context works

```
context:container  →  matches kubernetes OR container runtime
context:kubernetes →  matches kubernetes only  (unchanged)
context:host       →  matches host only         (unchanged)
context:standalone →  matches standalone only   (unchanged)
```

A rule tagged only `context:container` will fire on Kubernetes AND on ECS/standalone container nodes. A rule that needs Kubernetes-specific behaviour still uses `context:kubernetes` exclusively.

### Backward compatibility

Rules that should fire on both Kubernetes and ECS nodes should carry **both** tags:

```yaml
tags:
  - context:container    # fires on any containerised workload (new agents)
  - context:kubernetes   # still fires on old node-agents that don't know "container"
```

Old node-agents that don't recognise `container` will skip that tag and match `context:kubernetes` as before, so no existing rule behaviour is broken.

### Logic walk-through

```go
isContainerContext := currentContext == Kubernetes || currentContext == Container

for _, tag := range rule.Tags {
    if ctx == string(currentContext)              { return true } // exact match (all contexts)
    if ctx == "container" && isContainerContext   { return true } // meta-context match
}
```

No-context rules continue to default to Kubernetes-only (unchanged).

## Summary by CodeRabbit

* **New Features**
  * Added support for ECS (Amazon Elastic Container Service) context detection.
  * Runtime alerts now include ECS-specific metadata (cluster, task, launch type, availability zone).
  * Improved rule matching to support Container context across Kubernetes, Standalone, Container, and ECS environments.

* **Chores**
  * Updated Go module dependencies.

[Review Change Stack](https://app.coderabbit.ai/change-stack/kubescape/node-agent/pull/822?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pkg/contextdetection/types.go
+++ b/pkg/contextdetection/types.go
@@ -10,6 +10,7 @@ const (
 	Kubernetes EventSourceContext = "kubernetes"
 	Host       EventSourceContext = "host"
 	Standalone EventSourceContext = "standalone"
+	Container  EventSourceContext = "container"
 )
 
 type ContextInfo interface {
--- a/pkg/contextdetection/types.go
+++ b/pkg/contextdetection/types.go
@@ -11,6 +11,7 @@ const (
 	Host       EventSourceContext = "host"
 	Standalone EventSourceContext = "standalone"
 	Container  EventSourceContext = "container"
+	ECS        EventSourceContext = "ecs"
 )
 
 type ContextInfo interface {
--- a/go.mod
+++ b/go.mod
@@ -174,7 +174,7 @@ require (
 	github.com/cloudflare/circl v1.6.3 // indirect
 	github.com/cncf/xds/go v0.0.0-20251210132809-ee656c7534f5 // indirect
 	github.com/containerd/cgroups/v3 v3.0.5 // indirect
-	github.com/containerd/containerd v1.7.30 // indirect
+	github.com/containerd/containerd v1.7.32 // indirect
 	github.com/containerd/containerd/api v1.9.0 // indirect
 	github.com/containerd/continuity v0.4.5 // indirect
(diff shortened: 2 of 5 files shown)


pkg/contextdetection/types.go
```diff
@@ -10,6 +10,7 @@ const (
 	Kubernetes EventSourceContext = "kubernetes"
 	Host       EventSourceContext = "host"
 	Standalone EventSourceContext = "standalone"
+	Container  EventSourceContext = "container"
```
matthyx, line 13: I still want a pure `ecs` one here in case we have specific rules

matthyx: Added `ECS EventSourceContext = "ecs"` alongside `Container`. The rulelibrary PRs only added `context:container` to container-level behavioral rules (process exec, file access, escapes, network anomalies) — K8s-specific rules like R0006 (service-account-token), R0007 (kubernetes-client-executed), R2000 (exec-to-pod), R2001 (port-forward) were left untouched.

slashben (approved):

PR #834, opened 2026-06-10
refactor(rulemanager): reuse armoapi-go profileDataRequired schema

matthyx:
## What

Two commits:

1. **Schema dedup** — replaces node-agent's duplicate `profileDataRequired` schema with type aliases to the canonical armoapi-go types (`ProfileDataRequired`/`ProfileDataField`/`ProfileDataPattern`). Defining the matcher once in armoapi-go — imported by node-agent (query side), storage (generation side / rule-aware collapse), and the backend (rules in MongoDB) — guarantees it can't drift between the side that records a profile and the side that queries it. Shape change: a profile-data surface is now a pointer (`Opens *ProfileDataField`); a nil pointer means "not declared" (the role the old `FieldRequirement.Declared` bool played). `mergeField` updated to pointer + nil-check; the schema's own test moves to armoapi-go.

2. **Dependency fix** — resolves the transitive conflicts from the `kubescape/storage v0.0.282` bump using **additive replace directives only**, leaving the three frozen replaces (syft, inspektor-gadget, cilium/ebpf) and the storage pin untouched.

## Deps

- `armoapi-go` → `v0.0.719` (profileDataRequired schema + `UnionOpenProtection`).
- `kubescape/storage` → `v0.0.282`.

## Dependency resolution (was the blocker — now fixed)

The `storage v0.0.282` bump pulled a newer transitive stack that broke the build against the frozen forks:

- `opencontainers/runtime-spec` was forced to `v1.3.0`, which changed `LinuxPids.Limit` from `int64` to `*int64`; `containerd v1.7.32`'s `oci/spec_opts.go` assigns an `int64` and stopped compiling.
- `anchore/stereoscope` was forced to `v0.1.22`, whose docker/podman clients need a docker API (`client.New`, `client.PingOptions`) newer than the pinned `docker v28.5.2`; the frozen `kubescape/syft` fork expects `stereoscope v0.1.9`.

Resolved with two **additive** replaces pinning these transitives back to the versions the frozen set already uses:

```
replace github.com/opencontainers/runtime-spec => github.com/opencontainers/runtime-spec v1.2.1
replace github.com/anchore/stereoscope => github.com/anchore/stereoscope v0.1.9-0.20250826202322-ef061ea78385
```

(`runtime-spec v1.2.1` is what `origin/main` used; the stereoscope pseudo-version is the one the `kubescape/syft v1.32.0-ks.2` fork requires.) The three frozen replaces and `storage v0.0.282` are unchanged.

## Verified

- `go build ./...` — clean.
- `go test ./... -run='^$'` — all test binaries compile.
- Unit tests pass for the affected packages: `pkg/objectcache/containerprofilecache` (projection / `mergeField`) and `pkg/rulemanager/cel/libraries/applicationprofile` (`was_path_opened`).

## Related
- armosec/armoapi-go#656 (`UnionOpenProtection`, released in v0.0.719)
- kubescape/storage#335 (generation-side rule-aware collapse)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of profile data requirements, including clearer errors for unknown fields, invalid values, and malformed patterns.
  * Prevented invalid profile data from being accepted during rule conversion.
  * Updated profile requirement handling to correctly recognize undeclared fields.

* **Tests**
  * Added coverage for valid profile data conversion and validation failures.

* **Chores**
  * Updated the Go toolchain and project dependencies.


matthyx:
**Review — no merge blockers.** CI is fully green (build-and-push + all 24 component tests) and the refactor is sound. One non-blocking gap and a deps note below.

### ✅ Verified
- **Pointer migration is internally consistent.** In `armoapi-go v0.0.719`, `ProfileDataRequired.Opens … IngressAddresses` are all `*ProfileDataField`, so `mergeField(dst, src *typesv1.FieldRequirement)` with the `src == nil` check correctly takes over the old `Declared` bool's role, and the call sites need no edit (the field type itself went value→pointer). The `FieldRequirement = armotypes.ProfileDataField` / `PatternObject = armotypes.ProfileDataPattern` aliases line up with the upstream definitions.
- **Dependency resolution.** The two additive replaces (`runtime-spec → v1.2.1`, `stereoscope → v0.1.9-…ef061ea`) pin the transitives back to what the frozen syft/inspektor-gadget/ebpf set already uses; the three frozen replaces and the `storage v0.0.282` pin are untouched.

### ⚠️ Non-blocking — load-time validation is now silently weaker
The deleted local schema rejected malformed `profileDataRequired` **at unmarshal time**: unknown top-level surface keys, unknown keys inside a pattern object, and multi-key/empty pattern objects all errored on decode. armoapi-go folds that logic into an explicit `ProfileDataRequired.Validate()` — whose own doc calls it *"the single source of truth for both the rulelibrary lint and node-agent's load-time check"* — but node-agent never calls `Validate()` on the load path (`rule_manager.go` / `config.go` only check presence/`profileDependency`, not schema).

Net effect: `ProfileDataField` *shape* errors are still caught (its `UnmarshalJSON`/`UnmarshalYAML` survive in armoapi-go), but a **typo'd surface name** (e.g. `opnes:`) or a **malformed pattern object** is now silently dropped instead of rejected — which can quietly empty a rule's projection and turn into a missed detection. The upstream rulelibrary lint still validates, so this is defense-in-depth rather than a correctness break. If you want to preserve node-agent's prior strictness, add a `pdr.Validate()` call where rules are loaded.

### ℹ️ On CodeRabbit's two security findings
`docker v28.5.2` (HIGH) and the `containerd` GHSA-fqw6-gf59-qr4w note are **transitive and not introduced by this PR** — docker stays pinned at `v28.5.2` by the frozen syft/docker set, unchanged here. Out of scope for this refactor; worth tracking separately rather than blocking the merge.

PR #837, opened 2026-06-19
fix(container-watcher): do not close workerChan on Stop (send-on-closed panic)

matthyx:
## Summary

`ContainerWatcher.Stop()` closed `cw.workerChan` while `enrichAndProcess` still sends on it, so an in-flight send racing `Stop()` panics with `send on closed channel` on teardown (rolling restart / SIGTERM). Remove the close — both producer and consumer already terminate on `cw.ctx.Done()`, so the close was vestigial and only created the race.

## Details

`enrichAndProcess` sends on `cw.workerChan` in both the `select` case and the `BlockEvents` blocking branch. **A send on a closed channel panics even inside a `select`** (the `select` only avoids *blocking*, not send-on-closed), so on teardown:

```
panic: send on closed channel
	.../pkg/containerwatcher/v2/container_watcher.go  (enrichAndProcess send)
created by ...v2.(*ContainerWatcher).Start
```

Observed on several node-agent pods during a routine DaemonSet rollout.

Why removing the close is correct:
- consumer `workerPoolLoop`: `select { case <-cw.ctx.Done(): return; case e := <-cw.workerChan: ... }`
- producer `eventProcessingLoop` → `enrichAndProcess`: bounded by the same ctx.

There is no `range` over `workerChan`, and a `select`-reader on a *closed* channel busy-spins (returns zero values immediately) rather than exiting — so the close never drove worker shutdown, it only created the send-on-closed hazard. The channel is garbage-collected with the `ContainerWatcher`.

## Test
`go build` + `go vet ./pkg/containerwatcher/v2/...` green. The change deletes the `close()`; no behavior change beyond removing the panic.

## Release Notes

* **Bug Fixes**
  * Resolved potential race conditions and panics that could occur during container watcher shutdown, improving overall stability during application termination.

### DIFF
--- a/pkg/containerwatcher/v2/container_watcher.go
+++ b/pkg/containerwatcher/v2/container_watcher.go
@@ -8,9 +8,6 @@ import (
 
 	mapset "github.com/deckarep/golang-set/v2"
 	containercollection "github.com/inspektor-gadget/inspektor-gadget/pkg/container-collection"
-	"go.opentelemetry.io/otel/attribute"
-	"go.opentelemetry.io/otel/metric"
-	"github.com/kubescape/node-agent/pkg/otelsetup"
 	containerutilsTypes "github.com/inspektor-gadget/inspektor-gadget/pkg/container-utils/types"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators/socketenricher"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/runtime"
@@ -22,8 +19,8 @@ import (
 	"github.com/kubescape/node-agent/pkg/config"
 	"github.com/kubescape/node-agent/pkg/containerprofilemanager"
 	"github.com/kubescape/node-agent/pkg/containerwatcher"
-	"github.com/kubescape/node-agent/pkg/dedupcache"
 	"github.com/kubescape/node-agent/pkg/containerwatcher/v2/tracers"
+	"github.com/kubescape/node-agent/pkg/dedupcache"
 	"github.com/kubescape/node-agent/pkg/dnsmanager"
 	"github.com/kubescape/node-agent/pkg/ebpf/events"
 	"github.com/kubescape/node-agent/pkg/eventreporters/rulepolicy"
@@ -31,6 +28,7 @@ import (
 	"github.com/kubescape/node-agent/pkg/metricsmanager"
 	"github.com/kubescape/node-agent/pkg/networkstream"
 	"github.com/kubescape/node-agent/pkg/objectcache"
+	"github.com/kubescape/node-agent/pkg/otelsetup"
 	"github.com/kubescape/node-agent/pkg/processtree"
 	containerprocesstree "github.com/kubescape/node-agent/pkg/processtree/container"
 	"github.com/kubescape/node-agent/pkg/rulebindingmanager"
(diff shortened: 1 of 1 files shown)


matthyx (commented):
## Review: removing `close(cw.workerChan)` on `Stop()`

The core fix is **correct**: deleting the `close()` removes the `send on closed channel` panic, because `enrichAndProcess` can be mid-send (both the `select` case and the `BlockEvents` blocking branch send on `cw.workerChan`) when `Stop()` runs, and a send on a closed channel panics even inside a `select`. The accompanying observation is also right — a `select`-reader on a *closed* channel busy-spins on the zero value rather than exiting, so the `close()` never actually drove worker shutdown; it was purely a hazard. Good catch and a clean, minimal change.

One blocker before merge, about the *justification* (not the deletion itself):

### The comment's premise — "both loops terminate on `cw.ctx.Done()`" — is not true as wired

In `cmd/main.go` the watcher is started with `ctx := context.Background()` (never cancelled), and `Stop()` does not cancel any context — the struct holds only a `ctx context.Context` field, no `cancel func`. So in the running daemon:

- `workerPoolLoop` and `eventProcessingLoop` **never** return via `<-cw.ctx.Done()`; they live until the process itself exits.
- `Stop()` therefore does **not** stop either loop. It happens to be safe today only because the *sole* caller is `defer mainHandler.Stop()` in `main`, which runs immediately before the process exits, so the leaked goroutines die with the process.

Two consequences worth addressing:

1. **The new comment is misleading.** It tells a future reader that shutdown is ctx-driven and clean, when in reality shutdown here is process-exit-driven. Anyone who later calls `Stop()` outside of process teardown (in-place restart, a health-manager-triggered restart, or a unit test) will hit the real behavior:
2. **`Stop()` is a silent event-dropper if ever called while the process keeps running.** After `cw.workerPool.Release()`, `workerPoolLoop` keeps draining `workerChan` and calling `cw.workerPool.Invoke(...)` on the released pool — `ants` returns `ErrPoolClosed`, the return value is ignored, and every event is dropped silently while both goroutines leak.

**Requested change:** either
- (preferred) give the watcher its own cancelable context — store a `cancel` in the struct, derive `cw.ctx, cw.cancel = context.WithCancel(ctx)` in `Start`, and call `cw.cancel()` at the top of `Stop()`. Then the loops genuinely terminate on `ctx.Done()`, the comment becomes accurate, and `Stop()` is correct in any caller context. (Note the `BlockEvents` branch can still block on a full channel after the consumer exits — a small drain/`ctx`-aware send would close that gap.) Or
- (minimum) reword the comment to state the truth: the loops are not stopped by `Stop()`; they are bound to the process lifetime, and `Stop()` relies on being the last thing before process exit.

Non-blocking nit: `cw.workerPool.Release()` races `workerPoolLoop`'s `Invoke`; harmless (`ants` returns `ErrPoolClosed`, no panic), just noting it's expected.

matthyx:
You're right — my "both loops terminate on `cw.ctx.Done()`" justification was false as wired (`context.Background()` in `cmd/main.go`, no cancel in `Stop()`), so it only worked because `Stop()` runs right before process exit.

Implemented the **preferred** option (force-pushed `f32f765`):

- `Start` derives a cancelable context: `cw.ctx, cw.cancel = context.WithCancel(ctx)`.
- `Stop` calls `cw.cancel()` **first**, before tearing down the worker pool / channels, so `eventProcessingLoop` and `workerPoolLoop` actually return — fixing both the goroutine leak and the silent event-drop-on-released-pool when `Stop()` is called mid-process. The comment now states the truth.
- Kept the `close(workerChan)` removal.
- Made the `BlockEvents` blocking send **ctx-aware** (`select { case cw.workerChan <- e: case <-cw.ctx.Done(): e.Release() }`) so a full channel can't block forever once the consumer has exited — closing the gap you flagged.

`go build` + `go vet` (incl. `lostcancel`) clean. Left the `workerPool.Release()` vs `Invoke` race as-is per your non-blocking nit (`ants` returns `ErrPoolClosed`, no panic).

PR #838, opened 2026-06-22
feat(config): add alertDeduplication.bypass master switch

slashben:
## What

Adds `alertDeduplication.bypass` — a single, off-by-default config switch that disables all sensor-side alert suppression.

When `true`, config loading forces:
- `eventDedup.enabled=false` (no raw eBPF event deduplication)
- `ruleCooldown.ruleCooldownOnProfileFailure=false` (no per-signature rule cooldown)

so every rule firing is exported.

## Why

High-volume security-testing (DAST) scenarios need every alert to surface rather than being collapsed by deduplication/cooldown. This provides one clearly-scoped switch instead of toggling several low-level knobs.

## Notes
- Default `false` → no behavior change for existing deployments.
- Applied before `eventDedup.slotsExponent` validation, so an out-of-range exponent doesn't cause a spurious load error when bypass is on.
- Includes `pkg/config` unit tests and a feature doc.

## Summary by CodeRabbit

* **New Features**
  * Added `alertDeduplication.bypass` master switch (default `false`) to disable sensor-side alert suppression, ensuring all rule firings are exported.
  * Added `ruleCooldownDisabled` kill-switch for bypassing rule cooldown regardless of other cooldown settings.
* **Documentation**
  * Updated documentation to explain the bypass behavior, intent, and trade-offs (higher export volume/CPU).
* **Bug Fixes**
  * When bypass is enabled, related validation is skipped to prevent out-of-range values from failing config load.
* **Tests**
  * Added unit tests covering bypass behavior and bypass-related validation changes.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -248,7 +248,7 @@ func LoadConfigOptional(path string, errNotFound bool) (Config, error) {
 
 	if err := viper.ReadInConfig(); err != nil {
 		var notFound viper.ConfigFileNotFoundError
-		if !(errors.As(err, &notFound) && !errNotFound) {
+		if !errors.As(err, &notFound) || errNotFound {
 			return Config{}, err
 		}
 	}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -44,6 +44,13 @@ type EventDedupConfig struct {
 	SlotsExponent uint8 `mapstructure:"slotsExponent"`
 }
 
+// AlertDeduplicationConfig is a master switch that bypasses all sensor-side alert
+// deduplication and cooldown. Off by default; intended for high-volume
+// security-testing (DAST) use cases.
+type AlertDeduplicationConfig struct {
+	Bypass bool `mapstructure:"bypass"`
+}
+
 type Config struct {
 	BlockEvents                    bool                                 `mapstructure:"blockEvents"`
 	CelConfigCache                 cache.FunctionCacheConfig            `mapstructure:"celConfigCache"`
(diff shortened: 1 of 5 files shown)


slashben:
Helm chart wiring that exposes this config value: kubescape/helm-charts#864

pkg/config/config.go
```diff
@@ -256,6 +265,14 @@ func LoadConfigOptional(path string, errNotFound bool) (Config, error) {
 			config.SeccompProfileBackend, SeccompBackendStorage, SeccompBackendCRD)
 	}
 
+	// High-volume security testing (DAST): a single master switch that bypasses ALL sensor-side suppression.
```
matthyx, line 269: **"Bypasses ALL sensor-side suppression" is overstated — host FIM dedup is not covered.**

This switch disables eBPF event dedup and rule cooldown, but the host FIM sensor has its own independent dedup that this block does not touch: `fimConfig.dedupConfig.dedupEnabled` -> `pkg/hostfimsensor/v1/dedup_cache.go`, gated at `fimsensor_periodic.go:61` and `fimsensor_fanotify.go:78`. If a DAST run exercises file-integrity monitoring, those events still get deduplicated despite `bypass=true`.

Please either:
- (a) extend this block to also force `config.FIMConfig...DedupConfig.DedupEnabled = false`, or
- (b) narrow the wording here, in the PR description, and in `docs/features/alert-deduplication-bypass.md` to "eBPF event dedup and rule cooldown".

(FYI `pkg/cooldownqueue` is only used by the k8s `dynamicwatcher`, not an alert path, so it's correctly out of scope.)

pkg/config/config.go
```diff
@@ -256,6 +265,14 @@ func LoadConfigOptional(path string, errNotFound bool) (Config, error) {
...
+	// It forces the existing event-dedup and rule-cooldown switches off, so the
+	// downstream code paths (container watcher, rule manager) need no changes.
+	if config.AlertDeduplication.Bypass {
+		config.EventDedup.Enabled = false
+		config.RuleCoolDown.OnProfileFailure = false
```
matthyx: **Fragile coupling: the bypass disables rule cooldown via a flag marked deprecated.**

`config.RuleCoolDown.OnProfileFailure = false` works today only because `ShouldCooldown` early-returns `false` when this flag is off (`pkg/rulemanager/rulecooldown/rulecooldown.go:38`). But that field is explicitly marked deprecated (`viper.SetDefault("ruleCooldown::ruleCooldownOnProfileFailure", true) // NOTE: this is deprecated.`), and its name no longer reflects that it's the de-facto master enable gate for cooldown.

If anyone later removes or repurposes `OnProfileFailure`, this bypass silently stops disabling cooldown — and no test would catch it, because the tests assert the *field value*, not actual cooldown behavior.

Suggest either introducing a proper `Enabled` field for rule cooldown and toggling that here, or at minimum adding a comment at `rulecooldown.go:38` noting `OnProfileFailure` is the master switch this bypass depends on.

PR #839, opened 2026-06-23
fix: correct IgnoreRuleBindings handling for container monitoring

matthyx:
## What

Corrects how `IgnoreRuleBindings` is handled across container monitoring and the rule-binding cache.

### Background / root cause
In `IgnoreRuleBindings` mode every pod is evaluated against the full rule set, so `ruleManagedPods` is never populated (it is only fed from binding notifications). When a container's ApplicationProfile reaches end-of-life, the EOL-driven `unregisterContainer` tore down monitoring — detection silently stopped after learning completed. Validated on a live cluster (chart sets `IGNORERULEBINDINGS=true`; with no binding, attack-suite containers were unregistered at profile-EOL and a command-injection produced zero detections).

### The fix
The original one-line fix short-circuited the **entire** `unregisterContainer`, which also disabled removal of explicitly *ignored* containers (excludeNamespaces/labels, e.g. kube-system) and cleanup of learning-complete containers when runtime detection is off. This PR scopes the behavior correctly:

- **`unregisterContainer`** — the `IgnoreRuleBindings` early-return is placed *inside* the `EnableRuntimeDetection` block (the actual buggy decision), treating every pod as rule-managed so the container keeps being monitored. The real removal is extracted into `removeContainer`.
- **`containerCallback`** — ignored containers now call `removeContainer` directly, so they are always dropped regardless of rule bindings / runtime detection.

### Stop wasted rule-binding work when bindings are ignored
- **`resourcesToWatch`** no longer watches `RuntimeAlertRuleBinding` objects (RBCache is the sole watcher), which also removes the misleading `RBCache - refreshed rule bindings rules` log.
- **`RefreshRuleBindingsRules`** skips the per-binding rule rebuild but **still notifies consumers**, so `RuleManager` keeps recompiling the profile projection spec on `Rule` CRD changes.
- **RBCache handlers** skip pod→binding bookkeeping (`addPod`/`deletePod`) since it is never read in this mode.

## Tests
- `TestUnregisterContainer`: new case keeping a container in `ignoreRuleBindings` mode with no matching binding (fails on `main`, passes here).
- `TestResourcesToWatch`: new case asserting rule bindings are not watched when ignored.
- `go build ./...` clean; `pkg/rulebindingmanager/...`, `pkg/rulemanager`, `pkg/containerwatcher/v2` green.

## Docs
- `docs/CONFIGURATION.md` — expanded the `ignoreRuleBindings` entry to describe the now-explicit behavior.

## Follow-up (not in this PR)
Component-test blind spot: the test chart ships an unconditional `all-rules-all-pods` binding, so `ruleManagedPods` is always populated and the bug is masked. A component-test variant with `ignoreRuleBindings=true` and **no** binding would close the gap.

### DIFF
--- a/pkg/containerwatcher/v2/containercallback.go
+++ b/pkg/containerwatcher/v2/containercallback.go
@@ -31,7 +31,8 @@ func (cw *ContainerWatcher) containerCallback(notif containercollection.PubSubEv
 			helpers.String("containerID", notif.Container.Runtime.ContainerID))
 		// avoid loops when the container is being removed
 		if notif.Type == containercollection.EventTypeAddContainer {
-			cw.unregisterContainer(notif.Container)
+			// ignored containers must be dropped regardless of rule bindings / runtime detection
+			cw.removeContainer(notif.Container)
 		}
 		return
 	}
@@ -196,6 +197,15 @@ func (cw *ContainerWatcher) unregisterContainer(container *containercollection.C
 	// 1. Runtime detection is disabled, OR
 	// 2. Rule bindings have been initialized AND this pod is not in ruleManagedPods
 	if cw.cfg.EnableRuntimeDetection {
+		// When rule bindings are ignored, every pod is effectively rule-managed
+		// (all rules apply to all pods), so the container must keep being monitored.
+		if cw.cfg.IgnoreRuleBindings {
+			logger.L().Debug("ContainerWatcher - container should still be monitored for runtime detection (rule bindings ignored)",
+				helpers.String("container ID", container.Runtime.ContainerID),
+				helpers.String("namespace", container.K8s.Namespace), helpers.String("PodName", container.K8s.PodName), helpers.String("ContainerName", container.K8s.ContainerName),
+			)
+			return
+		}
 		// If the pod is currently in ruleManagedPods, it should still be monitored
 		if cw.ruleManagedPods.Contains(k8sPodID) {
 			logger.L().Debug("ContainerWatcher - container should still be monitored for runtime detection (in ruleManagedPods)",
@@ -217,6 +227,13 @@ func (cw *ContainerWatcher) unregisterContainer(container *containercollection.C
 		// meaning rule bindings exist but don't apply to this pod - proceed with unregistration
 	}
 
+	cw.removeContainer(container)
+}
+
+// removeContainer removes a container from monitoring unconditionally, releasing
+// its collection entry and shared cache data. Use this when the container must be
+// dropped regardless of rule bindings or runtime detection (e.g. ignored containers).
+func (cw *ContainerWatcher) removeContainer(container *containercollection.Container) {
 	logger.L().Debug("ContainerWatcher - stopping to monitor on container", helpers.String("container ID", container.Runtime.ContainerID), helpers.String("namespace", container.K8s.Namespace), helpers.String("PodName", container.K8s.PodName), helpers.String("ContainerName", container.K8s.ContainerName))
 
 	cw.containerCollection.RemoveContainer(container.Runtime.ContainerID)
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -160,7 +160,7 @@ correlation is an upgrade, not a dependency.
 | `fullPathTracingEnabled` | bool | `true` | Include full executable paths |
 | `enableEmbeddedSBOMs` | bool | `false` | Use embedded SBOMs from images |
 | `partialProfileGenerationEnabled` | bool | `true` | Allow partial profile generation |
-| `ignoreRuleBindings` | bool | `false` | Apply all rules regardless of bindings |
+| `ignoreRuleBindings` | bool | `false` | Apply all rules to all pods regardless of bindings. When enabled, RuntimeAlertRuleBinding objects are not watched, every monitored container is kept for runtime detection, and per-pod binding bookkeeping is skipped (rules are resolved directly from the rule creator). |
 
 ### Timing & Performance
 
--- a/pkg/containerwatcher/v2/container_watcher_private_test.go
+++ b/pkg/containerwatcher/v2/container_watcher_private_test.go
@@ -133,6 +133,7 @@ func TestUnregisterContainer(t *testing.T) {
 		expectedContainers      []string
 		enableRuntimeDetection  bool
 		ruleBindingsInitialized bool
+		ignoreRuleBindings      bool
 	}{
 		{
 			name:                    "Test unregister container with runtime detection enabled and bindings initialized",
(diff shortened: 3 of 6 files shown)


slashben (commented):
Confirmed this fixes the bug. Root cause: in `IgnoreRuleBindings` mode `ruleManagedPods` is never populated (it's only fed from binding notifications), so when a container's ApplicationProfile reaches end-of-life the EOL-driven `unregisterContainer` tears down monitoring — detection silently stops after learning completes. Returning early when the flag is set keeps the container monitored. On the production config (`EnableRuntimeDetection=true`) this behaves the same as a guard placed inside the runtime-detection branch.

A few comments:

**1. (Should-fix) Add a regression test.** This is a *silent-failure* bug — detection just stops, no error — so it's exactly the kind that regresses unnoticed. The existing `TestUnregisterContainer` table covers it with one case (add an `ignoreRuleBindings` field to the table struct + wire it into the `config.Config{...}`):

```go
{
    name:                    "Keep container in ignoreRuleBindings mode with no matching binding",
    unregisterContainer:     "container1",
    unregisterContainersPod: "pod1",
    podToContainers:         map[string][]string{"pod1": {"container1"}, "pod2": {"container2"}},
    preRuleManagedPods:      []string{},                           // empty: no bindings
    expectedContainers:      []string{"container1", "container2"}, // both kept
    enableRuntimeDetection:  true,
    ruleBindingsInitialized: true,
    ignoreRuleBindings:      true,
},
```

Fails on `main`, passes with this fix.

**2. (Nit) Add a one-line comment.** A bare `if cw.cfg.IgnoreRuleBindings { return }` doesn't tell the next reader why — one sentence ("in this mode every pod is evaluated against all rules, so `ruleManagedPods` stays empty and a container must never be unregistered for lacking a binding") would help.

**3. (Question, not a blocker) Effect on the excluded-namespace path.** `unregisterContainer` is also called from the `IgnoreContainer` branch in `containercallback.go` (the excludeNamespaces/labels cleanup for kube-system, kubescape, the agent itself). With this early-return, in `IgnoreRuleBindings` mode those ignored containers are no longer explicitly removed from the container collection. Likely benign — the monitoring callbacks are already skipped for ignored containers — but since armo mode runs with `IgnoreRuleBindings=true` in production, worth confirming excluded namespaces don't start lingering/monitored. If there's any doubt, scoping the guard inside the `if cw.cfg.EnableRuntimeDetection` block confines it to the actual buggy decision.

Net: LGTM with the test added; #2/#3 are polish.

Validated end-to-end on a live cluster (chart sets `IGNORERULEBINDINGS=true`; with no binding, attack-suite containers were unregistered at profile-EOL and a command-injection produced zero detections; restoring monitoring brought the alerts back).

pkg/containerwatcher/v2/containercallback.go
```diff
@@ -187,6 +187,9 @@ func (cw *ContainerWatcher) getSharedWatchedContainerData(container *containerco
 
 // unregisterContainer unregisters a container from monitoring
 func (cw *ContainerWatcher) unregisterContainer(container *containercollection.Container) {
+	if cw.cfg.IgnoreRuleBindings {
+		return
+	}
```
slashben: (nit) Worth a sentence on *why* this early return exists, so the next reader does not have to reverse-engineer the `ruleManagedPods` interaction:

```suggestion
	// In ignoreRuleBindings mode every pod is evaluated against the full rule set
	// regardless of RuntimeRuleAlertBindings, so ruleManagedPods stays empty; without
	// this guard a container is unregistered at profile end-of-life even though it
	// must remain monitored.
	if cw.cfg.IgnoreRuleBindings {
		return
	}
```

slashben: (should-fix) Add a regression test — this is a silent-failure path (detection just stops, no error), so it's easy to regress. The existing `TestUnregisterContainer` table in `container_watcher_private_test.go` covers it with one case (add an `ignoreRuleBindings bool` field to the table struct and pass it into the `config.Config{...}`):

```go
{
    name:                    "Keep container in ignoreRuleBindings mode with no matching binding",
    unregisterContainer:     "container1",
    unregisterContainersPod: "pod1",
    podToContainers:         map[string][]string{"pod1": {"container1"}, "pod2": {"container2"}},
    preRuleManagedPods:      []string{},                           // empty: no bindings
    expectedContainers:      []string{"container1", "container2"}, // both kept
    enableRuntimeDetection:  true,
    ruleBindingsInitialized: true,
    ignoreRuleBindings:      true,
},
```

Fails on `main`, passes with this fix. (Suggestion can't be one-click here since the test file isn't in this PR's diff.)

slashben:
### Why CI didn't catch this (component-test blind spot)

`Test_18_LearningMode` actually exercises the exact failure window — it execs `wget open.stealth.si` *after* the ApplicationProfile completes and asserts the alert fires (`tests/node-agent/component_test.go:1632`). So the gap isn't missing coverage of the scenario; it's the test **chart config**.

The component-test chart sets two things together:
- `tests/node-agent/chart/templates/node-agent/configmap.yaml:40` → `"ignoreRuleBindings": true` (same as prod)
- `tests/node-agent/chart/templates/node-agent/default-rule-binding.yaml` → an **unconditional** `all-rules-all-pods` `RuntimeRuleAlertBinding` (no `installDefault` gate — always rendered)

This bug only triggers with **`ignoreRuleBindings=true` AND no matching binding** (empty `ruleManagedPods`). Because the test always ships the binding, every pod lands in `ruleManagedPods` and `unregisterContainer` returns at the **first** guard (`ruleManagedPods.Contains`) — the buggy "bindings initialized + pod not managed" branch is never reached. So Test 18 stays green.

| | `ignoreRuleBindings` | default binding | `ruleManagedPods` | result |
|---|---|---|---|---|
| Component-test chart | `true` | always created | populated | bug masked → green |
| Prod armosec chart (default) | `true` | gated behind `alertCRD.installDefault=false` → none | empty | **bug triggers** |

CI runs the one config where the binding hides the bug; the shipped chart default runs the one config where it bites.

**Suggested coverage:** a component-test variant with `ignoreRuleBindings=true` and **no** `RuntimeRuleAlertBinding`, asserting a pod still alerts after its profile completes. (The unit-test case suggested above covers it at the `unregisterContainer` level; this would close the component-level gap.)

pkg/containerwatcher/v2/containercallback.go
matthyx: Addressed: the guard was relocated inside the `EnableRuntimeDetection` block and carries an explanatory comment — "When rule bindings are ignored, every pod is effectively rule-managed (all rules apply to all pods), so the container must keep being monitored."

matthyx: Done — added the `ignoreRuleBindings bool` field to the `TestUnregisterContainer` table plus the regression case ("Test keep container when rule bindings are ignored"), wired into `config.Config{...}`.

matthyx:
Thanks @slashben — addressed your review:

**#3 (excluded-namespace path / scope concern):** Confirmed real, and fixed exactly as you suggested. The `IgnoreRuleBindings` early-return is now scoped **inside** the `EnableRuntimeDetection` block, so it only governs the actual buggy decision. The `IgnoreContainer` (excludeNamespaces/labels) cleanup path no longer routes through that logic at all — it calls a new `removeContainer` helper directly, so ignored containers are always dropped regardless of `IgnoreRuleBindings`. No lingering/monitored excluded namespaces in armo mode.

**#1/#2 (regression test + comment):** done (replied inline).

While in here I also stopped the now-dead rule-binding work in this mode: RBCache no longer watches `RuntimeAlertRuleBinding` (removing the misleading `refreshed rule bindings rules` log), `RefreshRuleBindingsRules` skips the per-binding rebuild but still notifies so `RuleManager` keeps recompiling the projection spec, and pod→binding bookkeeping is skipped.

**CI blind spot:** great catch — left as a noted follow-up in the PR description (component-test variant with `ignoreRuleBindings=true` and no binding). Happy to add it here or in a separate PR, your call.

slashben (approved):

PR #841, opened 2026-06-24
fix: map network event pktType to direction in CEL rules engine

### LINKED ISSUE #840: bug: event.direction is empty for network events in CEL rules engine
### Describe the bug
When writing custom CEL rules for network events (eventType: `network`), referencing `event.direction` always returns an empty string `""`. Consequently, expressions like `event.direction == 'outbound'` (e.g., in a rule like `R1077`) always evaluate to `false` and can never fire.

### Root Cause
1. In `pkg/utils/cel.go`, the CEL property `direction` retrieves `x.Raw.GetDirection()`.
2. For network events, `GetDirection()` returns the raw event's `Direction` field (of type `consts.NetworkDirection`).
3. For network events generated by the network tracer, the `Direction` field is never populated (unlike HTTP events where it is set via `MakeHttpEvent`).
4. Network events instead carry packet type information (`GetPktType()`), containing `"OUTGOING"` or `"HOST"`.

### Proposed Fixes
**Option A: Map pktType inside GetDirection()**
Update `GetDirection()` in `pkg/utils/datasource_event.go` and `pkg/utils/struct_event.go` to fallback to checking packet type for network events:
```go
func (e *DatasourceEvent) GetDirection() consts.NetworkDirection {
	if e.Direction == "" && e.EventType == NetworkEventType {
		if strings.EqualFold(e.GetPktType(), "OUTGOING") {
			return consts.Outbound // "outbound"
		}
		return consts.Inbound // "inbound"
	}
	return e.Direction
}
```

**Option B: Document that rules must use pktType**
Alternatively, if `event.direction` is meant to be exclusive to L7 / HTTP, we should update documentation and custom rules to use `event.pktType == 'OUTGOING'` instead.

matthyx:
This PR fixes a bug where `event.direction` on a network event is always empty, causing rules evaluating `event.direction == 'outbound'` (such as R1077) to never trigger. It falls back to mapping Inspektor Gadget's `pktType` (`OUTGOING` vs others) to direction if `e.Direction` is empty.

Closes #840

## Summary by CodeRabbit

* **Bug Fixes**
  * Network events now automatically show the correct inbound or outbound direction when no direction is explicitly set.
  * Direction handling is more consistent across event types, including cases where packet type information is used to determine the result.

* **Tests**
  * Expanded coverage for event direction behavior, including explicit directions and packet-type-based inference.

### DIFF
--- a/pkg/utils/datasource_event.go
+++ b/pkg/utils/datasource_event.go
@@ -273,6 +273,9 @@ func (e *DatasourceEvent) GetCwd() string {
 }
 
 func (e *DatasourceEvent) GetDirection() consts.NetworkDirection {
+	if e.Direction == "" && e.EventType == NetworkEventType {
+		return DirectionFromPktType(e.GetPktType())
+	}
 	return e.Direction
 }
 
--- a/pkg/utils/struct_event.go
+++ b/pkg/utils/struct_event.go
@@ -154,6 +154,9 @@ func (e *StructEvent) GetCwd() string {
 }
 
 func (e *StructEvent) GetDirection() consts.NetworkDirection {
+	if e.Direction == "" && e.EventType == NetworkEventType {
+		return DirectionFromPktType(e.GetPktType())
+	}
 	return e.Direction
 }
 
--- a/pkg/utils/events.go
+++ b/pkg/utils/events.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net/http"
 	"path/filepath"
+	"strings"
 
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/types"
 	"github.com/kubescape/storage/pkg/apis/softwarecomposition/consts"
@@ -269,4 +270,12 @@ func protoNumToString(protoNum uint16) string {
 	default:
 		return ""
 	}
+}
+
+// DirectionFromPktType maps an Inspektor Gadget packet type string to a NetworkDirection
+func DirectionFromPktType(pktType string) consts.NetworkDirection {
+	if strings.EqualFold(pktType, OutgoingPktType) {
+		return consts.Outbound
+	}
+	return consts.Inbound
 }
--- a/pkg/utils/events_test.go
+++ b/pkg/utils/events_test.go
@@ -1,13 +1,131 @@
 package utils
 
-// Test struct that mimics the structure of events with Comm field
-type testEventWithComm struct {
-	Comm string
+import (
+	"testing"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/datasource"
+	"github.com/kubescape/storage/pkg/apis/softwarecomposition/consts"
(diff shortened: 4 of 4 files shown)


pkg/utils/datasource_event.go
```diff
@@ -273,6 +273,12 @@ func (e *DatasourceEvent) GetCwd() string {
 }
 
 func (e *DatasourceEvent) GetDirection() consts.NetworkDirection {
+	if e.Direction == "" && e.EventType == NetworkEventType {
+		if strings.EqualFold(e.GetPktType(), "OUTGOING") {
```
matthyx: This is the block that actually fixes #840 in production (network events flow through `DatasourceEvent`), yet there's no test for it — the new test only covers `StructEvent`. Worth a `DatasourceEvent.GetDirection()` test that populates the `egress` field.

Nit: prefer the existing `OutgoingPktType` constant (events.go) over the `"OUTGOING"` string literal.

pkg/utils/struct_event.go
```diff
@@ -154,6 +155,12 @@ func (e *StructEvent) GetCwd() string {
 }
 
 func (e *StructEvent) GetDirection() consts.NetworkDirection {
+	if e.Direction == "" && e.EventType == NetworkEventType {
+		if strings.EqualFold(e.GetPktType(), "OUTGOING") {
```
matthyx: Nit: this block is duplicated verbatim in `datasource_event.go`; a shared helper (e.g. `directionFromPktType(pktType)`) would keep the two in sync. Also prefer the `OutgoingPktType` constant over the `"OUTGOING"` literal.

matthyx (commented):
Reviewed the fix end-to-end. The approach is sound and the chain works: for network events the production type is `DatasourceEvent`, whose `GetPktType()` reads the `egress` field (egress==1 -> OUTGOING, else HOST), so the new fallback yields outbound/inbound, and that value reaches CEL via the `cel.go` "direction" field getter (`x.Raw.GetDirection()`). I confirmed `consts.Inbound=="inbound"` / `Outbound=="outbound"`, matching R1077's `direction == 'outbound'`, and that the prefilter does not drop network events on direction (`rule_manager.go` only calls `SetDirection` for HTTP events, and prefilter line 254 short-circuits when `e.Dir == DirNone`). No hard blockers — but one test-coverage gap is worth fixing before merge, plus two minor nits inline.

**Main concern (please address): the regression test does not cover the regressed code path.** `TestStructEventGetDirection` only exercises `StructEvent`, but the production network path uses `DatasourceEvent`, and `StructEvent.PktType` is never populated for network events outside tests. So the identical block in `datasource_event.go` — the code that actually fixes #840 — and its real wiring through the `egress` field are untested. Consider a table test on `DatasourceEvent.GetDirection()` that sets `egress` in `Data` and asserts inbound/outbound.

PR #845, opened 2026-06-26
fix(cooldownqueue): eliminate "send on closed channel" panic on shutdown

matthyx:
## Problem

The `istio.io/pkg/cache` TTL evicter goroutine races with `CooldownQueue.Stop()` during SIGTERM, panicking with `send on closed channel`.

A `select { case <-done: }` guard is insufficient: Go evaluates **all** select-case operands before choosing — so if `innerChan` is already closed when the evicter callback fires, the send-case evaluation panics before `<-done` can be taken.

## Approach

Split the single channel into two, owned by different actors:

| Channel | Writer | Closed by |
|---|---|---|
| `intermediate` (`innerChan`) | evicter callback | **never** — eliminates the panic |
| `result` (`resultChan`, consumer-facing) | relay goroutine | relay goroutine on `done` signal |

A new relay goroutine forwards `intermediate → result` and `defer close(result)` unblocks any `for range` consumer on shutdown.

```go
// intermediate is NEVER closed, so the evicter callback cannot panic.
callback := func(key, value any) {
    select {
    case <-done:
    case intermediate <- value.(T):
    }
}

// Relay closes result when done fires, unblocking consumers cleanly.
go func() {
    defer close(result)
    for {
        select {
        case <-done:
            return
        case v := <-intermediate:
            select {
            case <-done:
                return
            case result <- v:
            }
        }
    }
}()
```

`Stop()` now just `close(done)` — no direct channel close.

## Additional fix

- `closed bool` → `atomic.Bool` to eliminate a pre-existing data race between `Stop()` and concurrent `Enqueue()`/`Closed()` callers.

## Testing

Adds `TestCooldownQueue_StopDuringEviction`, which races `Stop()` against the evicter callback 50 times. The full package passes under `go test -race`.

---

This is the upstream version of armosec/private-node-agent#370 (which applied the same fix as a local override because `kubescape/node-agent` is consumed there as a non-vendored dependency). Fixes the shutdown panic tracked in armosec/private-node-agent#368.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved queue shutdown behavior to avoid panics or blocking during item eviction.
  * Separated internal event forwarding from the consumer-facing output so results close safely after shutdown.

* **Tests**
  * Added coverage for stopping the queue while evictions are in progress to verify stable behavior under concurrency.

### DIFF
--- a/pkg/cooldownqueue/cooldownqueue.go
+++ b/pkg/cooldownqueue/cooldownqueue.go
@@ -1,6 +1,7 @@
 package cooldownqueue
 
 import (
+	"sync/atomic"
 	"time"
 
 	"istio.io/pkg/cache"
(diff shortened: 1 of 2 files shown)


matthyx:
## Review: no blockers

I traced the concurrency design and ran `go test -race -count=1 ./pkg/cooldownqueue/...` (passes, ~11s) plus `go vet`. The fix is correct.

**Why it works:**
- `intermediate` (`innerChan`) is written only by the evicter callback and never closed → the `send on closed channel` panic is structurally impossible.
- `result` (`resultChan`) has a single owner (the relay goroutine), which both sends to it and closes it on `done` → no concurrent send/close race.
- The callback's `select { case <-done: case intermediate <- v }` can never block after shutdown: once `done` is closed, `<-done` is always ready, so the callback returns immediately even if the relay has already exited.
- The relay's inner `select { case <-done: case result <- v }` guarantees a blocked send to a slow/absent consumer is unblocked on shutdown, and the `defer close(result)` ends the consumer's `for range` (the `watch()` loop in `dynamicwatcher`).
- `Stop()` is idempotent via `closed.Swap(true)`, and the `closed bool → atomic.Bool` change closes the pre-existing `Stop()`/`Enqueue()`/`Closed()` data race. The remaining `Enqueue` TOCTOU after `Stop` is now harmless (worst case a `seenEvents.Set` whose later eviction hits the `<-done` fast path).

**Non-blocking notes (optional, no change required):**
1. `assert.NotPanics(t, func() { q.Stop() })` is slightly misleading — the panic it guards against would fire in the evicter goroutine, not inside the closure. The test still catches it (a goroutine panic crashes the test binary), so coverage is real; only the assertion framing is off.
2. Pre-existing and out of scope for this PR: `seenEvents` (the istio TTL cache) has no stop, so its evicter goroutine keeps firing after `Stop()`, each call now taking the `<-done` fast path. Bounded and harmless, but a lingering goroutine — worth a follow-up if these queues are created/destroyed frequently.

LGTM to merge.

matthyx:
Addressed both review nitpicks in 38d5dcaf:

- **`Stop()` doc comment** — reworded to state the relay exits immediately on `done`, so pending evictions not yet forwarded may be dropped rather than drained.
- **`TestCooldownQueue_StopDuringEviction`** — each iteration now sleeps `shortCooldown + shortInterval` after `Stop()` so the final iteration still exercises the post-stop eviction callback path.

Still passes under `go test -race -count=1 ./pkg/cooldownqueue/...`. Both commits are signed off.

PR #854, opened 2026-07-20
fix: validate rootfs diff-ids before use to prevent SBOM panic on digest-less images

### LINKED ISSUE #853: Node-Agent crash: panic no ':' separator in digest
Description
Node-Agent pod restarts many times. It crashes with a panic when trying to process container image metadata. The error says:
`panic: no ':' separator in digest ""`
This happens when SBOM manager works with images that do not have a proper digest.

Environment
- OS: OpenShift 4.x (RHEL CoreOS)
- Kubescape Version: Operator 1.40.2, Node-Agent image quay.io/kubescape/node-agent:v0.3.119
- Cluster: On-prem OpenShift (not cloud)

Steps To Reproduce

- Install Kubescape Operator with Node-Agent in OpenShift.
- Run workloads that use images without digest (for example :latest).
- Node-Agent tries to generate SBOM.
- Pod crashes and restarts.

Expected Behavior
Node-Agent should continue running even if an image has no digest. SBOM manager should skip or handle this case safely.

Actual Behavior
Pod crashes with panic and restarts many times. Logs show:
`panic: no ':' separator in digest ""`

matthyx:
## Overview
Node-agent crashed with `panic: no ':' separator in digest ""` when the SBOM manager processed a container image without a resolvable digest (e.g. an image referenced only by tag, such as `:latest`, which CRI-O/OpenShift can report with an empty layer diff-id). The empty digest string reached `go-digest`'s `Digest.Algorithm()`/`Digest.Hex()`, which panic on malformed input instead of returning an error, taking down the whole node-agent process.

This adds a `validateDiffIDs` helper (using the digest library's own non-panicking `Validate()` method) and calls it in `NewSource` before any diff-id is touched. Images with an invalid/empty diff-id now surface as a reported SBOM-generation failure through the existing `reportFailure` error path instead of crashing the process — the agent keeps running.

## Additional Information
This change is deliberately scoped to `pkg/sbommanager/v1/syftutil/source.go` + its test file only — no changes to `sbom_manager.go`. During root-cause investigation a **separate, pre-existing** defect was found (SBOM-generation-failure objects aren't marked terminal and get silently reprocessed/re-logged on every container restart); that is out of scope here and is addressed in a follow-up PR.

## How to Test
```
go test ./pkg/sbommanager/v1/syftutil/...
go build ./...
```
`Test_NewSource_InvalidDiffID` reproduces the exact crash input (`diff_ids: [""]`) and asserts no panic + a wrapped error.

## Related issues/PRs
* Resolved #853

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of image filesystem layer identifiers.
  * Invalid image metadata now returns a clear error instead of risking a crash.
  * Added detailed error reporting to identify the problematic layer identifier.

### DIFF
--- a/pkg/sbommanager/v1/syftutil/source.go
+++ b/pkg/sbommanager/v1/syftutil/source.go
@@ -42,6 +42,9 @@ func NewSource(imageName, imageDigest, imageID string, imageStatus *runtime.Imag
 	if err != nil {
 		return nil, fmt.Errorf("unable to unmarshal image info: %w", err)
 	}
+	if err := validateDiffIDs(imageInfo.ImageSpec.RootFS.DiffIDs); err != nil {
+		return nil, fmt.Errorf("invalid image diff-ids: %w", err)
+	}
 	reverseLayers := imageInfo.ImageSpec.RootFS.DiffIDs
 	slices.Reverse(reverseLayers)
 	configFile := v1.ConfigFile{
@@ -127,6 +130,15 @@ func toConfig(config imagespec.ImageConfig) v1.Config {
 		StopSignal:      config.StopSignal,
 		Shell:           nil,
 	}
+}
+
+func validateDiffIDs(ds []imagedigest.Digest) error {
+	for i, d := range ds {
+		if err := d.Validate(); err != nil {
+			return fmt.Errorf("invalid diff-id at index %d (%q): %w", i, d, err)
+		}
+	}
+	return nil
 }
 
 func toDiffIDs(ds []imagedigest.Digest) []v1.Hash {
--- a/pkg/sbommanager/v1/syftutil/source_test.go
+++ b/pkg/sbommanager/v1/syftutil/source_test.go
@@ -6,6 +6,7 @@ import (
 	"github.com/anchore/syft/syft/source"
 	imagedigest "github.com/opencontainers/go-digest"
 	"github.com/stretchr/testify/assert"
+	runtime "k8s.io/cri-api/pkg/apis/runtime/v1"
 )
 
 func Test_toLayers(t *testing.T) {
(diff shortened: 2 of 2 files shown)


matthyx:
## Review: no blockers

Verified locally on `fix/issue-853-digest-panic` (go1.25.8):

- `go build ./pkg/sbommanager/... ./pkg/sbomscanner/...` — clean.
- `go test ./pkg/sbommanager/v1/syftutil/...` — pass.
- **The regression test is real.** I removed just the `validateDiffIDs` call from `NewSource` (keeping the helper so it still compiles) and re-ran `Test_NewSource_InvalidDiffID`: it fails with the exact `panic: no ':' separator in digest ""` from #853, caught by `assert.NotPanics`. With the call restored it passes.
- **The fix covers every panic site in the repo.** `d.Algorithm()` / `d.Hex()` at `pkg/sbommanager/v1/syftutil/source.go:148-149` are the only two panicking `go-digest` calls in `pkg/` (no other `.Hex()`, `.Algorithm()`, `.Encoded()`, or `digest.Parse` outside tests), and both are reachable only through `toRootFS` → `toDiffIDs`, which now runs after validation.
- **Both callers handle the new error path.** `pkg/sbommanager/v1/sbom_manager.go:458` routes it to `reportFailure(..., ReasonSBOMGenerationFailed, ...)`; `pkg/sbomscanner/v1/server.go:57` returns `codes.Internal`. Neither swallows it, neither retries in a tight loop.

Scope is tight and the `Validate()`-instead-of-manual-parsing choice is the right one. Good to merge.

### Non-blocking notes

1. **`imageStatus.Image` is still dereferenced without a nil check** (`source.go:79,82,87`). CRI's `ImageStatus` is specified to return a response with a nil `Image` rather than an error when the image isn't in the store — the same digest-less/`:latest` situation that produced #853. In practice the `Info["info"]` unmarshal fails first (nil map → `""` → "unexpected end of JSON input"), so it's shadowed today, but that's incidental. A `if imageStatus == nil || imageStatus.Image == nil { return nil, fmt.Errorf(...) }` at the top of `NewSource` would close the same crash class for one line.

2. **`Validate()` is slightly stricter than panic-avoidance requires.** It returns `ErrDigestUnsupported` for a well-formed digest whose algorithm isn't registered in `go-digest` (i.e. anything outside sha256/sha384/sha512). Those inputs would *not* have panicked — `Algorithm()`/`Hex()` handle them fine — so such images now fail SBOM generation where they previously succeeded. Almost certainly irrelevant in practice, and I'd still keep `Validate()`; just noting the behavior isn't purely additive.

3. **The aliasing on the lines you're now sitting on top of.** `reverseLayers := imageInfo.ImageSpec.RootFS.DiffIDs; slices.Reverse(reverseLayers)` (`source.go:48-49`) reverses in place, so the struct field itself is mutated — and `toRootFS` (line 58, feeding `rawConfig`) and `toLayers` (line 68) both read it *after* the reversal. `toLayers`'s reverse mount indexing (`ms[msLen-i-1]`) means the size pairing works out, but the `RootFS.DiffIDs` written into `RawConfig` end up in the opposite order from the OCI base→top convention. Pre-existing and out of scope for this PR — flagging it because it's easy to miss and the new validation call is now the line above it. A `slices.Clone` before reversing would make the intent explicit either way.

4. **Merge-order thought on the follow-up you mention.** This PR turns a crash into a reported failure, which is strictly better — but for the exact OpenShift `:latest` fleet in #853 it means every one of those containers now re-enters SBOM generation and re-reports a failure on every restart, since the failure objects aren't marked terminal. Worth landing the follow-up close behind this one so the reporter doesn't trade a crashloop for a log/report loop.

PR #855, opened 2026-07-20
fix: persist terminal status on generic SBOM-generation failures to stop reprocessing loop

matthyx:
## Overview
SBOM-generation failures (sidecar scan error, invalid/unparseable image source, syft cataloging error, oversized image) are reported via `reportFailure`, but the reserved SBOM object was never marked with a terminal status. Since the reprocessing switch in `processContainerWithMetadata` only special-cased `TooLarge` and `Learning`, a permanently-failing image fell through to the `default:` "processing was interrupted, retrying" branch and was silently reprocessed on **every** subsequent container start — repeated identical error logs and repeated backend failure reports, forever.

## What changed
- **Bounded retries for content-free images**: a new `Incomplete` case in the reprocessing switch, version-gated like the existing `Learning` case, plus a `failureRetries` counter (`handleGenericFailure`) that only pins an image `Incomplete` after `maxScanRetries` (3) *consecutive* failures — a single transient error (a sidecar blip, a registry hiccup) is self-healing, not permanent. `failureRetries` is a bounded, TTL'd `expirable.LRU` (1000 entries, 30 min) rather than a plain map, so it doesn't leak entries for images seen once and never again.
- **Content preservation**: marking a terminal status now goes through a new `storage.SbomClient.PatchSBOMAnnotations` — a JSON merge patch on `metadata.annotations` only, which structurally can never send `spec`. This matters because the reprocessing path fetches the SBOM via `GetSBOMMeta`, which the storage layer returns **without its Spec** (a metadata-only fetch) — persisting that object via a full `ReplaceSBOM` would silently destroy a previously-successful SBOM's real content (its CVE-scan data) the moment a later reprocess attempt failed. All SBOM-generation failure paths (generic scan/source/syft errors, repeated sidecar OOM crashes, and `ErrImageTooLarge`) now go through this patch-based marking, so retries are bounded uniformly whether or not the image previously had content — no separate case for "already had good data" is needed anymore.

## Residual trade-off (documented, not fixed here)
`failureRetries`' TTL resets on every write, so the retry count only accumulates for failures spaced under 30 minutes apart. This fully closes the loop for the reported symptom (a tight failure/crash-loop cadence) but a permanent failure recurring more slowly than that (e.g. an infrequently-restarted long-lived pod) could in principle still reprocess indefinitely without ever hitting 3 *consecutive* failures. Flagged for follow-up if it turns out to matter in practice.

## Additional Information
This is a **pre-existing, independent defect** discovered while working on #853/#854 (the digest-panic crash fix) — it already affected ordinary syft SBOM-generation failures, unrelated to digests. It's deliberately split into its own PR to keep each change minimal and independently reviewable, and does not depend on #854.

## How to Test
```
go build ./...
go vet ./...
go test ./pkg/sbommanager/... ./pkg/storage/...
```
`Test_processContainerWithMetadata_IncompleteReprocessing` reproduces the bounded-retry behavior: `maxScanRetries-1` failures don't touch storage, the `maxScanRetries`-th does, and a version bump grants a fresh retry budget. `Test_processContainerWithMetadata_PreservesContentOn{ReprocessFailure,ScannerCrash,TooLarge}` seed a content-bearing SBOM and drive each failure path past its retry threshold, asserting the SBOM's real content survives (only its annotations change) — each was verified to actually fail (reproducing the underlying data-loss bug) when its corresponding fix is reverted.

## Related issues/PRs
* Follow-up to #854 (fix for #853) — same root investigation, split out as an independent, pre-existing defect.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved existing SBOM content when reprocessing fails, including scanner crashes and oversized images.
  * Improved retry handling for repeated SBOM-generation failures.
  * Prevented unnecessary reprocessing when the tool version has not changed.
  * Reset retry behavior when a new tool version is detected.
  * SBOMs are now marked incomplete only after repeated failures without existing content.

* **Tests**
  * Added coverage for retry limits, version changes, scanner crashes, oversized images, and content preservation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SBOM reprocessing reliability by retrying transient generation failures before marking results incomplete.
  * Preserved previously generated SBOM content when later scans fail, including scanner crashes and image-size errors.
  * Prevented repeated retries when an SBOM has already been marked incomplete for the current tool version.
  * Reset retry handling when the scanning tool version changes.
  * Freshly detected images that exceed size limits are now marked accordingly without unnecessary retries.
  * Updated status changes to avoid overwriting existing SBOM content.

### DIFF
--- a/pkg/sbommanager/v1/sbom_manager.go
+++ b/pkg/sbommanager/v1/sbom_manager.go
@@ -295,6 +295,15 @@ func (s *SbomManager) processContainerWithMetadata(notif containercollection.Pub
 		},
 	}
 	wipSbom, err = s.storageClient.CreateSBOM(wipSbom)
+	// wipSbomHadContent is true only when we're about to reprocess an SBOM that previously
+	// completed successfully (the Learning case below). It exists solely to keep a
+	// content-bearing SBOM from ever being marked TooLarge on the reprocess path: unlike
+	// Incomplete, TooLarge is a one-way door in the storage layer -- GuaranteedUpdate
+	// silently drops every future write once status=too-large is set, so persisting it here
+	// (with the real Spec still attached, since PatchSBOMAnnotations never clears it) would
+	// leave the SBOM permanently frozen with its old content, unfixable by any later version.
+	// Incomplete has no such short-circuit and stays safely retryable, so it's used instead.
+	var wipSbomHadContent bool
 	switch {
 	case k8serrors.IsAlreadyExists(err):
 		// get the existing SBOM metadata and check if it is ready or being processed by another node
@@ -333,6 +342,7 @@ func (s *SbomManager) processContainerWithMetadata(notif containercollection.Pub
 				"SBOM was created with an different version of tool, recreating it") {
 				return
 			}
+			wipSbomHadContent = true
 			// continue to create SBOM
 		case wipSbom.Annotations[helpersv1.StatusMetadataKey] == helpersv1.Incomplete:
 			if !s.shouldRetryAtCurrentVersion(wipSbom, sbomName, notif,
@@ -416,7 +426,7 @@ func (s *SbomManager) processContainerWithMetadata(notif containercollection.Pub
 				s.metrics.ObserveSBOMScanDuration("oom_killed", scanDuration)
 				s.metrics.ReportSBOMScannerRestart()
 				s.metrics.SetSBOMScannerReady(false)
-				s.handleScannerCrash(sbomName, notif, scanErr, imageTag, imageID)
+				s.handleScannerCrash(sbomName, notif, scanErr, imageTag, imageID, wipSbomHadContent)
 				return
 			}
 			s.metrics.ReportSBOMScan("error")
@@ -471,7 +481,13 @@ func (s *SbomManager) processContainerWithMetadata(notif containercollection.Pub
 				helpers.String("container", notif.Container.K8s.ContainerName),
 				helpers.String("sbomName", sbomName))
 			if errors.Is(srcErr, syftutil.ErrImageTooLarge) {
-				s.markSBOMStatus(sbomName, helpersv1.TooLarge, nil)
+				if wipSbomHadContent {
+					// don't let a content-bearing SBOM reach the TooLarge one-way door; treat
+					// it as a generic (retryable, eventually Incomplete) failure instead.
+					s.handleGenericFailure(sbomName)
+				} else {
+					s.markSBOMStatus(sbomName, helpersv1.TooLarge, nil)
+				}
 				s.reportFailure(notif, imageTag, imageID, scanfailure.ReasonImageTooLarge, srcErr)
 			} else {
 				s.handleGenericFailure(sbomName)
@@ -562,7 +578,11 @@ func (s *SbomManager) waitForSharedContainerData(containerID string) (*objectcac
 }
 
 // handleScannerCrash responds to repeated sidecar OOM crashes while scanning the same image.
-func (s *SbomManager) handleScannerCrash(sbomName string, notif containercollection.PubSubEvent, scanErr error, imageTag, imageID string) {
+// hadContent must be true when the SBOM being reprocessed previously completed successfully
+// (see the wipSbomHadContent doc comment in processContainerWithMetadata) -- in that case the
+// terminal status is Incomplete rather than TooLarge, since TooLarge is a one-way door in the
+// storage layer that would permanently freeze the SBOM's existing content.
+func (s *SbomManager) handleScannerCrash(sbomName string, notif containercollection.PubSubEvent, scanErr error, imageTag, imageID string, hadContent bool) {
 	s.scanRetries[sbomName]++
 	retryCount := s.scanRetries[sbomName]
 
@@ -576,9 +596,13 @@ func (s *SbomManager) handleScannerCrash(sbomName string, notif containercollect
 		helpers.Int("maxRetries", maxScanRetries))
 
 	if retryCount >= maxScanRetries {
-		s.markSBOMStatus(sbomName, helpersv1.TooLarge, map[string]any{
-			ScannerMemoryLimitAnnotation: fmt.Sprintf("%d", s.scannerMemLimit),
-		})
+		if hadContent {
+			s.markSBOMStatus(sbomName, helpersv1.Incomplete, nil)
+		} else {
+			s.markSBOMStatus(sbomName, helpersv1.TooLarge, map[string]any{
+				ScannerMemoryLimitAnnotation: fmt.Sprintf("%d", s.scannerMemLimit),
+			})
+		}
 		// Report OOM regardless of persist success — the user should know the scan failed
 		s.reportFailure(notif, imageTag, imageID, scanfailure.ReasonScannerOOMKilled, scanErr)
 		delete(s.scanRetries, sbomName)
--- a/pkg/sbommanager/v1/sbom_manager.go
+++ b/pkg/sbommanager/v1/sbom_manager.go
@@ -88,6 +88,7 @@ type SbomManager struct {
 	scannerClient      sbomscanner.SBOMScannerClient
 	scannerMemLimit    int64
 	scanRetries        map[string]int // safe without mutex: only accessed from pool workers (pool size 1)
+	failureRetries     map[string]int // consecutive generic SBOM-generation failures per sbomName; safe without mutex: only accessed from pool workers (pool size 1)
 	pendingScans       map[string]pendingScan
 	pendingOrder       []string
 	pendingMu          sync.Mutex
@@ -141,6 +142,7 @@ func CreateSbomManager(ctx context.Context, cfg config.Config, socketPath string
 		scannerClient:      scannerClient,
 		scannerMemLimit:    scannerMemLimit,
 		scanRetries:        make(map[string]int),
+		failureRetries:     make(map[string]int),
 		pendingScans:       make(map[string]pendingScan),
 		failureReporter:    failureReporter,
 		metrics:            metrics,
@@ -283,6 +285,12 @@ func (s *SbomManager) processContainerWithMetadata(notif containercollection.Pub
 		},
 	}
 	wipSbom, err = s.storageClient.CreateSBOM(wipSbom)
+	// wipSbomHadContent is true only when we're about to reprocess an SBOM that previously
+	// completed successfully (the Learning case below). GetSBOMMeta fetches metadata only, so
+	// wipSbom.Spec is empty here even though the stored object has real content -- if
+	// reprocessing fails, that content must be left untouched rather than wiped via
+	// markSBOMStatus/ReplaceSBOM. See handleGenericFailure.
+	var wipSbomHadContent bool
 	switch {
 	case k8serrors.IsAlreadyExists(err):
 		// get the existing SBOM metadata and check if it is ready or being processed by another node
@@ -316,45 +324,19 @@ func (s *SbomManager) processContainerWithMetadata(notif containercollection.Pub
 				helpers.String("nodeName", wipSbom.Annotations[NodeNameMetadataKey]))
 			return
 		case wipSbom.Annotations[helpersv1.StatusMetadataKey] == helpersv1.Learning:
-			// only skip if the SBOM was created with the same version of tool
-			if wipSbom.Annotations[helpersv1.ToolVersionMetadataKey] == s.version {
-				logger.L().Debug("SbomManager - SBOM is already created, skipping",
-					helpers.String("namespace", notif.Container.K8s.Namespace),
-					helpers.String("pod", notif.Container.K8s.PodName),
-					helpers.String("container", notif.Container.K8s.ContainerName),
-					helpers.String("sbomName", sbomName))
+			if !s.shouldRetryAtCurrentVersion(wipSbom, sbomName, notif,
+				"SBOM is already created, skipping",
+				"SBOM was created with an different version of tool, recreating it") {
 				return
 			}
-			logger.L().Debug("SbomManager - SBOM was created with an different version of tool, recreating it",
-				helpers.String("namespace", notif.Container.K8s.Namespace),
-				helpers.String("pod", notif.Container.K8s.PodName),
-				helpers.String("container", notif.Container.K8s.ContainerName),
-				helpers.String("sbomName", sbomName),
-				helpers.String("got version", wipSbom.Annotations[helpersv1.ToolVersionMetadataKey]),
-				helpers.String("expected version", s.version))
-			// update the version of the tool
-			wipSbom.Annotations[helpersv1.ToolVersionMetadataKey] = s.version
-		// continue to create SBOM
+			wipSbomHadContent = true
+			// continue to create SBOM
 		case wipSbom.Annotations[helpersv1.StatusMetadataKey] == helpersv1.Incomplete:
-			// only skip if the SBOM was created with the same version of tool, so a fixed
-			// node-agent build still retries images that previously failed
-			if wipSbom.Annotations[helpersv1.ToolVersionMetadataKey] == s.version {
-				logger.L().Debug("SbomManager - SBOM generation previously failed with this tool version, skipping",
-					helpers.String("namespace", notif.Container.K8s.Namespace),
-					helpers.String("pod", notif.Container.K8s.PodName),
-					helpers.String("container", notif.Container.K8s.ContainerName),
-					helpers.String("sbomName", sbomName))
+			if !s.shouldRetryAtCurrentVersion(wipSbom, sbomName, notif,
+				"SBOM generation previously failed with this tool version, skipping",
+				"SBOM generation previously failed with a different tool version, retrying") {
 				return
 			}
-			logger.L().Debug("SbomManager - SBOM generation previously failed with a different tool version, retrying",
-				helpers.String("namespace", notif.Container.K8s.Namespace),
-				helpers.String("pod", notif.Container.K8s.PodName),
-				helpers.String("container", notif.Container.K8s.ContainerName),
-				helpers.String("sbomName", sbomName),
-				helpers.String("got version", wipSbom.Annotations[helpersv1.ToolVersionMetadataKey]),
-				helpers.String("expected version", s.version))
-			// update the version of the tool
-			wipSbom.Annotations[helpersv1.ToolVersionMetadataKey] = s.version
 			// continue to create SBOM
 		case wipSbom.Annotations[NodeNameMetadataKey] != s.cfg.NodeName:
 			logger.L().Debug("SbomManager - SBOM is already being processed by another node, skipping",
@@ -431,7 +413,7 @@ func (s *SbomManager) processContainerWithMetadata(notif containercollection.Pub
 				s.metrics.ObserveSBOMScanDuration("oom_killed", scanDuration)
 				s.metrics.ReportSBOMScannerRestart()
 				s.metrics.SetSBOMScannerReady(false)
-				s.handleScannerCrash(sbomName, wipSbom, notif, scanErr, imageTag, imageID)
+				s.handleScannerCrash(sbomName, wipSbom, notif, scanErr, imageTag, imageID, wipSbomHadContent)
 				return
 			}
 			s.metrics.ReportSBOMScan("error")
@@ -442,7 +424,7 @@ func (s *SbomManager) processContainerWithMetadata(notif containercollection.Pub
 				helpers.String("pod", notif.Container.K8s.PodName),
 				helpers.String("container", notif.Container.K8s.ContainerName),
 				helpers.String("sbomName", sbomName))
-			s.markSBOMStatus(wipSbom, sbomName, helpersv1.Incomplete)
+			s.handleGenericFailure(wipSbom, sbomName, wipSbomHadContent)
 			s.reportFailure(notif, imageTag, imageID, scanfailure.ReasonSBOMGenerationFailed, scanErr)
 			return
 		}
@@ -486,10 +468,16 @@ func (s *SbomManager) processContainerWithMetadata(notif containercollection.Pub
 				helpers.String("container", notif.Container.K8s.ContainerName),
 				helpers.String("sbomName", sbomName))
 			if errors.Is(srcErr, syftutil.ErrImageTooLarge) {
-				s.markSBOMStatus(wipSbom, sbomName, helpersv1.TooLarge)
+				// totalSize is computed from the currently-mounted layer paths (see
+				// syftutil.toLayers), not a fixed property of the image, so this can also
+				// fire while reprocessing a content-bearing SBOM -- guard it the same way
+				// as handleGenericFailure/handleScannerCrash.
+				if !wipSbomHadContent {
+					s.markSBOMStatus(wipSbom, sbomName, helpersv1.TooLarge)
+				}
 				s.reportFailure(notif, imageTag, imageID, scanfailure.ReasonImageTooLarge, srcErr)
 			} else {
-				s.markSBOMStatus(wipSbom, sbomName, helpersv1.Incomplete)
+				s.handleGenericFailure(wipSbom, sbomName, wipSbomHadContent)
 				s.reportFailure(notif, imageTag, imageID, scanfailure.ReasonSBOMGenerationFailed, srcErr)
 			}
 			return
@@ -515,7 +503,7 @@ func (s *SbomManager) processContainerWithMetadata(notif containercollection.Pub
 				helpers.String("pod", notif.Container.K8s.PodName),
 				helpers.String("container", notif.Container.K8s.ContainerName),
 				helpers.String("sbomName", sbomName))
-			s.markSBOMStatus(wipSbom, sbomName, helpersv1.Incomplete)
+			s.handleGenericFailure(wipSbom, sbomName, wipSbomHadContent)
 			s.reportFailure(notif, imageTag, imageID, scanfailure.ReasonSBOMGenerationFailed, syftErr)
 			return
 		}
@@ -524,6 +512,7 @@ func (s *SbomManager) processContainerWithMetadata(notif containercollection.Pub
 	}
 
 	// prepare the SBOM
+	delete(s.failureRetries, sbomName)
 	delete(wipSbom.Annotations, NodeNameMetadataKey)
 	wipSbom.Spec.Metadata.Report.CreatedAt = wipSbom.CreationTimestamp
 	wipSbom.Spec.Metadata.Tool.Name = "syft"
@@ -575,7 +564,12 @@ func (s *SbomManager) waitForSharedContainerData(containerID string) (*objectcac
 	}, backoff.WithBackOff(backoff.NewExponentialBackOff()))
 }
 
-func (s *SbomManager) handleScannerCrash(sbomName string, wipSbom *v1beta1.SBOMSyft, notif containercollection.PubSubEvent, scanErr error, imageTag, imageID string) {
+// handleScannerCrash responds to repeated sidecar OOM crashes while scanning the same image.
+// hadContent must be true when wipSbom is a GetSBOMMeta-sourced object for an SBOM that
+// previously completed successfully (see handleGenericFailure) -- in that case the Spec-clearing
+// ReplaceSBOM below is skipped so existing SBOM content is never destroyed by a crash loop
+// during a version-bump reprocess; only the failure is reported.
+func (s *SbomManager) handleScannerCrash(sbomName string, wipSbom *v1beta1.SBOMSyft, notif containercollection.PubSubEvent, scanErr error, imageTag, imageID string, hadContent bool) {
 	s.scanRetries[sbomName]++
 	retryCount := s.scanRetries[sbomName]
 
@@ -589,14 +583,16 @@ func (s *SbomManager) handleScannerCrash(sbomName string, wipSbom *v1beta1.SBOMS
 		helpers.Int("maxRetries", maxScanRetries))
 
 	if retryCount >= maxScanRetries {
-		delete(wipSbom.Annotations, NodeNameMetadataKey)
-		wipSbom.Annotations[helpersv1.StatusMetadataKey] = helpersv1.TooLarge
-		wipSbom.Annotations[ScannerMemoryLimitAnnotation] = fmt.Sprintf("%d", s.scannerMemLimit)
-		wipSbom.Spec = v1beta1.SBOMSyftSpec{}
-		if _, replaceErr := s.storageClient.ReplaceSBOM(wipSbom); replaceErr != nil {
-			logger.L().Error("SbomManager - failed to mark SBOM as TooLarge after scanner crashes",
-				helpers.Error(replaceErr),
-				helpers.String("sbomName", sbomName))
+		if !hadContent {
+			delete(wipSbom.Annotations, NodeNameMetadataKey)
+			wipSbom.Annotations[helpersv1.StatusMetadataKey] = helpersv1.TooLarge
+			wipSbom.Annotations[ScannerMemoryLimitAnnotation] = fmt.Sprintf("%d", s.scannerMemLimit)
+			wipSbom.Spec = v1beta1.SBOMSyftSpec{}
+			if _, replaceErr := s.storageClient.ReplaceSBOM(wipSbom); replaceErr != nil {
+				logger.L().Error("SbomManager - failed to mark SBOM as TooLarge after scanner crashes",
+					helpers.Error(replaceErr),
+					helpers.String("sbomName", sbomName))
+			}
 		}
 		// Report OOM regardless of persist success — the user should know the scan failed
 		s.reportFailure(notif, imageTag, imageID, scanfailure.ReasonScannerOOMKilled, scanErr)
@@ -660,6 +656,53 @@ func (s *SbomManager) markSBOMStatus(wipSbom *v1beta1.SBOMSyft, sbomName, status
 			helpers.String("sbomName", sbomName),
 			helpers.String("status", status))
 	}
+}
+
+// shouldRetryAtCurrentVersion checks a status-gated SBOM's recorded tool version against the
+// running version. If they match, it logs skipMsg and returns false (the caller should skip
+// reprocessing). Otherwise it logs retryMsg, updates the tool-version annotation, and returns
+// true (the caller should continue to reprocess).
+func (s *SbomManager) shouldRetryAtCurrentVersion(wipSbom *v1beta1.SBOMSyft, sbomName string, notif containercollection.PubSubEvent, skipMsg, retryMsg string) bool {
+	if wipSbom.Annotations[helpersv1.ToolVersionMetadataKey] == s.version {
+		logger.L().Debug(skipMsg,
+			helpers.String("namespace", notif.Container.K8s.Namespace),
+			helpers.String("pod", notif.Container.K8s.PodName),
+			helpers.String("container", notif.Container.K8s.ContainerName),
+			helpers.String("sbomName", sbomName))
+		return false
+	}
+	logger.L().Debug(retryMsg,
+		helpers.String("namespace", notif.Container.K8s.Namespace),
+		helpers.String("pod", notif.Container.K8s.PodName),
+		helpers.String("container", notif.Container.K8s.ContainerName),
+		helpers.String("sbomName", sbomName),
+		helpers.String("got version", wipSbom.Annotations[helpersv1.ToolVersionMetadataKey]),
+		helpers.String("expected version", s.version))
+	wipSbom.Annotations[helpersv1.ToolVersionMetadataKey] = s.version
+	return true
+}
+
+// handleGenericFailure responds to a non-deterministic SBOM-generation failure (source
+// construction, syft cataloging, or sidecar scan error).
+//
+// If hadContent is true, we're reprocessing an SBOM that previously completed successfully;
+// wipSbom came from GetSBOMMeta, which returns metadata only (no Spec), so persisting it via
+// markSBOMStatus/ReplaceSBOM would silently wipe the existing, real SBOM content. In that case
+// the stored object is left untouched on failure -- the next container start simply retries.
+//
+// Otherwise, the SBOM never had real content (fresh reservation, or a previous attempt already
+// failed), so it's safe to persist -- but only after maxScanRetries consecutive failures, so a
+// single transient error doesn't permanently pin the image to Incomplete.
+func (s *SbomManager) handleGenericFailure(wipSbom *v1beta1.SBOMSyft, sbomName string, hadContent bool) {
+	if hadContent {
+		return
+	}
+	s.failureRetries[sbomName]++
+	if s.failureRetries[sbomName] < maxScanRetries {
+		return
+	}
+	delete(s.failureRetries, sbomName)
+	s.markSBOMStatus(wipSbom, sbomName, helpersv1.Incomplete)
 }
 
 // reportFailure sends a scan failure report to the backend via the failure reporter.
--- a/pkg/sbommanager/v1/sbom_manager.go
+++ b/pkg/sbommanager/v1/sbom_manager.go
@@ -334,6 +334,27 @@ func (s *SbomManager) processContainerWithMetadata(notif containercollection.Pub
 				helpers.String("expected version", s.version))
 			// update the version of the tool
 			wipSbom.Annotations[helpersv1.ToolVersionMetadataKey] = s.version
+		// continue to create SBOM
+		case wipSbom.Annotations[helpersv1.StatusMetadataKey] == helpersv1.Incomplete:
+			// only skip if the SBOM was created with the same version of tool, so a fixed
+			// node-agent build still retries images that previously failed
+			if wipSbom.Annotations[helpersv1.ToolVersionMetadataKey] == s.version {
+				logger.L().Debug("SbomManager - SBOM generation previously failed with this tool version, skipping",
+					helpers.String("namespace", notif.Container.K8s.Namespace),
+					helpers.String("pod", notif.Container.K8s.PodName),
+					helpers.String("container", notif.Container.K8s.ContainerName),
+					helpers.String("sbomName", sbomName))
+				return
+			}
+			logger.L().Debug("SbomManager - SBOM generation previously failed with a different tool version, retrying",
+				helpers.String("namespace", notif.Container.K8s.Namespace),
+				helpers.String("pod", notif.Container.K8s.PodName),
+				helpers.String("container", notif.Container.K8s.ContainerName),
+				helpers.String("sbomName", sbomName),
+				helpers.String("got version", wipSbom.Annotations[helpersv1.ToolVersionMetadataKey]),
+				helpers.String("expected version", s.version))
+			// update the version of the tool
+			wipSbom.Annotations[helpersv1.ToolVersionMetadataKey] = s.version
 			// continue to create SBOM
 		case wipSbom.Annotations[NodeNameMetadataKey] != s.cfg.NodeName:
 			logger.L().Debug("SbomManager - SBOM is already being processed by another node, skipping",
@@ -421,6 +442,7 @@ func (s *SbomManager) processContainerWithMetadata(notif containercollection.Pub
 				helpers.String("pod", notif.Container.K8s.PodName),
 				helpers.String("container", notif.Container.K8s.ContainerName),
 				helpers.String("sbomName", sbomName))
+			s.markSBOMStatus(wipSbom, sbomName, helpersv1.Incomplete)
 			s.reportFailure(notif, imageTag, imageID, scanfailure.ReasonSBOMGenerationFailed, scanErr)
 			return
 		}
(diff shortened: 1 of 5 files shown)


pkg/sbommanager/v1/sbom_manager.go
```diff
@@ -630,6 +648,20 @@ func (s *SbomManager) drainPendingScans() {
 	}
 }
 
+// markSBOMStatus persists the SBOM's terminal status (e.g. TooLarge, Incomplete) so a later
+// container start for the same image is handled by the matching case in
+// processContainerWithMetadata instead of retrying and failing indefinitely.
+func (s *SbomManager) markSBOMStatus(wipSbom *v1beta1.SBOMSyft, sbomName, status string) {
+	delete(wipSbom.Annotations, NodeNameMetadataKey)
+	wipSbom.Annotations[helpersv1.StatusMetadataKey] = status
+	if _, err := s.storageClient.ReplaceSBOM(wipSbom); err != nil {
```
matthyx: **Blocker — this wipes a previously-good SBOM and silently kills CVE scanning for that image.**

`ReplaceSBOM` is a full k8s `Update`, and `markSBOMStatus` never clears/repopulates `wipSbom.Spec`. In the `IsAlreadyExists` reprocessing path, `wipSbom` came from `GetSBOMMeta`, which fetches with `softwarecomposition.ResourceVersionMetadata` (`pkg/storage/v1/storage.go:93`) and therefore returns the object **without its Spec**. `SbomSyftStrategy.PrepareForUpdate` is a no-op, so nothing merges the old spec back.

Failure scenario:
1. Image scanned OK at node-agent `v1.0.0` → SBOM stored with full `Spec`, status `ready`.
2. Node-agent upgraded to `v2.0.0`. Container starts → `AlreadyExists` → `GetSBOMMeta` (**empty Spec**) → `Learning` case, version mismatch → continue to reprocess.
3. The scan fails *transiently* (sidecar gRPC blip, registry hiccup).
4. New code: `markSBOMStatus(wipSbom, ..., Incomplete)` → `Update` with an empty Spec → **the good SBOM's artifacts are destroyed** and it is pinned to `incomplete` at `v2.0.0`.
5. kubevuln then skips it forever: `core/services/scan.go` → `if sbom.Status == helpersv1.Incomplete || ... { skip }` (`ErrIncompleteSBOM`). The image loses vulnerability coverage until the next node-agent version bump.

Verified empirically — this test passes on `main` and fails on this branch:

```go
// metaOnlyClient mimics the real GetSBOMMeta, which strips Spec.
type metaOnlyClient struct{ *fakeSbomClient }

func (f *metaOnlyClient) GetSBOMMeta(name string) (*v1beta1.SBOMSyft, error) {
	f.mu.Lock(); defer f.mu.Unlock()
	if s, ok := f.sboms[name]; ok {
		c := s.DeepCopy()
		c.Spec = v1beta1.SBOMSyftSpec{} // <- what the apiserver actually returns
		return c, nil
	}
	return nil, k8serrors.NewNotFound(schema.GroupResource{Resource: "sbomsyfts"}, name)
}

func Test_goodSBOMWipedByTransientFailure(t *testing.T) {
	inner := newFakeSbomClient()
	mgr := ... // storageClient: &metaOnlyClient{inner}, version "v2.0.0", scanner always fails
	good := &v1beta1.SBOMSyft{}
	good.Name = sbomName
	good.Annotations = map[string]string{
		helpersv1.StatusMetadataKey:      helpersv1.Learning,
		helpersv1.ToolVersionMetadataKey: "v1.0.0",
	}
	good.Spec.Syft.Artifacts = make([]v1beta1.SyftPackage, 2)
	inner.sboms[sbomName] = good

	mgr.processContainerWithMetadata(notif, nil, imageStatus, imageTag, imageID)

	assert.Len(t, inner.sboms[sbomName].Spec.Syft.Artifacts, 2)
}
```

```
main:        status="ready"      toolVersion="v1.0.0"  artifacts=2   PASS
this branch: status="incomplete" toolVersion="v2.0.0"  artifacts=0   FAIL
```

Pre-PR this path just `reportFailure` + `return`, leaving the good SBOM intact — so this is a new regression, not a pre-existing one. (The existing `TooLarge` branch has the same shape, but it was reachable only via `ErrImageTooLarge`, which is deterministic per image; generalising it to *every* scan error is what makes it dangerous. Note `handleScannerCrash` deliberately does `wipSbom.Spec = v1beta1.SBOMSyftSpec{}` before its `ReplaceSBOM` — `markSBOMStatus` is missing that awareness entirely, but zeroing the spec here would be equally wrong since it would still destroy the good data.)

Suggested fixes, roughly in order of preference:
1. Only mark `Incomplete` when we actually own a *fresh reservation* — i.e. when the pre-scan status was `Initializing`. If we're reprocessing an SBOM that already had content, leave the stored object untouched on failure (report and return, as today).
2. Or update annotations only, via a JSON/merge `Patch` on the typed client (`GetStorageClient()`), so the spec can never be clobbered by a metadata-only object.

pkg/sbommanager/v1/sbom_manager.go
```diff
@@ -334,6 +334,27 @@ func (s *SbomManager) processContainerWithMetadata(notif containercollection.Pub
 				helpers.String("expected version", s.version))
 			// update the version of the tool
 			wipSbom.Annotations[helpersv1.ToolVersionMetadataKey] = s.version
+		// continue to create SBOM
+		case wipSbom.Annotations[helpersv1.StatusMetadataKey] == helpersv1.Incomplete:
```
matthyx, line 347: **Blocker (with the comment on `markSBOMStatus`) — pinning is terminal for far too long, and there's no escape hatch.**

The only retry gate here is the tool version. Compare the `TooLarge` case just above, which has a second escape hatch (`ScannerMemoryLimitAnnotation` change → retry). `Incomplete` has none, so an image that failed for a purely environmental reason — registry 503, sidecar restart, temporary `MaxImageSize`/config value, node pressure — is skipped on *every* subsequent container start until someone ships a new node-agent version. With the spec-wipe above, that also means no CVE data for that image for the whole window.

The PR description flags this trade-off as acceptable, but I don't think it is at `n=1` with no backoff, given the blast radius is "no vulnerability scanning for this image". Options:

- Reuse the existing `scanRetries`/`maxScanRetries` pattern (already in `handleScannerCrash`) and only pin to `Incomplete` after N consecutive failures. This is the smallest change that keeps the loop bounded *and* keeps transient failures self-healing.
- Or record a retry-after / attempt-count annotation and gate the skip on it, mirroring how `ScannerMemoryLimitAnnotation` gates the `TooLarge` skip.

Either way the "stop unbounded reprocessing" goal is met without making one bad minute permanent.

Nits while here:
- This case is a near-verbatim copy of the `Learning` case above (~20 lines). Worth collapsing to `case status == helpersv1.Learning || status == helpersv1.Incomplete:` with the log message varying, or extracting a small `shouldRetryAtCurrentVersion` helper.
- The stray `// continue to create SBOM` comment now sits *between* the `Learning` case body and this `case` line, so it reads as if it belongs to the new case. Move it back inside the `Learning` case.

pkg/sbommanager/v1/sbom_manager.go
```diff
@@ -421,6 +442,7 @@ func (s *SbomManager) processContainerWithMetadata(notif containercollection.Pub
 				helpers.String("pod", notif.Container.K8s.PodName),
 				helpers.String("container", notif.Container.K8s.ContainerName),
 				helpers.String("sbomName", sbomName))
+			s.markSBOMStatus(wipSbom, sbomName, helpersv1.Incomplete)
```
matthyx: Minor, but worth confirming: this now produces **two** backend failure reports for one failure. node-agent reports `ReasonSBOMGenerationFailed` here, and once the object is persisted as `incomplete`, kubevuln's `scan_failure_reasons.go` maps `helpersv1.Incomplete` → `ReasonSBOMIncomplete` and reports again when it picks the SBOM up. Previously the object stayed `initializing` and kubevuln never reported. Since reducing duplicate backend reports is one of the stated goals of this PR, it'd be good to check that this doesn't just move the duplication one layer down.

pkg/sbommanager/v1/sbom_manager_reprocessing_test.go
```diff
@@ -0,0 +1,155 @@
...
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	if _, ok := f.sboms[sbom.Name]; ok {
+		return nil, k8serrors.NewAlreadyExists(schema.GroupResource{Resource: "sbomsyfts"}, sbom.Name)
+	}
+	f.sboms[sbom.Name] = sbom.DeepCopy()
+	return sbom, nil
+}
+
+func (f *fakeSbomClient) GetSBOMMeta(name string) (*v1beta1.SBOMSyft, error) {
```
matthyx, line 59: **Test gap — this fake can't catch the bug in `markSBOMStatus`.**

`fakeSbomClient.GetSBOMMeta` returns `s.DeepCopy()` with the **full Spec**, but the real `Storage.GetSBOMMeta` fetches with `softwarecomposition.ResourceVersionMetadata` and returns the object with an **empty Spec** (`pkg/storage/v1/storage.go:93`). That divergence is exactly what hides the data-loss regression: the test only ever exercises the freshly-`CreateSBOM`'d object, never the metadata-only one from the reprocessing path.

Please make the fake mirror the real contract (`c.Spec = v1beta1.SBOMSyftSpec{}` before returning), and add a case that starts from a pre-existing `ready` SBOM with a populated `Spec.Syft.Artifacts` at an older tool version, fails the scan, and asserts the artifacts survive.

Also: `newTestManager` leaves `scanRetries` nil. It's unreachable in the current test (the fake scanner never returns `ErrScannerCrashed`), but `handleScannerCrash` does `s.scanRetries[sbomName]++`, which panics on a nil map — easy footgun for whoever extends this table next. Worth initialising it.

matthyx (commented):
Reviewed the change end-to-end against `main` (built, vetted and ran `./pkg/sbommanager/...` on both branches). The diagnosis is right — the `default:` "processing was interrupted, retrying" fall-through really does reprocess permanently-failing images on every container start, and `Incomplete` is the semantically correct status (kubevuln's `core/services/scan.go` explicitly skips `Incomplete`/`TooLarge`, and `scan_failure_reasons.go` maps it to `ReasonSBOMIncomplete`). But treat this as *changes requested*: as written the fix introduces a worse regression than the bug it fixes.

**Blockers**

1. **Silent SBOM data loss + loss of CVE coverage.** `markSBOMStatus` does a full `ReplaceSBOM` on a `wipSbom` that, in the reprocessing path, came from `GetSBOMMeta` and therefore has **no Spec**. A previously-good SBOM is overwritten with an empty one and pinned to `incomplete`, after which kubevuln skips CVE scanning for that image. Reproduced with a test that passes on `main` and fails here — details inline on `markSBOMStatus`.
2. **The pin is effectively permanent for transient failures.** The only retry gate is a tool-version bump; unlike the `TooLarge` case there's no secondary escape hatch. A registry 503 or a sidecar restart costs that image its SBOM until the next node-agent release. Inline suggestion: reuse the existing `scanRetries`/`maxScanRetries` pattern so we only pin after N consecutive failures — that still bounds the loop, which is the actual goal.

**Non-blocking**

3. Possible double failure-reporting to the backend (node-agent `ReasonSBOMGenerationFailed` + kubevuln `ReasonSBOMIncomplete`).
4. Test fake diverges from the real `GetSBOMMeta` contract, which is why the regression above isn't caught; plus `scanRetries` is left nil in `newTestManager`.
5. The new `Incomplete` case is a near-copy of the `Learning` case, and a `// continue to create SBOM` comment got orphaned between the two.

**CI:** the DCO check is `ACTION_REQUIRED` — commits need a `Signed-off-by` line before this can merge.

Happy to re-review quickly once (1) and (2) are addressed — the core insight is solid and I'd like to see it land.

pkg/sbommanager/v1/sbom_manager.go
matthyx: Confirmed — verified `GetSBOMMeta`'s metadata-only fetch directly in `pkg/storage/v1/storage.go:93`. Fixed in `db239df2`: `markSBOMStatus` is no longer called unconditionally. I now track whether the SBOM being reprocessed had prior successful content (`wipSbomHadContent`, set only when the `Learning` case takes its version-mismatch retry branch — that's the only path where `wipSbom` can carry real server-side content it wasn't given) and skip the destructive `ReplaceSBOM` whenever it did, reporting the failure but otherwise leaving the stored object untouched, per your suggested fix #1.

While auditing every reachable persist path for the same defect, I also found (and fixed) two more sites with the identical unguarded pattern that a second and third review pass caught: `handleScannerCrash` (clears `Spec` + `ReplaceSBOM` after repeated OOM crashes) and the `ErrImageTooLarge` branch (whose `totalSize` is computed from the currently-mounted layer paths, not a fixed image property, so it's reachable during a content-bearing reprocess too, contrary to my earlier assumption that it was safely deterministic). All three now share the same guard. Added `Test_processContainerWithMetadata_PreservesContentOnReprocessFailure/OnScannerCrash/OnTooLarge`, each empirically confirmed to reproduce your exact failure scenario (artifacts wiped, status flipped, `ReplaceSBOM` called) when the corresponding guard is removed.

One edge case I'm flagging rather than fixing: on the *success* path, if a fresh reprocess actually runs and produces a real (non-empty) `syftDoc` that then exceeds `MaxSBOMSize`, the code still clears `Spec` and marks `TooLarge` unconditionally (pre-existing behavior, `sbom_manager.go` around the size-check after "prepare the SBOM"). That's a different flavor — a real new result being rejected for size, not a stale empty object clobbering a good one — but the net effect (an older, smaller, valid SBOM getting replaced by nothing) is similar. Happy to guard that too if you think it's in scope here, or file it separately.

pkg/sbommanager/v1/sbom_manager.go
matthyx: Addressed in `db239df2`:

- **Bounded retry**: added a `failureRetries map[string]int` counter (separate from `scanRetries`, which `handleScannerCrash` already owns for OOM crashes specifically) and a new `handleGenericFailure` helper. A content-safe (no prior content) generic failure now only gets pinned to `Incomplete` after `maxScanRetries` (3) *consecutive* failures — reused the existing constant/pattern as you suggested, rather than inventing a new one. Failures below that threshold report but don't touch storage, so a transient blip self-heals on the next container start. The counter resets on success and (implicitly) whenever a version bump gives an `Incomplete` image a fresh budget.
- **Style nits**: extracted `shouldRetryAtCurrentVersion(wipSbom, sbomName, notif, skipMsg, retryMsg) bool`, now shared verbatim by both the `Learning` and `Incomplete` cases — eliminates the ~20-line duplication rather than just collapsing the `case` condition. The stray `// continue to create SBOM` comment placement is fixed as a side effect of the refactor (each case now owns its own trailing comment unambiguously).

Test coverage for the bounded-retry semantics is in `Test_processContainerWithMetadata_IncompleteReprocessing` (rewritten to assert `maxScanRetries-1` failures don't touch storage, the `maxScanRetries`-th does, and a version bump grants a fresh budget rather than re-pinning immediately).

pkg/sbommanager/v1/sbom_manager.go
matthyx: Fair point, and I don't have visibility to fully resolve it from this repo alone — kubevuln's report-on-`Incomplete` behavior is outside this PR's scope. What I can say: with the bounded-retry fix above, node-agent's own reporting is no longer unbounded (previously it reported once per attempt, forever; now it's capped at `maxScanRetries` reports before the image is pinned and stops reprocessing), so the total report volume from node-agent's side is strictly lower than before regardless of what kubevuln does on top. Whether kubevuln's `ReasonSBOMIncomplete` report is itself redundant with node-agent's `ReasonSBOMGenerationFailed` is a call for that side to make — flagging it here in case it's useful context if you or someone follows up there.

pkg/sbommanager/v1/sbom_manager_reprocessing_test.go
matthyx: Done in `db239df2`:

- `fakeSbomClient.GetSBOMMeta` now always strips `Spec` before returning (`meta.Spec = v1beta1.SBOMSyftSpec{}`), unconditionally mirroring the real `Storage.GetSBOMMeta` contract for every test, not just a special-cased wrapper.
- `newTestManager` now initializes both `scanRetries` and `failureRetries` (the new counter). Also added `newTestManagerWithScannerErr` / `newTestManagerInProcess` for the scanner-crash and too-large scenarios, all consistently initializing both maps.
- Added the exact case you described: a pre-existing `Learning`-status SBOM with `Spec.Syft.Artifacts` populated at an older tool version, driven through a failing reprocess, asserting the artifacts survive untouched (plus the scanner-crash and too-large variants of the same scenario, once the audit found those paths had the identical gap).

I verified each of the three new tests is a real regression guard, not a vacuous one — for each, I temporarily removed just its corresponding guard, confirmed the test fails with the exact symptom (artifacts wiped to 0, status flipped, `ReplaceSBOM` invoked), then restored the fix and reconfirmed the full suite is green.

matthyx:
@matthyx Thanks for the thorough review — the data-loss finding was real and important. Pushed `db239df2` addressing everything:

**Blocker 1 (data loss)**: Fixed as you suggested — never persist `wipSbom` on a failed reprocess when it came in with real prior content. Tracked via a new `wipSbomHadContent` flag, set only in the `Learning` case's version-mismatch branch (the one path where `wipSbom` can hold content it wasn't given, since `GetSBOMMeta` returns metadata only). While auditing every place that can persist `wipSbom`, I found the same unguarded pattern in two more places beyond what you flagged — `handleScannerCrash` (OOM-crash loop) and the `ErrImageTooLarge` branch (its size check depends on currently-mounted paths, not just the image itself, so it's not as deterministic as I'd assumed) — and fixed both the same way. All three are now covered by dedicated tests (`Test_processContainerWithMetadata_PreservesContentOnReprocessFailure` / `OnScannerCrash` / `OnTooLarge`), each verified to actually fail (reproducing your exact symptom) when its guard is removed.

**Blocker 2 (no escape hatch)**: Added a `failureRetries` counter and `handleGenericFailure`, reusing the `maxScanRetries` pattern from `handleScannerCrash` as you suggested — a content-free SBOM only gets pinned `Incomplete` after 3 consecutive failures, so a single transient error self-heals instead of permanently losing coverage.

**Non-blocking**: replied inline on both — the double-reporting one is now bounded on node-agent's side but the kubevuln half is out of scope here; the test-fake fix and near-duplicate-case extraction are both done.

One thing I'm flagging rather than fixing (see inline reply on the `markSBOMStatus` thread): a successful reprocess whose *real, freshly-generated* result happens to exceed `MaxSBOMSize` still clears `Spec` unconditionally today — different mechanism than the bug you found (real new data being rejected, not stale metadata clobbering good data), but similar net effect. Let me know if you'd like that folded in here or tracked separately.

All tests + `go build`/`go vet`/`gofmt` green, diff still scoped to the same two files.

pkg/sbommanager/v1/sbom_manager.go
```diff
@@ -630,6 +644,67 @@ func (s *SbomManager) drainPendingScans() {
...
+// If hadContent is true, we're reprocessing an SBOM that previously completed successfully;
+// wipSbom came from GetSBOMMeta, which returns metadata only (no Spec), so persisting it via
+// markSBOMStatus/ReplaceSBOM would silently wipe the existing, real SBOM content. In that case
+// the stored object is left untouched on failure -- the next container start simply retries.
+//
+// Otherwise, the SBOM never had real content (fresh reservation, or a previous attempt already
+// failed), so it's safe to persist -- but only after maxScanRetries consecutive failures, so a
+// single transient error doesn't permanently pin the image to Incomplete.
+func (s *SbomManager) handleGenericFailure(wipSbom *v1beta1.SBOMSyft, sbomName string, hadContent bool) {
+	if hadContent {
```
matthyx: Both of my blockers are resolved here — I re-verified independently rather than relying on the new tests, and the content-preservation guard holds even when hammered well past the retry budget (15 consecutive failures on a `ready` SBOM at a bumped version: `artifacts=2`, `status=ready`, `replaceCalls=0`).

One residual worth naming explicitly, non-blocking: `if hadContent { return }` means a previously-successful SBOM that now fails *permanently* goes back to being reprocessed on every single container start — which is the exact spam this PR set out to kill, just scoped to a narrower set of images:

```go
// content-bearing SBOM at v1.0.0, node-agent now v2.0.0, scanner always fails
for range 10 {
    mgr.processContainerWithMetadata(notif, nil, imageStatus, imageTag, imageID)
}
// scan attempts: 10  (stored tool version still "v1.0.0")
```

Ten container starts → ten full scans, ten `SbomManager - sidecar scan failed` error logs, ten `reportFailure` calls to the backend. `shouldRetryAtCurrentVersion` bumps `ToolVersionMetadataKey` **in memory only**, and since nothing is persisted on failure the stored object never advances past `v1.0.0`, so the next start makes the identical decision. Same shape in `handleScannerCrash`: with `hadContent` the OOM is still reported every `maxScanRetries` crashes, forever.

Trading unbounded logs for not destroying data is clearly the right call, so I'm fine with this shipping. But the cleaner endgame is the annotation-only update I mentioned last round: a JSON/merge `Patch` on the typed client (`GetStorageClient()`) touching only `metadata.annotations` can record the terminal status *and* the new tool version **without** ever sending a Spec, so the spec can't be clobbered by a metadata-only object. That gets you both properties at once and lets `hadContent` disappear entirely, along with the three separate guards it now threads through `handleGenericFailure` / `handleScannerCrash` / the `ErrImageTooLarge` branch.

If you'd rather keep this PR minimal (reasonable — it's already grown), the ask is just to soften the PR description: the reprocessing loop is now bounded for images that never had content, but still unbounded for images that had content and start failing.

matthyx (commented):
Re-reviewed at `db239df`. **Both blockers are fixed** — verified independently with my own tests rather than trusting the new ones:

| Check | Result |
|---|---|
| Content preserved through 15 consecutive failures on a `ready` SBOM at a bumped version | `artifacts=2`, `status=ready`, `replaceCalls=0` |
| Bounded retry on a fresh SBOM: 20 container starts, always-failing scanner | 3 scan attempts, then pinned `incomplete` and rescans stop |
| `go build` / `go vet` / `go test ./pkg/sbommanager/...` | pass |
| DCO | now `success` — both commits signed off |

The `wipSbomHadContent` approach is sound, and I like that the guard was extended to *all three* destructive paths rather than just the one I flagged — `handleScannerCrash` and the `ErrImageTooLarge` branch were both genuinely reachable while reprocessing, and the reasoning about `totalSize` being derived from mounted layer paths is correct. I also checked that `hadContent` is set on the only case that can hold real content: `Learning` is the sole status node-agent writes with a populated Spec (`TooLarge` explicitly zeroes it, `Initializing`/`Incomplete` never had one), and a completed SBOM has no `NodeName` annotation so it can't reach the `default:` branch instead. `shouldRetryAtCurrentVersion` cleans up the duplication and the orphaned comment nicely, and the test double now mirrors the real metadata-only `GetSBOMMeta` contract, with `scanRetries`/`failureRetries` initialised.

**Nothing blocking left.** Two non-blocking observations inline:

1. `hadContent` reintroduces unbounded reprocessing for previously-successful images that start failing permanently — 10 container starts still produce 10 scans, 10 error logs and 10 backend reports. Acceptable trade vs. data loss, but an annotation-only `Patch` would give you both properties and let `hadContent` go away. At minimum the PR description shouldn't claim the loop is fixed unconditionally.
2. `failureRetries` accumulates stale counters (50 entries after 50 one-shot failures).

Also still open from last round: node-agent's `ReasonSBOMGenerationFailed` plus kubevuln's `ReasonSBOMIncomplete` on the persisted object means one failure can still surface as two backend reports — now only after 3 failures instead of 1, so much less noisy, but worth a sanity check on the backend side.

CI: component tests and the benchmark job are still running; everything that has finished is green.

LGTM once you've decided what to do with (1) — happy either way, it just shouldn't be silent.

pkg/sbommanager/v1/sbom_manager.go
```diff
@@ -88,6 +88,7 @@ type SbomManager struct {
 	scannerClient      sbomscanner.SBOMScannerClient
 	scannerMemLimit    int64
 	scanRetries        map[string]int // safe without mutex: only accessed from pool workers (pool size 1)
+	failureRetries     map[string]int // consecutive generic SBOM-generation failures per sbomName; safe without mutex: only accessed from pool workers (pool size 1)
```
matthyx: Nit: this map only ever shrinks on success or on reaching `maxScanRetries`, so images that fail once or twice and are never seen again leave their counters behind forever. Easy to observe:

```
failureRetries entries after 50 distinct one-shot failures: 50
```

`scanRetries` has the same shape and predates this PR, so it's not a regression — but `failureRetries` is reachable from *every* scan error rather than just sidecar OOM crashes, so it accumulates a lot faster on a long-lived node-agent that sees many short-lived images. Worth capping it the way `pendingScans` already caps itself with `maxPendingScans`, or clearing entries when the container-collection drops the image.

matthyx:
Status re-check at `db239df` (no new commits since my last review): **CI is now fully green — 30/30 checks**, including all component tests, the benchmark job, DCO and CodeRabbit. `go build` / `go vet` / `go test ./pkg/sbommanager/...` still pass locally. No new code findings from me.

Two small things left, neither blocking:

**1. The PR description is now stale.** The "Known trade-off" paragraph describes the *old* behaviour and contradicts the shipped code:

> a one-off transient failure (e.g. a brief sidecar gRPC blip) on an image is now also skipped on the next container start at the same tool version

That is no longer true — `handleGenericFailure` gives each image `maxScanRetries` consecutive failures before pinning, so a one-off blip is fully self-healing now.

> A bounded retry-with-backoff before pinning to `Incomplete` (mirroring the existing `scanRetries`/`maxScanRetries` pattern for scanner crashes) was considered as a possible future refinement but is out of scope for this fix.

This is exactly what `db239df` implements, so it reads as an unfulfilled TODO when it's actually done.

The description also doesn't mention the content-preservation guard (`wipSbomHadContent`), which is now the more interesting half of the PR, nor the residual I flagged: a previously-successful SBOM that starts failing permanently is still reprocessed on every container start, because nothing is persisted for it. Worth a short rewrite of that section so the merge commit reflects what actually landed.

**2. I'd take @coderabbitai's `markSBOMStatus` nitpick.** Adding `wipSbom.Spec = v1beta1.SBOMSyftSpec{}` inside the helper is cheap and makes the invariant enforced rather than implicit — both sibling `TooLarge` paths already zero the Spec explicitly, and the whole point of this PR was a Spec-wipe caused by an implicit assumption about what the caller passes in. It doesn't change behaviour at any current call site.

Otherwise LGTM.

pkg/sbommanager/v1/sbom_manager.go
matthyx: Went with the "cleaner endgame" — pushed `d95955c7`. Added `storage.SbomClient.PatchSBOMAnnotations(name, annotations)`, a JSON merge patch on `metadata.annotations` only (implemented via the typed client's `.Patch(..., types.MergePatchType, ...)`), and switched all SBOM status marking to go through it instead of `ReplaceSBOM`. Since the patch payload only ever contains `{"metadata":{"annotations":{...}}}`, it structurally cannot carry `spec` — there's no path for it to clobber existing content regardless of what the caller does or doesn't know about the object. `hadContent` and every guard built on it (`handleGenericFailure`, `handleScannerCrash`, the `ErrImageTooLarge` branch) are gone entirely.

Practical effect: retries are now bounded uniformly for every image via the same `failureRetries` counter, whether or not it previously had content. I independently sanity-tested the real implementation (not just my test double) against a fake k8s clientset — created an SBOM with a `node-name` annotation, patched with `node-name: nil` + two new keys, and confirmed the null-valued key was actually deleted and the others set, both in the patch response and a subsequent `GetSBOMMeta`.

One caveat worth flagging (an independent review pass caught this): the new `failureRetries` counter uses `expirable.LRU` for the memory-leak fix below, and its TTL resets on every write — so two failures for the same image spaced more than 30 minutes apart don't accumulate toward the pin threshold. That fully closes the loop for a tight failure cadence (crash loops, the actual reported symptom) but a slow-cadence permanent failure (long-lived pod, infrequent CronJob) could still in principle reprocess indefinitely without ever hitting 3 *consecutive* failures. Documented on the field; happy to widen the TTL or decouple it from the counting logic if you'd rather close that gap too.

pkg/sbommanager/v1/sbom_manager.go
matthyx: Fixed in `d95955c7`: `failureRetries` is now an `expirable.LRU[string, int]` (`github.com/hashicorp/golang-lru/v2`, already a repo dependency, used the same way in `pkg/hostfimsensor/v1/dedup_cache.go` and `pkg/rulemanager/rulecooldown`) capped at 1000 entries with a 30-minute TTL, instead of a plain unbounded map — so one-shot failures for images never seen again get reclaimed automatically instead of accumulating forever.

Side effect worth calling out (see my reply on the sibling thread): the TTL means the counter also resets between sparse failures, so it bounds retries for a tight failure cadence but not an arbitrarily slow one. Trade-off documented in a comment on the field.

matthyx:
@matthyx Pushed `d95955c7` addressing both non-blocking items from your last review:

1. **Unbounded reprocessing for previously-successful images** — implemented the "cleaner endgame" you suggested: `storage.SbomClient` gained `PatchSBOMAnnotations`, a JSON merge patch on `metadata.annotations` only, which structurally cannot carry `spec`. All SBOM status marking now goes through it instead of `ReplaceSBOM`, so `hadContent` and every guard built on it are gone — retries are bounded uniformly for all images now, regardless of prior content.
2. **`failureRetries` leaking entries** — switched from a plain map to `expirable.LRU` (bounded at 1000 entries, 30-min TTL), matching the pattern already used elsewhere in this repo (`dedup_cache.go`, `rulecooldown`).

One new trade-off surfaced by an independent review pass and flagged inline: the LRU's TTL resets on every write, so the retry count only accumulates for failures spaced under 30 minutes apart — fully closes the loop for the reported symptom (crash loops) but not for an arbitrarily slow failure cadence. Happy to adjust if you'd like that closed too.

All tests green (including a from-scratch sanity check of the real `PatchSBOMAnnotations` implementation against a fake k8s clientset, confirming the null-value annotation deletion actually works end-to-end, not just against my own test double), `go build`/`go vet`/`gofmt` clean.

matthyx:
@matthyx Re: [#issuecomment-5022708354](https://github.com/kubescape/node-agent/pull/855#issuecomment-5022708354) — both addressed:

1. **PR description rewritten** to reflect what actually shipped: bounded `failureRetries` retry (with the TTL caveat), the `PatchSBOMAnnotations` content-preservation mechanism, and the residual slow-cadence trade-off you and I already discussed. The stale "known trade-off"/"out of scope" paragraphs from the original description are gone.
2. **The `markSBOMStatus`/CodeRabbit nitpick is now moot** — this comment was posted at `db239df`, right before I pushed `d95955c7` (the `PatchSBOMAnnotations` rewrite, in response to your two non-blocking items from the review right after this comment). `markSBOMStatus` no longer takes `wipSbom` at all; it only ever builds an annotations-only patch payload, so there's no `Spec` for it to zero or forget to zero — the invariant CodeRabbit wanted enforced is now structural rather than a helper needing to remember it.

Let me know if the updated description reads right, or if you'd like the slow-cadence TTL gap closed too.

pkg/sbommanager/v1/sbom_manager.go
```diff
@@ -464,15 +471,10 @@ func (s *SbomManager) processContainerWithMetadata(notif containercollection.Pub
...
 				helpers.String("sbomName", sbomName))
 			if errors.Is(srcErr, syftutil.ErrImageTooLarge) {
-				delete(wipSbom.Annotations, NodeNameMetadataKey)
-				wipSbom.Annotations[helpersv1.StatusMetadataKey] = helpersv1.TooLarge
-				if _, replaceErr := s.storageClient.ReplaceSBOM(wipSbom); replaceErr != nil {
-					logger.L().Ctx(s.ctx).Error("SbomManager - failed to persist TooLarge SBOM",
-						helpers.Error(replaceErr),
-						helpers.String("sbomName", sbomName))
-				}
+				s.markSBOMStatus(sbomName, helpersv1.TooLarge, nil)
```
matthyx: **New issue introduced by this commit — marking a content-bearing SBOM `TooLarge` freezes it permanently in storage.**

The patch mechanism itself is correct: `sbomsyfts.NewREST` is a plain `genericregistry.Store`, so a `MergePatchType` patch is applied server-side to the *full* current object (the registry's `getCurrentState` does a plain `Get`, not the `ResourceVersionMetadata` one), and `SbomSyftStrategy.PrepareForUpdate` is a no-op. So spec really is preserved. That part works.

But `TooLarge` is special in the storage layer in a way `Incomplete` is not. `StorageImpl.GuaranteedUpdate` short-circuits **before** running `tryUpdate`:

```go
// storage@v0.0.290 pkg/registry/file/storage.go:900-906
annotations := origState.obj.(metav1.Object).GetAnnotations()
if annotations != nil && annotations[helpersv1.StatusMetadataKey] == helpersv1.TooLarge {
    logger.L().Debug("GuaranteedUpdate - already too large object, skipping update", ...)
    v.Set(reflect.ValueOf(origState.obj).Elem())
    return nil   // <- silently drops the write, returns nil error
}
```

Once an object carries `status: too-large`, **no update or patch ever lands on it again** — silently, with no error returned to the client.

Combine that with the guard removal here and you get a one-way door on two paths — the `ErrImageTooLarge` branch above, and `handleScannerCrash` after `maxScanRetries` OOMs — both now reachable while reprocessing a previously-successful `ready` SBOM:

1. Image scanned fine at `v1.0.0`; SBOM has real artifacts.
2. Version bump to `v2.0.0` → `Learning` case reprocesses.
3. `ErrImageTooLarge` fires (your own `db239df` message correctly noted `totalSize` comes from the currently-mounted layer paths, so this is reachable), or the sidecar OOMs three times.
4. `markSBOMStatus(sbomName, TooLarge, ...)` patches `status: too-large` **and leaves the full spec in place**.

The resulting object is now:
- **skipped by kubevuln forever** — `core/services/scan.go` skips on `Incomplete || TooLarge`, so the image silently loses CVE coverage;
- **unfixable by any future node-agent version** — the `TooLarge` switch case only retries when `ScannerMemoryLimitAnnotation` changed, and even if it does retry and *succeeds*, the final `ReplaceSBOM` is silently swallowed by the short-circuit above. There is no node-agent code path that can ever repair it;
- **still carrying the full spec**, which defeats the whole point of `TooLarge` (storage's own `ObjectTooLargeError` handler calls `clearSpec(ret)` with the comment "clear spec to not bloat the storage"), and both other `TooLarge` writers (`handleScannerCrash` pre-this-commit, and the `sz > MaxSBOMSize` path at line ~530) zero the Spec explicitly.

Your new `Test_..._PreservesContentOnScannerCrash` asserts exactly this state (`status == TooLarge`, `artifacts == 2`) as the desired outcome — that's the assertion I'd revisit.

Note this is **not** a regression against `main` (there, both paths did a spec-clearing `ReplaceSBOM`, so they froze the object *and* destroyed the content). It's a regression against `db239df`, where the `hadContent` guard left the SBOM `ready` and scannable, at the cost of the reprocessing noise I flagged. Trading noisy-but-recoverable for silent-and-permanent is the wrong direction, and I don't think I made that trade-off clear enough last round — sorry.

Suggested fix, smallest first:

1. **Keep the annotation-only patch for `Incomplete`, but don't let a content-bearing SBOM reach `TooLarge` at all on the reprocess path.** `Incomplete` is retryable on a version bump and is not short-circuited by the storage layer, so it's the correct terminal status when we're not deliberately discarding content.
2. Or keep a narrow `hadContent` guard **only** for the two `TooLarge` call sites, and let `Incomplete` flow freely through the patch — that preserves everything this commit achieves for the common case.
3. Or, if you do want `TooLarge` semantics here, it must clear the spec, which means a full `ReplaceSBOM` — i.e. deliberately discarding the old SBOM. Defensible for a genuinely-too-large image, much less so for three sidecar OOMs.

I'd go with (1) or (2).

pkg/sbommanager/v1/sbom_manager.go
```diff
@@ -88,11 +92,18 @@ type SbomManager struct {
...
-	pendingMu          sync.Mutex
-	failureReporter    sbommanager.SbomFailureReporter
-	metrics            metricsmanager.MetricsManager
+	// failureRetries tracks consecutive generic SBOM-generation failures per sbomName. Bounded
+	// + TTL'd so short-lived images don't leak entries; a side effect is that the count resets
+	// if two failures for the same image are spaced more than failureRetryTTL apart, so it
+	// bounds retries for a tight failure cadence (e.g. a crash loop) rather than every possible
+	// one -- a slow-cadence permanent failure (long-lived pod, infrequent CronJob) can still
+	// reprocess indefinitely without ever accumulating enough consecutive failures to pin.
+	failureRetries  *expirable.LRU[string, int]
```
matthyx, line 101: The bounded + TTL'd LRU is a good call, and I appreciate that the TTL's downside is documented honestly rather than papered over — a slow-cadence permanent failure (long-lived pod, infrequent CronJob) never accumulating `maxScanRetries` consecutive failures within 30 minutes is a real hole, and saying so here is much better than silently pretending the loop is bounded in all cases.

Non-blocking thought for later: if that hole ever bites, persisting the attempt count as an annotation (alongside the status) would make the budget survive both the TTL and a node-agent restart, at the cost of one extra patch per failure. Not worth doing now — just noting where the ceiling is.

matthyx (commented):
Re-reviewed at `d95955c`. The annotation-only patch is the right shape and I verified the mechanism end-to-end rather than assuming it works:

- `sbomsyfts.NewREST` is a plain `genericregistry.Store` (no `immutableStorage`, no `Patch` override in the `registry.REST` wrapper), so `MergePatchType` is fully wired.
- The registry's `getCurrentState` does a plain `Get` — not the `ResourceVersionMetadata` one — so the merge patch applies to the **full** stored object, and `PrepareForUpdate` is a no-op. Spec genuinely survives.
- `null` values for annotation deletion are correct merge-patch semantics, and the mock mirrors that.
- `go build ./...`, `go vet`, `go test ./pkg/sbommanager/... ./pkg/storage/...` all pass.

Removing `hadContent` entirely, uniform retry bounding for every image, the `expirable.LRU` (with its TTL caveat documented honestly), and the tests flipping from `replaceCalls` to `patchCalls` with an explicit "must never use a full ReplaceSBOM" assertion — all good. This is a genuinely better design than `db239df`.

**One new problem, though, and I think it needs addressing before merge.** `TooLarge` is not just another status: `StorageImpl.GuaranteedUpdate` (storage@v0.0.290, `pkg/registry/file/storage.go:900`) short-circuits and **silently drops every subsequent write** to any object annotated `status: too-large`. So the two `markSBOMStatus(..., TooLarge, ...)` call sites — the `ErrImageTooLarge` branch and `handleScannerCrash` — can now take a previously-successful, content-bearing SBOM and freeze it permanently: skipped by kubevuln (no CVE coverage), unrepairable by any future node-agent version, and still carrying the full spec that `TooLarge` exists to shed.

Against `main` this is still an improvement (there it froze *and* wiped). Against `db239df` it's a step back, because the `hadContent` guard kept those SBOMs `ready` and scannable. That's on my last review for pushing on the noise without flagging the freeze property — details and three fix options inline; I'd keep the patch for `Incomplete` and just not let `TooLarge` land on a content-bearing object.

CI on this head is still running; everything green on the previous head.

pkg/sbommanager/v1/sbom_manager.go
matthyx: Went with your fix (1) — pushed `a5c562be`. Reintroduced `wipSbomHadContent`, but scoped narrowly to just the two `TooLarge` write sites this time (not threaded through every generic-failure path like the old `hadContent` was): the `ErrImageTooLarge` branch and `handleScannerCrash`'s post-`maxScanRetries` marking. When a content-bearing SBOM would otherwise hit either, it now falls back to `Incomplete` (via the same annotation-only patch, retryable, no storage-layer short-circuit) instead of `TooLarge`. Content-free images still get marked `TooLarge` immediately on either path — unchanged.

`Incomplete` keeps flowing entirely through `PatchSBOMAnnotations` as of `d95955c7`, untouched by this change — this only affects the two `TooLarge` writers.

Added a new test, `Test_processContainerWithMetadata_MarksFreshImageTooLargeImmediately`, since all the prior `TooLarge` tests happened to be content-bearing scenarios and there was no coverage left for the unchanged "fresh reservation, no content to lose" case. Verified both modified tests (`_PreservesContentOnScannerCrash`, `_PreservesContentOnTooLarge`) reproduce your exact one-way-door symptom (status flips to `too-large` instead of staying `incomplete`) when the new guard is reverted.

One thing I deliberately left alone, flagging rather than fixing: `handleScannerCrash` keeps its own independent `scanRetries` counter rather than sharing `failureRetries` with the generic-failure path. An image that accumulates some generic failures and then hits 3 OOM crashes gets a fresh crash-specific budget on top, so total attempts before a terminal status can exceed `maxScanRetries` in a mixed-failure-mode scenario. That's a pre-existing inconsistency (both counters already existed independently before this PR), not something introduced by this fix — happy to unify them as a follow-up if you'd like, but didn't want to fold another counter-architecture change into round 5.

pkg/sbommanager/v1/sbom_manager.go
```diff
@@ -571,14 +596,12 @@ func (s *SbomManager) handleScannerCrash(sbomName string, wipSbom *v1beta1.SBOMS
...
-		delete(wipSbom.Annotations, NodeNameMetadataKey)
-		wipSbom.Annotations[helpersv1.StatusMetadataKey] = helpersv1.TooLarge
-		wipSbom.Annotations[ScannerMemoryLimitAnnotation] = fmt.Sprintf("%d", s.scannerMemLimit)
-		wipSbom.Spec = v1beta1.SBOMSyftSpec{}
-		if _, replaceErr := s.storageClient.ReplaceSBOM(wipSbom); replaceErr != nil {
-			logger.L().Error("SbomManager - failed to mark SBOM as TooLarge after scanner crashes",
-				helpers.Error(replaceErr),
-				helpers.String("sbomName", sbomName))
+		if hadContent {
+			s.markSBOMStatus(sbomName, helpersv1.Incomplete, nil)
```
matthyx, line 600: Small behavioural consequence worth knowing about, not a change request: this branch drops `ScannerMemoryLimitAnnotation`, which the `TooLarge` branch below still records. So a content-bearing SBOM that OOM-crash-looped is pinned `Incomplete`, and the `Incomplete` switch case gates only on tool version — raising the scanner memory limit will no longer make node-agent retry that image; only a version bump will.

That's still strictly better than before: the old `TooLarge` marking recorded the memory-limit hint, but the storage-layer short-circuit meant that even when the hint *did* trigger a retry, a successful rescan could never be persisted. So a recoverable-in-principle path became an actually-recoverable one. Just flagging it since "bump the sidecar memory limit to recover OOM-killed images" is presumably the documented operator workflow, and it now silently doesn't apply to this subset.

If you want them symmetric, passing the same `map[string]any{ScannerMemoryLimitAnnotation: ...}` here plus a memory-limit escape hatch in the `Incomplete` case (mirroring the one in the `TooLarge` case) would do it — but that's scope creep on an already-long PR and fine as a follow-up.

matthyx (commented):
Re-reviewed at `a5c562b`. **This is the fix I was asking for, and I verified the invariant independently across every failure mode rather than trusting the updated tests.**

Hammered a content-bearing SBOM (`ready` at `v1.0.0`, 2 artifacts, node-agent at `v2.0.0`) with `maxScanRetries * 4` container starts through each of the three paths that could previously reach `TooLarge`:

| Failure mode | Final status | Artifacts | patch / replace |
|---|---|---|---|
| generic sidecar error | `incomplete` | 2 | 1 / 0 |
| scanner OOM crash loop | `incomplete` | 2 | 1 / 0 |
| `ErrImageTooLarge` on reprocess | `incomplete` | 2 | 1 / 0 |

Never `TooLarge`, content intact, exactly one patch, zero full replaces, then it stops. And the other half still holds — a fresh content-free image hitting `ErrImageTooLarge` is pinned `TooLarge` immediately with no retry budget, Spec empty.

I also re-checked the invariant across *all three* `TooLarge` writers in the file, since that's what makes the one-way door safe: lines 489 and 602 are now both guarded by `hadContent`, and the `sz > MaxSBOMSize` path at 545 explicitly clears the Spec. So **every** object that can carry `status: too-large` has an empty Spec again — which is exactly the property the storage-layer short-circuit assumes. Scoping `wipSbomHadContent` to just the two `TooLarge` sites, and letting `Incomplete` keep flowing through the annotation-only patch, is the right split.

`go build ./...`, `go vet`, `go test ./pkg/sbommanager/... ./pkg/storage/...` all pass. CI is green apart from the benchmark job, still running.

**LGTM.** Two notes inline (a memory-limit-recovery asymmetry, and a pre-existing `main` issue worth a follow-up issue rather than more churn here), plus one thing to fix in the description before merge:

The rewritten description is much better, but it's now stale by one commit. Under **Content preservation** it still says:

> All SBOM-generation failure paths (generic scan/source/syft errors, repeated sidecar OOM crashes, and `ErrImageTooLarge`) now go through this patch-based marking, so retries are bounded uniformly whether or not the image previously had content — no separate case for "already had good data" is needed anymore.

`a5c562b` reintroduced exactly that separate case, deliberately. Worth replacing that sentence with the one-way-door reasoning — it's now a load-bearing design constraint that the next person to touch `markSBOMStatus` needs to know, and the commit message explains it well.

---

**Pre-existing, out of scope — recording it so it isn't lost, since it's the last remaining way a good SBOM's content can be destroyed.** if a reprocess *succeeds* but the resulting SBOM exceeds `MaxSBOMSize`, the `sz > MaxSBOMSize` path (`sbom_manager.go:545`) clears the Spec and pins `TooLarge` via a full `ReplaceSBOM`. On the reprocess path that overwrites a previously-successful, correctly-sized SBOM with an empty one, and the storage-layer short-circuit then freezes it permanently — the same one-way door `a5c562b` closes for the failure paths, still open for the success-but-oversized path.

Arguably defensible (we did generate a real SBOM and it genuinely is too large), and unlike the failure paths there's no transient-error component here, so it's much less likely to fire spuriously. Worth a follow-up issue rather than more churn on this PR.

PR #857, opened 2026-07-20
fix: unify SBOM scan-retry budgets to bound total attempts

### LINKED ISSUE #856: SBOM manager: scanRetries and failureRetries are independent budgets, allowing more than maxScanRetries total attempts before a terminal status
## Description

`SbomManager` (`pkg/sbommanager/v1/sbom_manager.go`) tracks retries for a given image via two separate counters, keyed independently by `sbomName`:

- `scanRetries` (plain `map[string]int`) — incremented only in `handleScannerCrash`, counting consecutive sidecar OOM crashes.
- `failureRetries` (`*expirable.LRU[string, int]`) — incremented only in `handleGenericFailure`, counting consecutive generic SBOM-generation failures (source construction errors, syft cataloging errors, sidecar scan errors, and content-free `ErrImageTooLarge`).

Both are compared against the same `maxScanRetries` constant (currently `3`) to decide when to pin an image to a terminal status (`Incomplete` or `TooLarge`). Because the two counters are independent, an image whose failures alternate between the two categories can take more than `maxScanRetries` total attempts before ever reaching a terminal status.

### Example

1. Container starts 2 times, each hitting a generic scan/source/syft error → `failureRetries[sbomName]` reaches 2 (below the `maxScanRetries=3` threshold, no terminal status yet).
2. Container starts a 3rd time, but this time the sidecar OOM-crashes 3 consecutive times → `scanRetries[sbomName]` independently counts 1, 2, 3 and pins the image at its own threshold.

Total attempts before a terminal status: 5 (2 generic + 3 crash), not 3 — because `handleScannerCrash`'s crash-loop budget is layered on top of, rather than sharing, the generic-failure budget.

### Why this matters

The whole point of `maxScanRetries` (introduced in #855, alongside `failureRetries` and the annotation-only `PatchSBOMAnnotations` marking) is to bound the number of times node-agent will silently reprocess a permanently-broken image before giving up and marking it terminal. With two independent counters, that bound is only enforced per-failure-category, not per-image — a real image experiencing a mix of failure modes gets a larger effective budget than `maxScanRetries` implies, with correspondingly more redundant scan attempts, error logs, and backend failure reports before the loop is finally broken.

This is a **pre-existing inconsistency**, not a regression from #855 — `scanRetries` already existed independently before that PR; #855 just added a second, similarly-shaped counter (`failureRetries`) alongside it rather than unifying them.

### Suggested fix

Unify the two counters into a single per-`sbomName` failure budget that both `handleGenericFailure` and `handleScannerCrash` increment against, so `maxScanRetries` genuinely caps *total* attempts regardless of failure category. This likely means:
- Sharing one counter (e.g. keep `failureRetries` as the single source of truth, and have `handleScannerCrash` increment/check it too instead of its own `scanRetries` map), or
- Deciding scanner-crash and generic failures deserve genuinely separate budgets (in which case this should be a documented design decision rather than an accidental side effect, and the total combined budget should be made explicit).

### References
- Discussed during review of #855: https://github.com/kubescape/node-agent/pull/855#discussion_r3614844668
- `handleGenericFailure`: `pkg/sbommanager/v1/sbom_manager.go`
- `handleScannerCrash`: `pkg/sbommanager/v1/sbom_manager.go`

matthyx:
## Overview

`SbomManager` tracked SBOM-generation retry attempts via two independent counters, both compared against the same `maxScanRetries` (3):
- `scanRetries` (plain map, no TTL) — incremented only by `handleScannerCrash` for consecutive sidecar OOM crashes.
- `failureRetries` (TTL'd LRU) — incremented only by `handleGenericFailure` for consecutive generic SBOM-generation failures.

Because the counters were independent, an image alternating between failure categories could take more than `maxScanRetries` total attempts (e.g. 2 generic failures + 3 crashes = 5) before ever reaching a terminal status.

This PR unifies both onto the existing TTL'd `failureRetries` LRU via a shared `incrementFailureCount` helper, so the retry budget now bounds *total* attempts per `sbomName` regardless of failure category.

Two intentional behavior consequences are documented inline rather than left silent:
- Scanner-crash counting now inherits the 30-min TTL-reset that generic failures already had (a crash loop with gaps >30min between crashes won't accumulate to a pin).
- Terminal-status classification (`Incomplete` vs `TooLarge`) is now attempt-attributed — whichever handler's call crosses the combined threshold decides, based on that attempt. Safety is preserved by the existing `hadContent` gate, which is untouched: content-bearing SBOMs are always pinned `Incomplete`, never `TooLarge`.

## How to Test

```
go build ./...
go test ./pkg/sbommanager/v1/... -run . -v
```

The new `Test_processContainerWithMetadata_MixedFailureCategoriesShareBudget` reproduces the bug scenario from the issue: alternating generic-failure/scanner-crash calls now pin a terminal status at exactly the 3rd combined attempt instead of up to 5-6.

## Related issues/PRs

* Resolved #856
* Follow-up to #855, which introduced `failureRetries` alongside the pre-existing `scanRetries`

AI-skills: oh-my-claudecode:ralph,oh-my-claudecode:ai-slop-cleaner,oh-my-claudecode:cancel | cmds: /clear,/oh-my-claudecode:ralplan

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SBOM retry handling by consolidating generic failure and scanner-crash retries into shared, TTL-bounded counters, with clearer crash-loop vs. general failure classification.
  * Refined terminal status selection for contentless images: **TooLarge** is set only on a pure crash-loop backstop; otherwise **Incomplete** is used.
  * Preserved existing SBOM artifacts and updated how terminal status is written (annotation patching).
* **Tests**
  * Added regression tests covering mixed failure-category sharing, pure crash-loop **TooLarge** behavior (including scanner memory limit capture), and TTL-expiry survival across sparse restarts.

### DIFF
--- a/pkg/sbommanager/v1/sbom_manager.go
+++ b/pkg/sbommanager/v1/sbom_manager.go
@@ -91,13 +91,17 @@ type SbomManager struct {
 	version            string
 	scannerClient      sbomscanner.SBOMScannerClient
 	scannerMemLimit    int64
-	scanRetries        map[string]int // safe without mutex: only accessed from pool workers (pool size 1)
-	// failureRetries tracks consecutive generic SBOM-generation failures per sbomName. Bounded
-	// + TTL'd so short-lived images don't leak entries; a side effect is that the count resets
-	// if two failures for the same image are spaced more than failureRetryTTL apart, so it
-	// bounds retries for a tight failure cadence (e.g. a crash loop) rather than every possible
-	// one -- a slow-cadence permanent failure (long-lived pod, infrequent CronJob) can still
-	// reprocess indefinitely without ever accumulating enough consecutive failures to pin.
+	// failureRetries tracks consecutive SBOM-generation failures per sbomName, governing BOTH
+	// generic-failure (handleGenericFailure) and scanner-crash (handleScannerCrash) counting under
+	// the same TTL-reset policy -- see incrementFailureCount, the single source of truth for both.
+	// Previously scanner-crash counting lived in an untimed map that accumulated indefinitely
+	// across arbitrary time gaps; unifying onto this LRU means it now inherits the same 30-min
+	// TTL-reset that generic failures already had. That is an intentional consistency choice, not
+	// an oversight. Bounded + TTL'd so short-lived images don't leak entries; a side effect is that
+	// the count resets if two failures for the same image are spaced more than failureRetryTTL
+	// apart, so it bounds retries for a tight failure cadence (e.g. a crash loop) rather than every
+	// possible one -- a slow-cadence permanent failure (long-lived pod, infrequent CronJob) can
+	// still reprocess indefinitely without ever accumulating enough consecutive failures to pin.
 	failureRetries  *expirable.LRU[string, int]
 	pendingScans    map[string]pendingScan
 	pendingOrder    []string
@@ -151,7 +155,6 @@ func CreateSbomManager(ctx context.Context, cfg config.Config, socketPath string
 		version:            packageVersion("github.com/anchore/syft"),
 		scannerClient:      scannerClient,
 		scannerMemLimit:    scannerMemLimit,
-		scanRetries:        make(map[string]int),
 		failureRetries:     expirable.NewLRU[string, int](maxFailureRetryEntries, nil, failureRetryTTL),
 		pendingScans:       make(map[string]pendingScan),
 		failureReporter:    failureReporter,
@@ -443,7 +446,6 @@ func (s *SbomManager) processContainerWithMetadata(notif containercollection.Pub
 		}
 		s.metrics.ReportSBOMScan("success")
 		s.metrics.ObserveSBOMScanDuration("success", time.Since(scanStart))
-		delete(s.scanRetries, sbomName)
 		syftDoc = result.SyftDocument
 	} else if s.scannerClient != nil {
 		s.metrics.SetSBOMScannerReady(false)
@@ -582,9 +584,19 @@ func (s *SbomManager) waitForSharedContainerData(containerID string) (*objectcac
 // (see the wipSbomHadContent doc comment in processContainerWithMetadata) -- in that case the
 // terminal status is Incomplete rather than TooLarge, since TooLarge is a one-way door in the
 // storage layer that would permanently freeze the SBOM's existing content.
+//
+// The retry budget is now shared with handleGenericFailure via incrementFailureCount, so
+// terminal-status classification (Incomplete vs TooLarge) is attempt-attributed: whichever
+// handler's call crosses the combined threshold decides the status based on that specific
+// attempt, not a blended view of prior mixed-category attempts. Safety is preserved by the
+// hadContent gate in this same function -- content-bearing SBOMs are always pinned Incomplete,
+// never TooLarge, regardless of which handler crossed the threshold, so TooLarge's one-way-door
+// risk to real content is never in play. The one residual edge is honest to disclose: a
+// contentless SBOM can occasionally reach TooLarge via a mix of generic+crash failures (rather
+// than 3 pure crashes). That is harmless -- a contentless TooLarge freezes no real content --
+// and is not a new risk introduced by unifying the counter.
 func (s *SbomManager) handleScannerCrash(sbomName string, notif containercollection.PubSubEvent, scanErr error, imageTag, imageID string, hadContent bool) {
-	s.scanRetries[sbomName]++
-	retryCount := s.scanRetries[sbomName]
+	retryCount := s.incrementFailureCount(sbomName)
 
 	logger.L().Error("SbomManager - SBOM scanner sidecar crashed during scan",
 		helpers.Error(scanErr),
@@ -605,7 +617,6 @@ func (s *SbomManager) handleScannerCrash(sbomName string, notif containercollect
 		}
 		// Report OOM regardless of persist success — the user should know the scan failed
 		s.reportFailure(notif, imageTag, imageID, scanfailure.ReasonScannerOOMKilled, scanErr)
-		delete(s.scanRetries, sbomName)
 	}
 }
 
@@ -697,19 +708,32 @@ func (s *SbomManager) shouldRetryAtCurrentVersion(wipSbom *v1beta1.SBOMSyft, sbo
 	return true
 }
 
+// incrementFailureCount increments the shared per-sbomName failure budget (used by both
+// handleGenericFailure and handleScannerCrash) and returns the new count. Once the count
+// reaches maxScanRetries the LRU entry is removed, so a later successful attempt or a tool
+// version bump starts with a fresh budget.
+func (s *SbomManager) incrementFailureCount(sbomName string) int {
+	count, _ := s.failureRetries.Get(sbomName)
+	count++
+	if count >= maxScanRetries {
+		s.failureRetries.Remove(sbomName)
+	} else {
+		s.failureRetries.Add(sbomName, count)
+	}
+	return count
+}
+
 // handleGenericFailure responds to a non-deterministic SBOM-generation failure (source
 // construction, syft cataloging, or sidecar scan error). markSBOMStatus only ever patches
 // annotations, never Spec, so it's always safe to call regardless of whether the SBOM
 // previously had real content -- but the image is only pinned Incomplete after
 // maxScanRetries consecutive failures, so a single transient error doesn't lose coverage.
+// The retry budget is shared with handleScannerCrash via incrementFailureCount, so failures
+// alternating between generic and scanner-crash categories count against the same budget.
 func (s *SbomManager) handleGenericFailure(sbomName string) {
-	count, _ := s.failureRetries.Get(sbomName)
-	count++
-	if count < maxScanRetries {
-		s.failureRetries.Add(sbomName, count)
+	if s.incrementFailureCount(sbomName) < maxScanRetries {
 		return
 	}
-	s.failureRetries.Remove(sbomName)
 	s.markSBOMStatus(sbomName, helpersv1.Incomplete, nil)
 }
 
--- a/pkg/sbommanager/v1/sbom_manager.go
+++ b/pkg/sbommanager/v1/sbom_manager.go
@@ -66,6 +66,11 @@ const (
 	maxPendingScans               = 1000
 	maxFailureRetryEntries        = 1000
 	failureRetryTTL               = 30 * time.Minute
+	// crashLoopRetryTTL is deliberately much longer than failureRetryTTL: a sidecar OOM crash
+	// stalls the shared scanner for every image on the node, so a chronically-crashing image
+	// with restarts spaced further apart than failureRetryTTL must still eventually be pinned,
+	// or it can stall the node's scanning indefinitely.
+	crashLoopRetryTTL = 24 * time.Hour
 )
 
 // pendingScan holds the data needed to retry a container scan after the sidecar becomes ready.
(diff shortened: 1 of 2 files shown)


pkg/sbommanager/v1/sbom_manager.go
```diff
@@ -582,9 +584,19 @@ func (s *SbomManager) waitForSharedContainerData(containerID string) (*objectcac
...
+// hadContent gate in this same function -- content-bearing SBOMs are always pinned Incomplete,
+// never TooLarge, regardless of which handler crossed the threshold, so TooLarge's one-way-door
+// risk to real content is never in play. The one residual edge is honest to disclose: a
+// contentless SBOM can occasionally reach TooLarge via a mix of generic+crash failures (rather
+// than 3 pure crashes). That is harmless -- a contentless TooLarge freezes no real content --
+// and is not a new risk introduced by unifying the counter.
 func (s *SbomManager) handleScannerCrash(sbomName string, notif containercollection.PubSubEvent, scanErr error, imageTag, imageID string, hadContent bool) {
-	s.scanRetries[sbomName]++
-	retryCount := s.scanRetries[sbomName]
+	retryCount := s.incrementFailureCount(sbomName)
```
matthyx, line 599: **Blocker — a single OOM can now pin a contentless SBOM `TooLarge`.**

`retryCount` is now the *combined* budget, but the classification below it (`retryCount >= maxScanRetries` → `TooLarge` when `!hadContent`) still assumes the count was consumed by scanner crashes. It no longer is.

Concrete sequence on a brand-new image (never succeeded, so `wipSbomHadContent == false`):

1. sidecar scan fails generically (transient) → count 1
2. sidecar scan fails generically (transient) → count 2
3. sidecar OOMs **once** → `incrementFailureCount` returns 3 → `!hadContent` → `markSBOMStatus(TooLarge, {scannerMemoryLimit: ...})`

PoC (drop into `pkg/sbommanager/v1`, scanner returns generic/generic/`ErrScannerCrashed`):

```go
type genericThenCrashClient struct{ calls int }

func (a *genericThenCrashClient) CreateSBOM(_ context.Context, _ sbomscanner.ScanRequest) (*sbomscanner.ScanResult, error) {
	a.calls++
	if a.calls < 3 {
		return nil, errors.New("scan failed")
	}
	return nil, sbomscanner.ErrScannerCrashed
}
func (a *genericThenCrashClient) Ready() bool  { return true }
func (a *genericThenCrashClient) Close() error { return nil }

func Test_POC_SingleOOMAfterGenericFailuresPinsTooLarge(t *testing.T) {
	fake := newFakeSbomClient()
	mgr := &SbomManager{
		cfg:             config.Config{NodeName: "node-1"},
		ctx:             context.Background(),
		processing:      mapset.NewSet[string](),
		storageClient:   fake,
		scannerClient:   &genericThenCrashClient{},
		metrics:         metricsmanager.NewMetricsNoop(),
		version:         "v2.0.0",
		scannerMemLimit: 1024,
		failureRetries:  newFailureRetries(),
	}
	notif, imageStatus, imageTag, imageID := testNotifAndImageStatus()
	sbomName, _ := names.ImageInfoToSlug(imageTag, imageID)
	for range 3 {
		mgr.processContainerWithMetadata(notif, nil, imageStatus, imageTag, imageID)
	}
	raw := fake.get(sbomName)
	assert.Equal(t, helpersv1.TooLarge, raw.Annotations[helpersv1.StatusMetadataKey])
}
```

Result:

* this branch → `too-large`, `scannerMemoryLimit: "1024"` — **passes**
* `main` (same test, plus `scanRetries: make(map[string]int)`) → `initializing` — **fails**, no terminal status pinned at all

Why this isn't the harmless edge the PR body describes — `TooLarge` is not merely "a status that freezes no content", it's the *least recoverable* terminal state in this file:

* the `TooLarge` case in `processContainerWithMetadata` returns early and, unlike `Incomplete`, has **no tool-version escape hatch** — `shouldRetryAtCurrentVersion` is never consulted. A node-agent/syft upgrade re-attempts an `Incomplete` SBOM; it will never re-attempt a `TooLarge` one.
* the only way out is `ScannerMemoryLimitAnnotation != scannerMemLimit`, i.e. an operator changing the sidecar memory limit — for an image whose real problem was two transient scan errors.
* and per the `wipSbomHadContent` comment above, `GuaranteedUpdate` silently drops every future write once `status=too-large`, so even that escape hatch can't rewrite the object.

Net effect: two transient sidecar errors plus one unlucky OOM permanently costs that image its SBOM (and therefore its vulnerability coverage), where before it would have landed on retryable `Incomplete`. That's a behavior change the issue didn't ask for, and it lands on first-time images — the common case, since `hadContent` is only true on the `Learning` reprocess path.

Suggested fix — keep the unified budget (it's the right call), but make `TooLarge` require crash attribution rather than "the crossing attempt happened to be a crash". Cheapest version: have `incrementFailureCount` also return whether *every* counted failure in the current window was a crash (a parallel bool/count in the same LRU value, e.g. a small struct), and only take the `TooLarge` branch when that holds. Simplest version if you'd rather not carry state: pin `Incomplete` unconditionally from the shared path and treat `TooLarge` as reachable only via `syftutil.ErrImageTooLarge` / the `MaxSBOMSize` check, which are the two places it's actually true by measurement.

pkg/sbommanager/v1/sbom_manager.go
```diff
@@ -697,19 +708,32 @@ func (s *SbomManager) shouldRetryAtCurrentVersion(wipSbom *v1beta1.SBOMSyft, sbo
 	return true
 }
 
+// incrementFailureCount increments the shared per-sbomName failure budget (used by both
+// handleGenericFailure and handleScannerCrash) and returns the new count. Once the count
+// reaches maxScanRetries the LRU entry is removed, so a later successful attempt or a tool
+// version bump starts with a fresh budget.
+func (s *SbomManager) incrementFailureCount(sbomName string) int {
```
matthyx, line 721: **Design question — crash counting silently lost its only unbounded backstop.**

The struct comment is upfront that scanner-crash counting now inherits the 30-min TTL reset, and calls it an intentional consistency choice. I'd push back on the choice itself, not the disclosure.

The untimed `scanRetries` map wasn't an oversight — the failure it bounds is qualitatively different. A generic failure is cheap and image-local. A sidecar OOM kills the **shared** scanner process (`ReportSBOMScannerRestart`, `SetSBOMScannerReady(false)`), which stalls scanning for every other image on the node. The untimed counter guaranteed such an image was eventually pinned and stopped costing the node.

With the 30-min TTL, any image whose container starts are spaced >30 min apart never accumulates to a pin: a CronJob on an hourly schedule, a slow rollout, node churn. Each start OOMs the sidecar again, indefinitely, and the counter is back to zero every time. That's an availability regression for the whole node, hidden behind a per-image consistency argument.

Options, cheapest first: bump `failureRetryTTL` (30 min is tuned for transient scan errors, not for "is this image fundamentally too big for the sidecar"); or don't refresh the TTL on the crash path; or keep the crash count on a separate longer-lived TTL and only unify the *budget check*.

Minor, same function: the doc comment says removal at threshold is what lets "a later successful attempt or a tool version bump start with a fresh budget." Neither is accurate — a successful attempt resets via `s.failureRetries.Remove(sbomName)` on the success path, and a tool-version bump doesn't touch the LRU at all. Worth correcting so the next reader doesn't rely on a mechanism that isn't there.

pkg/sbommanager/v1/sbom_manager_reprocessing_test.go
```diff
@@ -328,6 +326,79 @@ func Test_processContainerWithMetadata_PreservesContentOnScannerCrash(t *testing
...
+
+	mgr := &SbomManager{
+		cfg:            config.Config{NodeName: "node-1"},
+		ctx:            context.Background(),
+		processing:     mapset.NewSet[string](),
+		storageClient:  fake,
+		scannerClient:  &alternatingScannerClient{},
+		metrics:        metricsmanager.NewMetricsNoop(),
+		version:        "v2.0.0",
+		failureRetries: newFailureRetries(),
```
matthyx: **Test gap.** The new test seeds a content-bearing SBOM, so `hadContent` is true and the crossing attempt can only ever produce `Incomplete`. It proves the budget is shared but can't see the classification change the sharing introduces — the contentless path is exactly where the blocker lives, and it's currently uncovered.

Worth adding the mirror case: no seeded SBOM (`hadContent == false`), generic → generic → crash, asserting the status the fix decides on (`Incomplete` under either suggested fix), plus that `ScannerMemoryLimitAnnotation` is absent. That test fails on this branch today, which is the point.

Nit while you're here: the struct-field comment is 12 lines and the `handleScannerCrash` comment is 16, and both spend most of that pre-arguing with a reviewer ("That is an intentional consistency choice, not an oversight", "The one residual edge is honest to disclose"). Rationale like that belongs in the PR description or the commit message; the surviving code comment should just say what the invariant is. Roughly: "Shared by handleGenericFailure and handleScannerCrash; TTL-reset, so it bounds tight failure cadences rather than every possible one." The rest is recoverable from git blame.

matthyx (commented):
Reviewed the unification itself and it's correct — the shared budget does bound total attempts, and the regression test genuinely fails on `main`. One blocker before merge, plus one design question and a test gap.

**Blocker:** sharing the *counter* also silently shares the *classification*, so a single sidecar OOM can now pin a contentless SBOM `TooLarge` — a strictly less recoverable state than the `Incomplete` it would have gotten before. Details inline; I verified this with a PoC test that passes on this branch and fails on `main`.

The PR body calls that edge "harmless -- a contentless TooLarge freezes no real content". That framing is about content preservation, but the actual cost of `TooLarge` here is *recoverability*, and by that measure it isn't harmless.

matthyx:
Thanks — the blocker's PoC and the availability point are both fair, and I was wrong to frame the contentless-crash edge as harmless.

Pushed a fix in 328e3e83:

- Added a second, dedicated `crashLoopRetries` counter that only `handleScannerCrash` increments, on a much longer TTL (24h vs the existing 30min). `TooLarge` is now reachable **only** when this crash-only counter itself crosses `maxScanRetries` — a threshold crossing driven by the shared `failureRetries` budget (which may include generic failures) always falls back to `Incomplete`, regardless of mix. Your PoC sequence (2 generic + 1 crash, contentless) now pins `Incomplete`.
- The long TTL also restores the unbounded-time protection scanner-crash counting had before #855/#857 for the shared-sidecar-availability concern you raised — a sparse-cadence crash loop (restarts spaced >30min apart) still eventually accumulates to a pin via `crashLoopRetries`, even though the shared budget keeps resetting.
- Corrected the inaccurate `incrementFailureCount` doc comment (a successful scan resets the budget explicitly on the success path; a tool-version bump doesn't touch the LRU at all) and trimmed the `failureRetries` struct comment per your comment-verbosity note.
- Added three tests: the mixed-category contentless-Incomplete regression (your exact PoC scenario, `Test_processContainerWithMetadata_MixedFailureCategoriesPinIncomplete`), a guard for the still-supported pure-crash-loop `TooLarge` path (previously untested — `Test_processContainerWithMetadata_MarksContentlessImageTooLargeOnPureCrashLoop`), and a test proving the sparse-cadence backstop actually works, using short direct-constructed TTLs with real sleeps so the shared budget provably expires between crashes while the backstop survives (`Test_processContainerWithMetadata_CrashLoopBackstopSurvivesSparseCadence`).

`go build ./...`, `go vet`, `gofmt -l`, and the full `pkg/sbommanager/v1/...` suite are all green.

pkg/sbommanager/v1/sbom_manager_reprocessing_test.go
```diff
@@ -389,3 +468,123 @@ func Test_processContainerWithMetadata_MarksFreshImageTooLargeImmediately(t *tes
...
+		cfg:              config.Config{NodeName: "node-1"},
+		ctx:              context.Background(),
+		processing:       mapset.NewSet[string](),
+		storageClient:    fake,
+		scannerClient:    &fakeScannerClient{err: sbomscanner.ErrScannerCrashed},
+		metrics:          metricsmanager.NewMetricsNoop(),
+		version:          "v2.0.0",
+		scannerMemLimit:  4096,
+		failureRetries:   expirable.NewLRU[string, int](maxFailureRetryEntries, nil, 5*time.Millisecond),
+		crashLoopRetries: expirable.NewLRU[string, int](maxFailureRetryEntries, nil, 500*time.Millisecond),
```
matthyx: Nit (non-blocking): the 500ms `crashLoopRetries` TTL is a fairly tight margin for a wall-clock-dependent test. The run needs 3 iterations plus 2×20ms sleeps to finish inside 500ms; a GC pause or a loaded CI runner stalling past that budget expires the backstop entry and flips the assertion to `Incomplete` — a confusing failure, since it looks like the feature regressed rather than the runner hiccuped.

I ran it 15× under `-race` without a flake, so this is speculative rather than observed. Still, widening to `5*time.Second` costs nothing in runtime (the test doesn't wait on this TTL, it only needs it *not* to expire) and removes the failure mode entirely. The 5ms `failureRetries` TTL is fine as-is — that one you do wait on, and the 20ms sleep gives it 4× headroom.

matthyx (commented):
Re-reviewed at `328e3e8`. **Both findings are resolved** — the fix is the right shape, and I verified it independently rather than just reading the diff.

Splitting `crashLoopRetries` out as a crash-only counter on its own longer TTL addresses both findings with one mechanism, which is neater than either option I suggested: the shared budget keeps bounding total attempts (the original #856 fix), while `TooLarge` now requires crash-attributed evidence, and the 24h TTL restores the unbounded-time backstop that the 30-min TTL had removed.

Verification — I re-ran my original PoC verbatim against this commit and wrote three more for paths your tests don't cover:

| scenario | result |
|---|---|
| 2 generic + crash, contentless (the original blocker) | `incomplete`, no memory-limit annotation, exactly 1 patch ✅ |
| crash → generic → crash, contentless (3 combined, 2 crashes) | `incomplete` ✅ |
| content-bearing + **sparse-cadence** pure crash loop | `incomplete`, 2 artifacts intact, no memory-limit annotation ✅ |
| `go vet` + full package, and 15× `-race` on the sleep-based test | clean, no flakes observed ✅ |

That third row is the combination your three new tests don't hit — the long-TTL backstop crossing its threshold on an SBOM that *does* have content. The `hadContent` gate holds there, so the backstop can't route a content-bearing SBOM into `TooLarge`'s one-way door. Good.

One interaction I chased and satisfied myself is a non-issue, noting it so it doesn't have to be re-derived later: when `handleGenericFailure` pins `Incomplete`, `crashLoopRetries` keeps its stale count for up to 24h, so in principle a later crash could cross the crash threshold on evidence that straddles a pin. It isn't reachable — an `Incomplete` pin blocks reprocessing at the same tool version, and the only escapes (version bump, memory-limit change) both imply a process restart, which starts with empty LRUs. No action needed.

Also confirmed every `SbomManager` construction site (1 production + 6 in tests) initializes the new LRU, so there's no nil-`crashLoopRetries` panic path into `handleScannerCrash`.

Only a nit left, inline. Nothing blocking — LGTM.

pkg/sbommanager/v1/sbom_manager_reprocessing_test.go
matthyx: Resolved in `332442e` — 5s, and the inline comment now explains *why* the value is generous (that the test never waits on this TTL) so it doesn't get "optimized" back down later. Stressed the retry/classification tests 20× under `-race` on the new head: clean, no flakes. Thanks.

Issue #848, opened 2026-06-27
SbomManager logs "container not found in shared data" at ERROR when a short-lived container exits early


devantler:
## Description
node-agent logs `SbomManager - container not found in shared data` at ERROR level for something that is not really an error: a container that started and then exited very quickly, before node-agent finished collecting the data it needs to scan it.

When that happens there is nothing left to scan (the container is already gone), so skipping the SBOM is the correct outcome. The problem is only that it is logged at ERROR. On a cluster that runs many short-lived pods (for example frequent jobs), every container that exits early produces an error line, so this turns into a steady stream of ERROR logs and false alerts for a harmless, expected race.

The line is here (current `main`, unchanged at the v0.3.142 release tag), `pkg/sbommanager/v1/sbom_manager.go:250`:

    sharedData, err := s.waitForSharedContainerData(notif.Container.Runtime.ContainerID)
    if err != nil {
        logger.L().Error("SbomManager - container not found in shared data",
            helpers.String("container ID", notif.Container.Runtime.ContainerID))
        return
    }

`waitForSharedContainerData` retries with a backoff and then gives up; on failure the SBOM is skipped and nothing is requeued. Other "nothing to do" cases in the same file are logged at Debug, which is why this one stands out.

One caveat I want to be upfront about: the same code path is also hit if a still-running container's data never arrives (a genuinely stuck pipeline), which would be a real, silently dropped SBOM. So I am not sure simply lowering this to Debug is right. It may be better to tell the two cases apart: "container already gone" (harmless, low level) versus "data missing for a container that is still running" (keep it visible). I am reporting the symptom and leaving the remedy to you.

## Environment
- OS: Talos Linux
- node-agent: v0.3.142 (quay.io/kubescape/node-agent:v0.3.142)

## Steps To Reproduce
1. Run node-agent on a cluster that frequently creates very short-lived containers (for example fast-exiting jobs).
2. Tail node-agent logs at the default level.
3. Watch for recurring `SbomManager - container not found in shared data` ERROR lines whose container IDs belong to containers that exited almost immediately.

## Expected behavior
A container that exits before its data is ready should not be logged as an ERROR, or it should be told apart from a genuine data-missing failure for a still-running container so that only the latter is an ERROR.

## Actual Behavior
The harmless race is logged at ERROR, adding noise at the default log level. SBOM generation otherwise works for normal-lifetime containers.

## Additional context
[...]

PR #859, opened 2026-07-21
fix(sbommanager): decouple shared-data wait from single-worker pool to stop node-wide SBOM stall (#850)

### LINKED ISSUE #850: SBOM generation stalls node-wide when a short-lived container blocks the single-worker SbomManager pool
### Summary

A single short-lived container can silently halt SBOM generation **node-wide**. When a container exits before the container watcher populates its "shared container data" (pod/workload metadata), `SbomManager.processContainer` blocks the manager's **single** worker in a ~15-minute uncancelled backoff. Because the pool size is 1, this head-of-line-blocks SBOM generation for **every other container on that node** — no `sbomsyft` CRs are written until the wait unblocks or the node-agent pod restarts. A frequently-firing short-lived workload (e.g. a CronJob every few minutes) sustains the stall indefinitely.

This is related to but distinct from #848 (closed as benign log noise). The `SbomManager - container not found in shared data` ERROR line is harmless on its own, but the *same underlying condition* also stalls SBOM generation node-wide, which is not documented anywhere and is not benign.

### Version

Observed on **v0.3.142**. The relevant code paths are unchanged through the latest tag **v0.3.147** and current `main`.

### Mechanism (code refs, as of v0.3.147 / `main`)

- `pkg/sbommanager/v1/sbom_manager.go`: the manager uses a single-worker pool (`pool: workerpool.New(1)`); `ContainerCallback` submits `processContainer` to that pool.
- `processContainer` blocks in `waitForSharedContainerData`, which polls `GetSharedContainerData` via `backoff.Retry(context.Background(), ...)` with a default `ExponentialBackOff`. Because the context is `context.Background()` (never cancelled), the retry only terminates at cenkalti/backoff's `DefaultMaxElapsedTime` = **15m**.
- For a container that exits almost immediately, shared data may never be populated (it is deleted on the container-remove event in `pkg/containerwatcher/v2/containercallback.go`), so the single worker waits the full 15m — during which all other containers' SBOMs on that node are stuck in the queue.
- The correlated `pkg/containerprofilemanager/v1/lifecycle.go` message `"timeout while adding container to the container profile manager"` (hardcoded `MaxWaitForSharedContainerData = 10m` in `helpers.go`) shares the same root cause, but it runs in a per-container goroutine and does not itself block other containers — so it is a correlated symptom, not the node-wide cause.

### Impact (observed)

On a cluster with a linkerd-meshed CronJob firing every 10 minutes (two ultra-short-lived containers per run: the job container plus its injected proxy sidecar), node-agent silently stopped writing `sbomsyft` for newly-deployed long-lived workloads that landed on the affected nodes. Downstream, those images never produced a `vulnerabilitymanifest` and never reached our vulnerability-management pipeline — while all job-liveness / pod-health checks stayed green. Diagnosis required correlating the missing SBOMs with node-agent uptime and the timeout log lines.

### Workaround

Exclude the short-lived pods via `excludeLabels` / `excludeNamespaces` — both gate `Config.IgnoreContainer` at the top of the SBOM callback, before the shared-data wait, so an excluded pod never occupies the worker. This is an effective per-workload mitigation but not a fix for the general case (any short-lived container in a monitored namespace can trigger it).

### Suggested fix

Make the SBOM shared-data wait resilient to short-lived containers, e.g.:
- cancel `waitForSharedContainerData` when the container-remove event fires (tie the wait to a context cancelled on removal), and/or
- bound the wait with a short timeout instead of the 15m backoff default, and/or
- process containers concurrently (worker pool > 1) so a single stuck container cannot head-of-line-block SBOM generation for the whole node.

Aaradhya-07:
## Overview

Current behavior: A single short-lived container can silently halt SBOM generation node-wide. SbomManager runs a single-worker pool, and the shared-container-data wait executed on that worker via backoff.Retry(context.Background(), …) with no timeout — running to backoff v5's 15m DefaultMaxElapsedTime on a context that is never cancelled. Since shared data is deleted on the container-remove event, a container that exits before its data is populated holds the sole worker for the full window, head-of-line-blocking SBOM generation for every other container on the node.

Future behavior: The shared-data wait is decoupled from the single-worker pool. ContainerCallback now performs the wait in a per-container goroutine — bounded by a timeout and cancelled early on the container-remove event — and submits only the actual SBOM work to the pool once metadata is available. A stuck/short-lived container can no longer occupy the worker (so other containers' SBOMs proceed unblocked) nor park a goroutine past its own lifetime.

## Additional Information

- waitForSharedContainerData now takes a context.Context, bounded via context.WithTimeout(s.ctx, maxWaitForSharedContainerData) instead of context.Background(). The bound is 15m, tied to the producer's budget: the shared data is populated by ContainerWatcher.setSharedWatchedContainerData, which retries via cenkalti/backoff NewExponentialBackOff (15m DefaultMaxElapsedTime). A shorter consumer window would give up while the producer is still succeeding, silently dropping the SBOM.
- The per-container wait is cancelled on the container-remove event via a map[containerID]context.CancelFunc guarded by a mutex (populated when the wait starts, drained/cancelled in ContainerCallback's EventTypeRemoveContainer branch). This keeps the fix from trading a worker stall for an unbounded goroutine leak — a short-lived container's wait ends at its actual lifetime rather than parking a goroutine (retaining notif/mounts/imageStatus) for the full timeout. A guard also prevents new work being enqueued to the pool during shutdown.
- The pool-of-1 invariant is intentionally preserved — I did not raise the worker count, because scanRetries is documented as mutex-free precisely because the pool size is 1 (pool > 1 would introduce a data race).
- The shared-data-wait failure log now distinguishes cause: Debug when the wait ends in context.DeadlineExceeded/Canceled (the benign #848 outcome for a container that exits before its data lands, including cancellation on container-remove), and Error only for genuinely unexpected failures. It also gained .Ctx(s.ctx) and namespace/pod/container fields for consistency with the rest of the file.
- Out of scope: SBOM generation remains serial (pool size 1) by design — that was not the bug; cmd/main.go's non-cancellable root context and the pre-existing Set-after-Delete ordering race between the producer and DeleteSharedContainerData are left untouched.

## How to Test

- Run the package tests (with the race detector): go test -race ./pkg/sbommanager/v1/
- Test_awaitAndSubmit_StuckContainerDoesNotBlockOthers is the #850 regression guard for the fix itself — container A's data never arrives while B's is present, and it asserts awaitAndSubmit returns without blocking and B's work reaches the pool worker while A is stuck (deleting the goroutine wrapper fails it).
- Test_waitForSharedContainerData_HonorsContextDeadline asserts the wait stops at the context deadline with context.DeadlineExceeded rather than backoff's 15m default; Test_cancelWait_StopsInFlightWait covers container-remove cancellation; Test_waitForSharedContainerData_ReturnsWhenDataArrives covers the happy path.
- In-cluster: deploy a frequently-firing short-lived workload (e.g. a CronJob every few minutes) alongside a newly-deployed long-lived workload on the same node; confirm the long-lived workload's sbomsyft CR is still written promptly instead of stalling.

## Related issues/PRs:

- Resolved #850

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the maximum wait for shared container data to 15 minutes to avoid prolonged delays.
  * Container processing now performs shared-data waiting asynchronously and is bounded by the provided timeout.
  * Container removal now properly cancels pending data waits to prevent resource leaks.
  * Waiting behavior stops promptly when timeouts expire.
* **Tests**
  * Added regression tests covering context deadline handling, immediate data return, non-blocking behavior across containers, and cancellation of pending waits.

### DIFF
--- a/pkg/sbommanager/v1/sbom_manager.go
+++ b/pkg/sbommanager/v1/sbom_manager.go
@@ -258,7 +258,11 @@ func (s *SbomManager) ContainerCallback(notif containercollection.PubSubEvent) {
 		defer cancel()
 		sharedData, err := s.waitForSharedContainerData(ctx, notif.Container.Runtime.ContainerID)
 		if err != nil {
-			logger.L().Error("SbomManager - container not found in shared data",
+			logger.L().Ctx(s.ctx).Error("SbomManager - container not found in shared data",
+				helpers.Error(err),
+				helpers.String("namespace", notif.Container.K8s.Namespace),
+				helpers.String("pod", notif.Container.K8s.PodName),
+				helpers.String("container", notif.Container.K8s.ContainerName),
 				helpers.String("container ID", notif.Container.Runtime.ContainerID))
 			return
 		}
--- a/pkg/sbommanager/v1/sbom_manager.go
+++ b/pkg/sbommanager/v1/sbom_manager.go
@@ -66,6 +66,8 @@ const (
 	maxPendingScans               = 1000
 	maxFailureRetryEntries        = 1000
 	failureRetryTTL               = 30 * time.Minute
+	// maxWaitForSharedContainerData bounds the wait for a container's shared data; mirrors containerprofilemanager's MaxWaitForSharedContainerData.
+	maxWaitForSharedContainerData = 10 * time.Minute
 )
 
 // pendingScan holds the data needed to retry a container scan after the sidecar becomes ready.
@@ -250,20 +252,20 @@ func (s *SbomManager) ContainerCallback(notif containercollection.PubSubEvent) {
 			helpers.String("container", notif.Container.K8s.ContainerName))
 		return
 	}
-	// enqueue the container for processing
-	s.pool.Submit(func() {
-		s.processContainer(notif, mounts, imageStatus)
-	}, utils.FuncName(s.processContainer))
-}
-
-func (s *SbomManager) processContainer(notif containercollection.PubSubEvent, mounts []string, imageStatus *runtime.ImageStatusResponse) {
-	sharedData, err := s.waitForSharedContainerData(notif.Container.Runtime.ContainerID)
-	if err != nil {
-		logger.L().Error("SbomManager - container not found in shared data",
-			helpers.String("container ID", notif.Container.Runtime.ContainerID))
-		return
-	}
-	s.processContainerWithMetadata(notif, mounts, imageStatus, sharedData.ImageTag, sharedData.ImageID)
+	// Wait for shared data off the single-worker pool so a short-lived container whose data never arrives can't head-of-line-block SBOM generation node-wide (#850); only real work is submitted to the pool.
+	go func() {
+		ctx, cancel := context.WithTimeout(s.ctx, maxWaitForSharedContainerData)
+		defer cancel()
+		sharedData, err := s.waitForSharedContainerData(ctx, notif.Container.Runtime.ContainerID)
+		if err != nil {
+			logger.L().Error("SbomManager - container not found in shared data",
+				helpers.String("container ID", notif.Container.Runtime.ContainerID))
+			return
+		}
+		s.pool.Submit(func() {
+			s.processContainerWithMetadata(notif, mounts, imageStatus, sharedData.ImageTag, sharedData.ImageID)
+		}, utils.FuncName(s.processContainerWithMetadata))
+	}()
 }
 
 func (s *SbomManager) processContainerWithMetadata(notif containercollection.PubSubEvent, mounts []string, imageStatus *runtime.ImageStatusResponse, imageTag, imageID string) {
@@ -568,8 +570,8 @@ func (s *SbomManager) processContainerWithMetadata(notif containercollection.Pub
 		helpers.String("sbomName", sbomName))
 }
 
-func (s *SbomManager) waitForSharedContainerData(containerID string) (*objectcache.WatchedContainerData, error) {
-	return backoff.Retry(context.Background(), func() (*objectcache.WatchedContainerData, error) {
+func (s *SbomManager) waitForSharedContainerData(ctx context.Context, containerID string) (*objectcache.WatchedContainerData, error) {
+	return backoff.Retry(ctx, func() (*objectcache.WatchedContainerData, error) {
 		if sharedData := s.k8sObjectCache.GetSharedContainerData(containerID); sharedData != nil {
 			return sharedData, nil
 		}
--- a/pkg/sbommanager/v1/sbom_manager.go
+++ b/pkg/sbommanager/v1/sbom_manager.go
@@ -66,8 +66,13 @@ const (
 	maxPendingScans               = 1000
 	maxFailureRetryEntries        = 1000
 	failureRetryTTL               = 30 * time.Minute
-	// maxWaitForSharedContainerData bounds the wait for a container's shared data; mirrors containerprofilemanager's MaxWaitForSharedContainerData.
-	maxWaitForSharedContainerData = 10 * time.Minute
+	// maxWaitForSharedContainerData bounds the per-container wait for shared data. It must be
+	// >= the producer's budget, else the consumer gives up while the producer is still
+	// populating the data and the container silently never gets an SBOM: the producer
+	// (ContainerWatcher.setSharedWatchedContainerData) retries via cenkalti/backoff
+	// NewExponentialBackOff, whose DefaultMaxElapsedTime is 15m. Waits are cancelled early on
+	// container-remove, so this cap is only reached for a container still alive at 15m.
+	maxWaitForSharedContainerData = 15 * time.Minute
 	// crashLoopRetryTTL is deliberately much longer than failureRetryTTL: a sidecar OOM crash
 	// stalls the shared scanner for every image on the node, so a chronically-crashing image
 	// with restarts spaced further apart than failureRetryTTL must still eventually be pinned,
@@ -114,6 +119,11 @@ type SbomManager struct {
 	pendingMu        sync.Mutex
 	failureReporter  sbommanager.SbomFailureReporter
 	metrics          metricsmanager.MetricsManager
+	// waitCancels holds the cancel func for each in-flight shared-data wait, keyed by container
+	// ID, so a container-remove event cancels the wait immediately instead of leaking a goroutine
+	// (retaining notif/mounts/imageStatus) until the timeout (#850).
+	waitCancels   map[string]context.CancelFunc
+	waitCancelsMu sync.Mutex
 }
 
 var _ sbommanager.SbomManagerClient = (*SbomManager)(nil)
@@ -166,6 +176,7 @@ func CreateSbomManager(ctx context.Context, cfg config.Config, socketPath string
 		pendingScans:       make(map[string]pendingScan),
 		failureReporter:    failureReporter,
 		metrics:            metrics,
+		waitCancels:        make(map[string]context.CancelFunc),
 	}
 	if scannerClient != nil {
 		sm.startScannerReadinessWatcher()
@@ -220,8 +231,15 @@ func (s *SbomManager) getMountedVolumes(pid string) ([]string, error) {
 }
 
 func (s *SbomManager) ContainerCallback(notif containercollection.PubSubEvent) {
-	// only consider container start events
-	if notif.Type != containercollection.EventTypeAddContainer {
+	switch notif.Type {
+	case containercollection.EventTypeAddContainer:
+		// handled below
+	case containercollection.EventTypeRemoveContainer:
+		// Cancel any in-flight shared-data wait: the data is deleted on remove, so the wait can
+		// no longer succeed and would otherwise park a goroutine until the timeout (#850).
+		s.cancelWait(notif.Container.Runtime.ContainerID)
+		return
+	default:
 		return
 	}
 	if utils.IsHostContainer(notif.Container) {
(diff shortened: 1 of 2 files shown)


matthyx:
@Aaradhya-07 can you solve the conflicts?

Aaradhya-07:
@matthyx  sorry I didnt see that,  resolved .

pkg/sbommanager/v1/sbom_manager.go
```diff
@@ -66,6 +66,8 @@ const (
 	maxPendingScans               = 1000
 	maxFailureRetryEntries        = 1000
 	failureRetryTTL               = 30 * time.Minute
+	// maxWaitForSharedContainerData bounds the wait for a container's shared data; mirrors containerprofilemanager's MaxWaitForSharedContainerData.
+	maxWaitForSharedContainerData = 10 * time.Minute
```
matthyx: **Blocking — this shortens the effective wait from 15m to 10m and can silently drop SBOMs.**

The comment justifies 10m by mirroring `containerprofilemanager`, but the relevant counterpart is the *producer*, not another consumer. Shared container data is written by `ContainerWatcher.setSharedWatchedContainerData` (`pkg/containerwatcher/v2/containercallback.go:84`), which retries with:

```go
err := backoff.Retry(func() error { ... }, backoff.NewExponentialBackOff())
```

using `github.com/cenkalti/backoff` v2.2.1, whose `NewExponentialBackOff()` sets `MaxElapsedTime = DefaultMaxElapsedTime = 15 * time.Minute`.

Before this PR the consumer side ran on `context.Background()` and was bounded by backoff **v5**'s own `DefaultMaxElapsedTime`, also 15m — so producer and consumer windows matched exactly. After this PR the consumer gives up at 10m while the producer keeps retrying to 15m.

Concretely: a pod whose `GetWorkload` calls fail for ~11 minutes (API server pressure, etcd slowness — i.e. exactly the degraded cluster where you least want to lose data) will have its shared data populated successfully, but `waitForSharedContainerData` has already returned `context.DeadlineExceeded` and the goroutine returned. `ContainerCallback` only fires on `EventTypeAddContainer` and there is no re-arm path, so that container never gets an SBOM. For a singleton long-lived workload that is permanent until the next restart.

Suggestions, in order of preference:
- Bound the wait to at least the producer's window (15m), or
- Export the producer's budget as a single shared constant so the two can't drift, or
- Keep 10m but re-arm: on the deadline path, re-check the cache once when data eventually shows up (e.g. via the remove/add event) rather than dropping the container.

Whatever the number ends up being, please add a comment tying it to `setSharedWatchedContainerData`'s budget, since that is the constraint that actually matters.

pkg/sbommanager/v1/sbom_manager.go
```diff
@@ -258,20 +260,24 @@ func (s *SbomManager) ContainerCallback(notif containercollection.PubSubEvent) {
...
-func (s *SbomManager) processContainer(notif containercollection.PubSubEvent, mounts []string, imageStatus *runtime.ImageStatusResponse) {
-	sharedData, err := s.waitForSharedContainerData(notif.Container.Runtime.ContainerID)
-	if err != nil {
-		logger.L().Error("SbomManager - container not found in shared data",
-			helpers.String("container ID", notif.Container.Runtime.ContainerID))
-		return
-	}
-	s.processContainerWithMetadata(notif, mounts, imageStatus, sharedData.ImageTag, sharedData.ImageID)
+	// Wait for shared data off the single-worker pool so a short-lived container whose data never arrives can't head-of-line-block SBOM generation node-wide (#850); only real work is submitted to the pool.
+	go func() {
```
matthyx, line 293: **Blocking — unbounded goroutines, none of which can be cancelled, and most of which are provably futile.**

Two facts that together make this worse than it looks:

1. `s.ctx` is `context.Background()`. It comes from `CreateSbomManager(ctx, ...)` at `cmd/main.go:430`, and that `ctx` is `context.Background()` from `cmd/main.go:78` — never wrapped in `WithCancel`, never cancelled on the `SIGTERM` path at `cmd/main.go:524`. So `context.WithTimeout(s.ctx, maxWaitForSharedContainerData)` is not "bounded by shutdown or 10m", it is always a flat 10-minute timer.

2. As the PR description itself notes, shared data is deleted on the container-remove event (`containercallback.go`, `EventTypeRemoveContainer` → `DeleteSharedContainerData`). So for any container that exits before its data is populated, this wait **cannot** succeed — it will always burn the entire 10 minutes and then log an error.

That is precisely the #850 population. A node with a CronJob firing every 10s accumulates ~600 live goroutines, each retaining `notif` (`*containercollection.Container`), `mounts`, and an `ImageStatusResponse` fetched with `Verbose: true` — that response carries the full image config in its `Info` map, routinely tens of KB. Nothing bounds this; `maxPendingScans` doesn't apply on this path.

I don't think "cancel on container-remove" is a follow-up here — it's what keeps the fix from trading one unbounded resource for another, and the signal is already in hand: `ContainerCallback` receives `EventTypeRemoveContainer` and currently early-returns at the top. A `map[containerID]context.CancelFunc` guarded by a mutex, populated here and drained in the remove branch, is roughly 15 lines and turns the worst case from 10 minutes into the container's actual lifetime.

At minimum, please add a concurrency bound (a buffered-channel semaphore around the goroutine) so a burst can't fan out without limit.

Separately, and much smaller: `s.pool.Submit` at the end has no `s.ctx.Done()` check, so work is still enqueued during shutdown. That's moot while `s.ctx` is `Background()`, but it will matter the moment someone makes it cancellable.

pkg/sbommanager/v1/sbom_manager_sharedwait_test.go
```diff
@@ -0,0 +1,47 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
+	defer cancel()
+
+	start := time.Now()
+	data, err := sm.waitForSharedContainerData(ctx, "never-populated")
+	elapsed := time.Since(start)
+
+	assert.Error(t, err)
+	assert.Nil(t, data)
+	assert.Less(t, elapsed, 5*time.Second, "wait must be bounded by the context deadline, not backoff's 15m default")
```
matthyx: **Blocking — the tests don't guard the regression the PR is about, and this one fails badly when it does fail.**

Three separate points:

**a) The assertion doesn't prove *why* the wait returned.** `assert.Error` + `assert.Less(elapsed, 5s)` is also satisfied if the backoff returns `Stop`, or if the operation returns a `*backoff.PermanentError`, or if the mock is misconfigured — none of which are the behavior under test. Add `assert.ErrorIs(t, err, context.DeadlineExceeded)` so the test actually pins "the context deadline is what stopped it."

**b) The 5s bound is a 25x slack on a 200ms deadline, and the failure mode is ugly.** If the regression comes back (`ctx` → `context.Background()`), this test does not fail in 5 seconds — `elapsed` is only measured *after* `waitForSharedContainerData` returns, so the call sits there for backoff v5's full 15m `DefaultMaxElapsedTime`, blows past `go test`'s 10-minute default panic timeout, and takes down the whole package's test binary with a stack dump instead of producing a readable assertion failure. Please run the wait in a goroutine and `select` on `time.After(2*time.Second)` so the regression reports as a clean failure, and tighten the bound while you're there.

**c) The actual fix is untested.** Both tests exercise `waitForSharedContainerData` in isolation, i.e. the context plumbing. Neither asserts the behavioral claim in the PR title — that the wait no longer occupies the single pool worker. Delete the `go func(){...}` wrapper in `ContainerCallback` and call `waitForSharedContainerData(ctx, ...)` inline: both tests still pass, and the node-wide stall is fully restored.

`ContainerCallback` is awkward to test directly (it needs `/proc` and a live CRI socket), so the cheap seam is to extract the goroutine body into something like `func (s *SbomManager) awaitAndSubmit(notif, mounts, imageStatus)` and assert that (i) it returns to the caller without blocking and (ii) with container A's data never arriving and container B's already present, B's `processContainerWithMetadata` runs promptly. That's the regression guard #850 needs.

pkg/sbommanager/v1/sbom_manager.go
```diff
@@ -258,20 +260,24 @@ func (s *SbomManager) ContainerCallback(notif containercollection.PubSubEvent) {
...
+		ctx, cancel := context.WithTimeout(s.ctx, maxWaitForSharedContainerData)
+		defer cancel()
+		sharedData, err := s.waitForSharedContainerData(ctx, notif.Container.Runtime.ContainerID)
+		if err != nil {
+			logger.L().Ctx(s.ctx).Error("SbomManager - container not found in shared data",
```
matthyx, line 309: Nit, non-blocking. The PR description says "the benign `container not found in shared data` log (#848) is unchanged" — it did change here: it gained `.Ctx(s.ctx)` and three extra fields. The added fields are a clear improvement and the `.Ctx()` is consistent with the rest of the file (and currently a no-op, since `s.ctx` carries no span).

The severity is the part worth reconsidering. Post-PR this fires once for every short-lived container that exits before its data lands — a benign, expected outcome that #848 already tracks as noise — and it fires at `Error`. Consider demoting to `Warning`, or to `Debug` specifically when `errors.Is(err, context.DeadlineExceeded)`, keeping `Error` for genuinely unexpected failures. Either way the description should be corrected.

matthyx (commented):
Reviewed at `8511422` (post-merge-with-main). Build, `go vet` and `go test ./pkg/sbommanager/...` all pass locally, and the direction is right: moving the wait off the pool-of-1 does remove the node-wide head-of-line block, and keeping `workerpool.New(1)` is the correct call.

Three things I'd want addressed before merge, plus two notes. Details are in the line comments; summary:

**Blocking**

1. **The wait window is now shorter than the producer's.** `maxWaitForSharedContainerData = 10m` is compared against `containerprofilemanager`, but the thing that actually *populates* this data is `ContainerWatcher.setSharedWatchedContainerData`, which retries on `github.com/cenkalti/backoff` v2 with `DefaultMaxElapsedTime = 15m`. Before this PR the SBOM side also ran to 15m (backoff v5's own `DefaultMaxElapsedTime`), so the two matched. Now there is a 10–15m window where the producer still succeeds but the consumer has already given up, and `ContainerCallback` fires once per container add with no retry path — so that container silently never gets an SBOM.

2. **Nothing ever cancels these goroutines.** `s.ctx` traces back to `ctx := context.Background()` in `cmd/main.go:78` — it is never cancelled, so `context.WithTimeout(s.ctx, 10m)` is always a flat 10-minute wall clock. Combined with the fact that shared data is *deleted* on container-remove, every short-lived container — the exact population that motivated #850 — parks a goroutine that is guaranteed to fail, for the full 10 minutes, holding `notif`, `mounts` and a `Verbose: true` `ImageStatusResponse`.

3. **Neither new test covers the actual fix.** Both exercise `waitForSharedContainerData` in isolation. Revert the `go func(){...}` wrapper (keep the ctx parameter) and both still pass, while the node-wide stall is back.

**Non-blocking**

4. The PR description says the benign "container not found in shared data" log from #848 is unchanged — it isn't; it gained `.Ctx()` and four fields, and it now fires for *every* short-lived container at `Error` level. Worth demoting when the cause is `context.DeadlineExceeded`, and worth correcting the description.

5. Pre-existing, but it interacts with the premise here: `containerCallbackAsync` runs `go cw.setSharedWatchedContainerData(...)` on add and calls `DeleteSharedContainerData` synchronously on remove, with no ordering between them. A late `Set` can land after the `Delete`, which both leaks a permanent cache entry and lets this wait *succeed* for a container that is already gone — the subsequent scan then runs against stale `/proc/<pid>` lowerdirs and burns a `failureRetries` slot. Not for this PR, but it means "data is deleted on remove, so the wait will time out" isn't reliably true.

Aaradhya-07:
Thanks for the thorough review . Pushed a commit addressing them:

1. Wait window shorter than the producer (10m vs 15m). You're right that the relevant bound is the producer, not the other consumer. Bumped maxWaitForSharedContainerData to 15m to match setSharedWatchedContainerData's NewExponentialBackOff budget, and replaced the misleading "mirrors containerprofilemanager" comment with one that ties the value to the producer's 15m DefaultMaxElapsedTime — since that's the constraint that actually matters. (Went with "bound to the producer's window" rather than exporting a shared constant, since the producer only uses backoff's implicit default and there's no clean constant to export without touching containerwatcher — happy to switch to the shared-constant approach if you'd prefer it.)

2. Unbounded, uncancellable, futile goroutines. Implemented cancel-on-remove as you suggested: a map[containerID]context.CancelFunc guarded by a mutex, populated when the wait starts and drained/cancelled in the EventTypeRemoveContainer branch of ContainerCallback. So a short-lived container's wait is cancelled at its actual lifetime instead of parking a goroutine (retaining notif/mounts/imageStatus) for the full timeout. Also added an s.ctx.Err() guard before pool.Submit so no work is enqueued during shutdown — moot while s.ctx is Background(), but correct for when it becomes cancellable. I left cmd/main.go's context wiring and the pre-existing Set-after-Delete ordering race alone, since you flagged both as out of scope here.

3. Tests didn't guard the actual regression. Reworked them.
- Test_waitForSharedContainerData_HonorsContextDeadline now asserts ErrorIs(err, context.DeadlineExceeded) so it pins why the wait
stopped, and runs the wait in a goroutine with a 2 now fails cleanly instead of blocking 15m andpanicking the package's test binary.
- Extracted the goroutine body into awaitAndSubmitt_StuckContainerDoesNotBlockOthers: container A'sdata never arrives, B's is already present, and it asserts awaitAndSubmit returns without blocking and B's work reaches the pool
worker while A is stuck. Deleting the go wrapper f
- Added Test_cancelWait_StopsInFlightWait covering the remove-event cancellation.

Nit (log severity). Corrected — the log is now Debug when the cause is context.DeadlineExceeded/Canceled (the benign #848 outcome for a container that exits before its data lands) inely unexpected failures. You're also right the PR description was wrong that the log was "unchanged"; I've updated the description.

go test -race ./pkg/sbommanager/v1/ passes.

matthyx:
Hi @Aaradhya-07 — thanks for the detailed write-up of the changes! Just a heads-up: I'm not seeing the commit on this PR. The branch head is still `85114224` (the `main` merge from Jul 22), and `git fetch origin pull/859/head` returns that same SHA — `maxWaitForSharedContainerData` is still `10 * time.Minute` here, and none of the new symbols you described (`awaitAndSubmit`, the cancel-on-remove `CancelFunc` map, the reworked tests) are on the branch.

Did the push maybe land on a different branch or your fork? Could you double-check `git push` succeeded and confirm the PR head SHA moves? Happy to re-review as soon as it shows up. 🙏

Aaradhya-07:
Sorry for the runaround , the commit's up now at 464d0e4. The push had been rejected as a non-fast-forward (main got merged into the branch on my side), and when I rebased to fix that, it silently reverted the wait bound back to 10m and dropped a couple of struct fields, so the branch ended up half-reverted. Caught and re-applied it, and reconfirmed 15m + go build/tests before pushing. Should be good to re-review now , thanks for the patience. 🙏

pkg/sbommanager/v1/sbom_manager.go
```diff
@@ -258,20 +278,72 @@ func (s *SbomManager) ContainerCallback(notif containercollection.PubSubEvent) {
...
+				helpers.String("namespace", notif.Container.K8s.Namespace),
+				helpers.String("pod", notif.Container.K8s.PodName),
+				helpers.String("container", notif.Container.K8s.ContainerName),
+				helpers.String("container ID", containerID))
+			return
+		}
+		// Don't enqueue new work once the manager is shutting down.
+		if s.ctx.Err() != nil {
+			return
+		}
```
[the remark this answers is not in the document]
Aaradhya-07: the guard is currently a no-op. Making s.ctx cancellable means wiring cmd/main.go's root context (passed agent-wide) to cancel on SIGTERM, which changes shutdown semantics for every component , flagged  as out of scope for this PR.

matthyx (commented):
Re-reviewed at `464d0e4`. This lands the push cleanly and all three blockers are resolved. `go build`, `go vet`, and `go test -race ./pkg/sbommanager/v1/` all pass locally. Thanks for the thorough follow-up.

**1. Wait window (15m).** `maxWaitForSharedContainerData` is now `15m` and the comment ties it to the producer's `NewExponentialBackOff` `DefaultMaxElapsedTime`, with the "cancelled early on remove, so only reached for a container still alive at 15m" rationale. Correct — producer and consumer windows match again.

**2. Cancel-on-remove.** The `waitCancels` map + mutex, `register`/`unregister`/`cancelWait`, and the `EventTypeRemoveContainer` branch do exactly what was needed: a short-lived container's wait is cancelled at its actual lifetime instead of parking a goroutine for the full timeout. `registerWaitCancel` runs synchronously before the goroutine starts, so a remove can't miss it. The `s.ctx.Err()` guard before `pool.Submit` is a good touch. The demotion to `Debug` on `DeadlineExceeded`/`Canceled` also correctly addresses the #848 log-noise nit.

**3. Tests.** All three now guard the real behavior: `HonorsContextDeadline` asserts `ErrorIs(context.DeadlineExceeded)` and runs in a goroutine with a 2s `select` so a regression fails cleanly instead of hanging 15m; `awaitAndSubmit_StuckContainerDoesNotBlockOthers` is the actual #850 guard (stuck A, present B, B's work must reach the worker); and `cancelWait_StopsInFlightWait` covers the remove path. Nice use of the `signalingSbomClient` to prove work reached the pool worker.

One minor, non-blocking observation (not for this PR): because add and remove callbacks are each dispatched on the outer `containerCallback` worker pool, there's no guaranteed ordering between them, so a remove processed before its add registers the cancel would leave the wait to run to the full 15m for that container. It's rare, bounded, and no worse than before — just noting it as a known edge of the pubsub/pool model rather than something to fix here.

LGTM. 👍

matthyx (approved):

PR #866, opened 2026-07-27
fix(containerprofile): split oversized chunks on HTTP 413 instead of ending learning

### LINKED ISSUE #865: node-agent retries oversized ContainerProfiles forever: HTTP 413 from storage is not classified as too-large
## Summary

node-agent does not recognise an HTTP 413 from storage, so an oversized ContainerProfile chunk is retried every 5s forever and the container is never flagged too-large.

Reported by a customer (ISO Gruppe) on Rancher/RKE2, on `cattle-system` workloads:

```
QueueManager - request entity too large  contentLength=3231797  maxObjectSize=2500000
```

## Root cause

`storage`'s QueueManager rejects a create whose `Content-Length` exceeds `kindQueues.containerprofiles.maxObjectSize` before it reaches the registry ([manager.go:144-148](https://github.com/kubescape/storage/blob/main/pkg/queuemanager/manager.go#L144-L148)), replying `413` with a **plain-text** body — so the response is not a `Status` object and carries no sentinel.

node-agent classifies failures by exact string equality ([containerprofile_queue.go:256](https://github.com/kubescape/node-agent/blob/main/pkg/containerprofilemanager/v1/queue/containerprofile_queue.go#L256)):

```go
if err.Error() == file.ObjectTooLargeError.Error() || err.Error() == file.ObjectCompletedError.Error() { ... continue }
...
qd.requeueImmediate(queuedProfile)
```

A 413 matches neither, so it falls through to `requeueImmediate`. `apierrors.IsRequestEntityTooLargeError` is never called anywhere in this repo.

## Why it compounds

- Retried every `DefaultRetryInterval` (5s) forever — no attempt counter, no dead-letter.
- The queue is a disk-persistent `dque`, so the poison item survives pod restarts.
- `enforceMaxSize()` evicts from the **head** at `MaxQueueSize=1000` while `requeueImmediate` appends to the **tail**, so the poison chunk can outlive newer chunks that storage would have accepted.
- The container is never flagged too-large, because that only happens via `OnQueueError`.

Net effect: a silent hole in the profile, a permanent stream of warnings in storage logs, and no UI signal.

## Secondary defect: exact equality is too narrow for the sentinels too

Even for the real sentinels, exact equality is fragile. They cross the API boundary as a `StatusError` whose `Error()` returns only `Message`, and storage prefixes the sentinel with the limit on some paths:

```go
fmt.Errorf("application profile size exceeds the limit of %d: %w", a.maxApplicationProfileSize, ObjectTooLargeError)
```

That message never equals `ObjectTooLargeError.Error()`, so those rejections also fall into the infinite requeue loop.

## Fix

Classify with a tested helper instead:

- `apierrors.IsRequestEntityTooLargeError` for 413 — it matches on the status code alone, which is what makes it work against a plain-text body — reported as `ObjectTooLargeError` so `handleSaveProfileError` sets `status=too-large` and ends learning with reason `too_large`.
- `errors.Is` plus a substring match on the message for the sentinels, covering both the wrapped and the relayed shapes.
- Bound retries with `MaxAttempts` (default 10). 413 is only one way to get a permanently failing item; without a bound, any such item occupies the queue indefinitely.

PR: #866

## Related

- kubescape/storage — size caps are inconsistent across the three layers, and `Create` swallows `ObjectTooLargeError`
- kubescape/helm-charts — `maxTsProfileSize` and `maxApplicationProfileSize` are not chart-exposed

rotemamsa:
Fixes #865

## Problem

customer of Rancher/RKE2 saw storage log this on repeat for `cattle-system` workloads:

```
QueueManager - request entity too large  contentLength=3231797  maxObjectSize=2500000
```

storage's `QueueManager` rejects a create whose `Content-Length` exceeds `kindQueues.containerprofiles.maxObjectSize` before it reaches the registry, replying `413` with a **plain-text** body — so there is no `Status` object and no sentinel.

## What changed since the last revision of this PR

@matthyx's review (https://github.com/kubescape/node-agent/pull/866#pullrequestreview-4794269159) verified the original fix's diagnosis and bounded-retry mechanism, but flagged one blocking concern: mapping a bare 413 onto `ObjectTooLargeError` conflates two different signals. `ObjectTooLargeError` is storage's own **aggregate** findings-count check — an authoritative "this container's profile is genuinely full" verdict that's fair to end learning on. A bare 413 is a **transport** rejection of one delta based on byte size alone, and treating it the same way throws away all future learning for a container that merely produced one oversized chunk.

This revision implements the reviewer's suggested fix: **on a bare 413, split the rejected chunk in half and resend both halves**, recursing until each piece is accepted, instead of ending learning.

## Change

**`classifyFailure()`** gains a third reaction, `failureSplit`, alongside the existing `failureRetryable`/`failureTerminal`. The two genuine storage sentinels (`ObjectTooLargeError`, `ObjectCompletedError`, matched via `errors.Is` + substring) are checked **first** and remain terminal exactly as before — only a bare 413 detected purely by HTTP status code (`apierrors.IsRequestEntityTooLargeError`) now triggers `failureSplit` instead of being treated as terminal.

**Splitting (`containerprofile_split.go`, new).** `splitProfile` partitions the rejected chunk's list/map fields (`Capabilities`, `Execs`, `Opens`, `Syscalls`, `Endpoints`, `IdentifiedCallStacks`, `Ingress`, `Egress`, `PolicyByRuleId`) roughly in half, with a non-empty guarantee so recursion always makes progress, and gives each half a fresh one-time slug over the same base name (recovered via storage's own `file.SplitProfileName`, so agreement with the server's aggregation key is true by construction — a `-partN` suffix scheme was considered and rejected, since `SplitProfileName` cuts on the last hyphen and would scatter a differently-suffixed half into a phantom aggregate). A byte-progress guard refuses to keep splitting a chunk that isn't shrinking.

**Preserving the report chain — the trickiest part.** Storage reconstructs a container's lifecycle by walking a strictly linear `previousReportTimestamp → reportTimestamp` chain (`consolidateContinuousTimeSeries`). Naively giving both halves the parent's identical timestamp pair permanently forks that chain and leaves the profile in `Learning` forever — worse than today's behavior. Instead, `chainHalves` interposes a fresh timestamp `X` so one half spans `(P, X]` and the other `(X, T]`, keeping the chain linear (including for the *next* real report, and correctly under recursive re-splitting). This required care around: a monotonic-clock suffix on production timestamps (`time.Now().String()`) that needs stripping before parsing; a numeric (non-letter) timezone-abbreviation fallback for several IANA zones; and never taking a naive midpoint against a first-report's zero `previousReportTimestamp` (which would otherwise produce an implausible timestamp that storage's expiry check matches unconditionally, re-ending learning through a different route).

**Bounding the recursion.** `MaxSplitDepth` (default 4) plus the byte-progress guard cap how many times a single chunk can be halved. A chunk that can't be split further (a single oversized element, or depth exhausted) is dropped with a `Warning` — and replaced by a **stitch chunk**: a metadata-and-scalars-only stand-in that preserves the dropped chunk's `(previousReportTimestamp, reportTimestamp)` link so the chain doesn't fork, without carrying any of the oversized data. A stitch is never itself re-stitched. Neither drop path calls `ErrorCallback.OnQueueError` — that's the callback that ends learning, and avoiding it on every 413 path is the entire point of this change.

**Observability.** Two new counters (`ReportContainerProfileSplit`, `ReportContainerProfileChunkDropped(reason)`) across all `MetricsManager` implementations, since after this change there's no remaining status signal on the 413 path at all — these counters are the only way to see the underlying estimator issue (#870) firing in the field.

The bounded-retry mechanism from the previous revision (`MaxAttempts`, default 360) is unchanged and still governs genuinely retryable failures.

## Tests

`containerprofile_split_test.go` (new) and `containerprofile_queue_errors_test.go` (extended) — partition correctness (no loss/duplication across halves), shared-field verbatim-copy vs. the two report-timestamp fields (deliberately *not* equal across halves), naming/base-name preservation, the floor case and non-empty rebalance, determinism, the timestamp-chain interposition and its zero-time/monotonic-suffix/numeric-timezone edge cases (including a golden port of storage's own consolidation algorithm so drift against the real thing is caught), recursive re-splitting staying chain-linear, the byte-progress guard, and end-to-end queue tests for split-and-resend, the unsplittable floor case (with stitch-chunk emission), `MaxSplitDepth` bounding, and `SplitDepth`/stitch persistence across a queue restart.

Verified:
```
go build ./...                                                                          ok
go vet ./pkg/containerprofilemanager/... ./pkg/metricsmanager/...                        ok
go test ./pkg/containerprofilemanager/... ./pkg/metricsmanager/... -race -count=3        ok
gofmt -l (all changed files)                                                             clean
```

Also fixed a pre-existing data race in the test-only `MockProfileCreator` (unsynchronized fields read across goroutines), found while getting the package's `-race` gate green.

## Not in this PR

- **#870** — node-agent's own pre-send size estimator (`config.MaxTsProfileSize`) mixes byte-size and element-count units, which is the root cause of reaching a 413 at all. Fixing it should make splitting a rare safety net rather than something regularly exercised.
- **#871** — a `MaxAttempts` exhaustion drop or LRU eviction can already break a container's report chain today, independent of this PR, permanently preventing that container's profile from completing. Found as a side effect of verifying this PR's own chain-preservation logic; appears to have never been reported.
- **kubescape/storage#352** (cross-repo) — `consolidateContinuousTimeSeries` should collapse rows sharing an identical `(previousReportTimestamp, reportTimestamp)` pair. This is the reason node-agent has to manufacture an intermediate timestamp at all; once it lands (and once a minimum storage version can be assumed), the `chainHalves` interposition logic in this PR becomes unnecessary and can be deleted.
- storage's swallowing of `ObjectTooLargeError` into an HTTP 201 with the data discarded server-side (`clearSpec`) is unrelated to the 413 path this PR addresses, but was a lower-priority thing worth someone eventually surfacing as a client-visible signal.
- Previously tracked: kubescape/storage#350 (the size caps themselves are inconsistent) and kubescape/helm-charts#885.

## Summary by CodeRabbit

* **New Features**
  * Container-profile learning now splits profiles rejected for excessive size and retries manageable chunks.
  * Added safeguards for retry limits, split-depth limits, unsplittable chunks, and queue shutdown.
  * Added stitch chunks to preserve continuity when portions are dropped.
  * Added monitoring metrics for profile splits and dropped chunks.

* **Documentation**
  * Added guidance on size rejection handling, retries, timestamp processing, observability, and limitations.

* **Tests**
  * Added comprehensive coverage for splitting, retries, persistence, stitching, queue limits, and error handling.

### DIFF
--- a/pkg/containerprofilemanager/v1/queue/containerprofile_queue.go
+++ b/pkg/containerprofilemanager/v1/queue/containerprofile_queue.go
@@ -8,11 +8,13 @@ import (
 	"path/filepath"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/joncrlsn/dque"
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
+	"github.com/kubescape/node-agent/pkg/metricsmanager"
 	"github.com/kubescape/node-agent/pkg/storage"
 	"github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1"
 )
@@ -49,7 +51,35 @@ type QueuedContainerProfile struct {
 	// retryable error. Items persisted before this field existed decode with Attempts
 	// at zero, so they simply get a full budget of retries.
 	Attempts int `json:"attempts"`
+	// SplitDepth counts how many times this item's lineage has been halved after an HTTP 413.
+	// Items persisted before this field existed decode with SplitDepth at zero.
+	SplitDepth int `json:"splitDepth"`
+	// IsStitch marks a metadata-only chunk emitted by dropChunk to repair the report chain.
+	// A stitch is never split and is never itself stitched - dropping a stitch leaves the chain
+	// forked, which is strictly better than an unbounded stitch->413->stitch loop.
+	IsStitch bool `json:"isStitch"`
 }
+
+// dropReason labels why a chunk was discarded by dropChunk. The set is closed and every value
+// is a compile-time constant, so it is safe as a metric label.
+type dropReason string
+
+const (
+	// dropReasonUnsplittable: splitProfile refused - floor case, unconstructable chain
+	// timestamp, or no byte progress. Collapsed into one reason because all three mean
+	// "this chunk cannot be made smaller".
+	dropReasonUnsplittable dropReason = "unsplittable"
+	// dropReasonDepthExhausted: MaxSplitDepth reached before the chunk became acceptable.
+	dropReasonDepthExhausted dropReason = "depth-exhausted"
+	// dropReasonStitchRejected: a stitch chunk was itself rejected. Not re-stitched, so this
+	// is the one drop that knowingly leaves the report chain forked.
+	dropReasonStitchRejected dropReason = "stitch-rejected"
+	// dropReasonEnqueueFailed: the replacement stitch could not be enqueued.
+	dropReasonEnqueueFailed dropReason = "enqueue-failed"
+)
+
+// ErrQueueNotRunning is returned by enqueueLocked once the queue has been closed.
+var ErrQueueNotRunning = errors.New("queue is not running")
 
 // QueuedContainerProfileBuilder creates a new QueuedContainerProfile instance for dque
 func QueuedContainerProfileBuilder() interface{} {
@@ -69,7 +99,14 @@ type QueueData struct {
 	errorCallback ErrorCallback
 	maxQueueSize  int
 	maxAttempts   int
+	maxSplitDepth int
 	retryInterval time.Duration
+	metrics       metricsmanager.MetricsManager
+
+	// Written by the queue processor goroutine and read by GetQueueStats, so they carry
+	// their own synchronization rather than widening qd.mu.
+	splits        atomic.Int64
+	chunksDropped atomic.Int64
 
 	stopChan chan struct{}
 	wg       sync.WaitGroup
@@ -83,9 +120,11 @@ type QueueConfig struct {
 	QueueDir        string
 	MaxQueueSize    int
 	MaxAttempts     int
+	MaxSplitDepth   int
 	RetryInterval   time.Duration
 	ItemsPerSegment int
 	ErrorCallback   ErrorCallback
+	MetricsManager  metricsmanager.MetricsManager
 }
 
 // NewQueueData creates a new QueueData instance with simple LRU behavior
@@ -102,6 +141,12 @@ func NewQueueData(ctx context.Context, creator storage.ProfileCreator, config Qu
 	}
 	if config.MaxAttempts <= 0 {
 		config.MaxAttempts = DefaultMaxAttempts
+	}
+	if config.MaxSplitDepth <= 0 {
+		config.MaxSplitDepth = DefaultMaxSplitDepth
+	}
+	if config.MetricsManager == nil {
+		config.MetricsManager = &metricsmanager.MetricsNoop{}
 	}
 	if config.RetryInterval == 0 {
 		config.RetryInterval = DefaultRetryInterval
@@ -143,7 +188,9 @@ func NewQueueData(ctx context.Context, creator storage.ProfileCreator, config Qu
 		errorCallback: config.ErrorCallback,
 		maxQueueSize:  config.MaxQueueSize,
 		maxAttempts:   config.MaxAttempts,
+		maxSplitDepth: config.MaxSplitDepth,
 		retryInterval: config.RetryInterval,
+		metrics:       config.MetricsManager,
 		stopChan:      make(chan struct{}),
 		running:       true,
 	}
@@ -174,29 +221,34 @@ func (qd *QueueData) Enqueue(profile *v1beta1.ContainerProfile, containerID stri
 	qd.mu.Lock()
 	defer qd.mu.Unlock()
 
-	if !qd.running {
-		return fmt.Errorf("queue is not running")
-	}
-
-	// Create queued profile
-	queuedProfile := &QueuedContainerProfile{
+	if err := qd.enqueueLocked(&QueuedContainerProfile{
 		Profile:     profile,
 		ContainerID: containerID,
-	}
-
-	// Remove oldest items if we're at capacity
-	qd.enforceMaxSize()
-
-	// Add new item
-	err := qd.queue.Enqueue(queuedProfile)
-	if err != nil {
-		return fmt.Errorf("failed to enqueue profile: %w", err)
+	}); err != nil {
+		return err
 	}
 
 	logger.L().Debug("container profile enqueued",
 		helpers.String("name", profile.Name),
 		helpers.String("namespace", profile.Namespace),
 		helpers.Int("queueSize", qd.queue.Size()))
+
+	return nil
+}
+
+// enqueueLocked adds item to the queue, evicting the oldest items first.
+// Callers must hold qd.mu.
+func (qd *QueueData) enqueueLocked(item *QueuedContainerProfile) error {
+	if !qd.running {
+		return ErrQueueNotRunning
+	}
+
+	// Remove oldest items if we're at capacity
+	qd.enforceMaxSize()
+
+	if err := qd.queue.Enqueue(item); err != nil {
+		return fmt.Errorf("failed to enqueue profile: %w", err)
+	}
 
 	return nil
 }
@@ -250,6 +302,7 @@ func (qd *QueueData) processAllItems() {
 	logger.L().Debug("processing queue", helpers.Int("size", queueSize))
 
 	// Process each item in the queue
+processLoop:
 	for i := 0; i < queueSize; i++ {
 		// Try to get an item from the queue
 		iface, err := qd.queue.Dequeue()
@@ -273,52 +326,80 @@ func (qd *QueueData) processAllItems() {
 		// Attempt to create the profile
 		err = qd.creator.CreateContainerProfileDirect(queuedProfile.Profile)
 		if err != nil {
+			// An explicit switch, so that a future fourth failureKind cannot silently fall
+			// through to the retryable branch.
 			kind, reported := classifyFailure(err)
-			if kind == failureTerminal {
+			switch kind {
+			case failureTerminal:
 				logger.L().Debug("got rejected to create container profile, skipping",
 					helpers.String("name", queuedProfile.Profile.Name),
 					helpers.Error(err))
 
 				// Call error callback if provided to propagate the error. The sentinel is
 				// reported rather than err itself, so that the manager can map it onto the
-				// right terminal container status even when the transport lost the sentinel
-				// (an HTTP 413 from storage's QueueManager carries no sentinel at all).
+				// right terminal container status even when the transport lost the sentinel.
 				if qd.errorCallback != nil {
 					qd.errorCallback.OnQueueError(queuedProfile.Profile, queuedProfile.ContainerID, reported)
 				}
-				continue
-			}
 
-			// Retryable failure. Bound the number of retries so that an item which never
-			// succeeds cannot occupy the queue indefinitely: requeuing appends to the tail
-			// while enforceMaxSize evicts from the head, so an unbounded retry would push
-			// out newer profiles that could have been saved.
-			queuedProfile.Attempts++
-			if queuedProfile.Attempts >= qd.maxAttempts {
-				logger.L().Warning("dropping container profile after too many failed attempts",
+			case failureSplit:
+				// A bare 413 rejects this one delta for its size; it says nothing about the
+				// container, so learning continues and only the chunk is shed.
+				switch {
+				case queuedProfile.IsStitch:
+					qd.dropChunk(queuedProfile, dropReasonStitchRejected)
+				case queuedProfile.SplitDepth >= qd.maxSplitDepth:
+					qd.dropChunk(queuedProfile, dropReasonDepthExhausted)
+				default:
+					a, b, ok := splitProfile(queuedProfile.Profile)
+					if !ok {
+						qd.dropChunk(queuedProfile, dropReasonUnsplittable)
+						continue
+					}
+
+					logger.L().Debug("splitting container profile rejected with 413",
+						helpers.String("name", queuedProfile.Profile.Name),
+						helpers.String("namespace", queuedProfile.Profile.Namespace),
+						helpers.Int("splitDepth", queuedProfile.SplitDepth),
+						helpers.Int("elements", countPartitionableElements(&queuedProfile.Profile.Spec)),
+						helpers.Int("bytes", serializedSize(queuedProfile.Profile)))
+
+					qd.splits.Add(1)
+					qd.metrics.ReportContainerProfileSplit()
+					qd.requeueSplit(queuedProfile, a, b)
+				}
+
+			case failureRetryable:
+				// Bound the number of retries so that an item which never succeeds cannot
+				// occupy the queue indefinitely: requeuing appends to the tail while
+				// enforceMaxSize evicts from the head, so an unbounded retry would push out
+				// newer profiles that could have been saved.
+				queuedProfile.Attempts++
+				if queuedProfile.Attempts >= qd.maxAttempts {
+					logger.L().Warning("dropping container profile after too many failed attempts",
+						helpers.String("name", queuedProfile.Profile.Name),
+						helpers.String("namespace", queuedProfile.Profile.Namespace),
+						helpers.String("containerID", queuedProfile.ContainerID),
+						helpers.Int("attempts", queuedProfile.Attempts),
+						helpers.Error(err))
+					continue
+				}
+
+				logger.L().Debug("failed to create container profile, requeuing",
 					helpers.String("name", queuedProfile.Profile.Name),
-					helpers.String("namespace", queuedProfile.Profile.Namespace),
-					helpers.String("containerID", queuedProfile.ContainerID),
 					helpers.Int("attempts", queuedProfile.Attempts),
 					helpers.Error(err))
-				continue
-			}
 
-			logger.L().Debug("failed to create container profile, requeuing",
-				helpers.String("name", queuedProfile.Profile.Name),
-				helpers.Int("attempts", queuedProfile.Attempts),
-				helpers.Error(err))
+				qd.requeueImmediate(queuedProfile)
 
-			qd.requeueImmediate(queuedProfile)
+				// Check for "too many requests" error
+				if strings.Contains(err.Error(), "the server has received too many requests and has asked us to try again later") {
+					logger.L().Debug("server overloaded, breaking processing loop and waiting for next interval",
+						helpers.String("name", queuedProfile.Profile.Name),
+						helpers.Error(err))
 
-			// Check for "too many requests" error
-			if strings.Contains(err.Error(), "the server has received too many requests and has asked us to try again later") {
-				logger.L().Debug("server overloaded, breaking processing loop and waiting for next interval",
-					helpers.String("name", queuedProfile.Profile.Name),
-					helpers.Error(err))
-
-				// Break from the for loop
-				break
+					break processLoop
+				}
 			}
 
 		} else {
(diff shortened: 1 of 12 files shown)


pkg/containerprofilemanager/v1/queue/containerprofile_queue_errors.go
```diff
@@ -0,0 +1,62 @@
...
+//     which is why it must be matched on the status code.
+func classifyFailure(err error) (failureKind, error) {
+	if err == nil {
+		return failureRetryable, nil
+	}
+
+	// QueueManager rejects oversized requests up front with 413 and a plain-text body, so
+	// there is no Status object to carry a reason. IsRequestEntityTooLargeError matches on
+	// the status code alone, which is what makes this work.
+	if apierrors.IsRequestEntityTooLargeError(err) {
```
matthyx: This is the part I'd like changed. Detecting the 413 here is right — `IsRequestEntityTooLargeError` matching on the status code alone is exactly what makes it work against the plain-text body. Reporting it as `file.ObjectTooLargeError` is what I'd push back on.

The two conditions aren't the same. `ObjectTooLargeError` is a findings-count verdict on the **aggregated** profile — and note that for a TS chunk, `PreSave` returns early and never size-checks the chunk at all, so the sentinel only ever reaches us when the *aggregate* is already flagged. The 413 is a `Content-Length` check on a single delta.

Suggest a third `failureKind` here — something like `failureOversized` — so the queue can split the chunk and requeue both halves instead of ending learning for the container. The sentinel path stays exactly as you have it.

pkg/containerprofilemanager/v1/queue/containerprofile_queue.go
```diff
@@ -253,20 +273,40 @@ func (qd *QueueData) processAllItems() {
...
 					helpers.Error(err))
 
-				// Call error callback if provided to propagate the error
+				// Call error callback if provided to propagate the error. The sentinel is
+				// reported rather than err itself, so that the manager can map it onto the
+				// right terminal container status even when the transport lost the sentinel
+				// (an HTTP 413 from storage's QueueManager carries no sentinel at all).
 				if qd.errorCallback != nil {
-					qd.errorCallback.OnQueueError(queuedProfile.Profile, queuedProfile.ContainerID, err)
+					qd.errorCallback.OnQueueError(queuedProfile.Profile, queuedProfile.ContainerID, reported)
```
matthyx, line 342: This is where the consequence lands: `reported` is the sentinel, so `handleSaveProfileError` sets `status=too-large`, calls `deleteContainer` and ends learning with reason `too_large` — permanently, for the whole container, on one oversized delta.

Since `saveContainerProfile` calls `emptyEvents()` after each enqueue, every queued profile is a delta, and the first report after container start carries everything observed so far — so this fires on the peak, not on a steady state. On `main` the container kept learning and reached `Completed` at max sniff time, losing only that one delta.

For the oversized case I'd like: split the profile in two, requeue both, no `OnQueueError`, learning continues. The `ObjectCompletedError` / aggregate-`ObjectTooLargeError` sentinels should keep going through this branch unchanged — those genuinely mean "stop sending".

pkg/containerprofilemanager/v1/queue/containerprofile_queue.go
```diff
@@ -253,20 +273,40 @@ func (qd *QueueData) processAllItems() {
...
 				}
 				continue
 			}
 
+			// Retryable failure. Bound the number of retries so that an item which never
+			// succeeds cannot occupy the queue indefinitely: requeuing appends to the tail
+			// while enforceMaxSize evicts from the head, so an unbounded retry would push
+			// out newer profiles that could have been saved.
+			queuedProfile.Attempts++
```
matthyx: Once the split lands, a chunk being usefully halved shouldn't consume this budget — otherwise productive splits burn retries meant for storage being unreachable. Worth carrying a separate split-depth counter on `QueuedContainerProfile` alongside `Attempts`, with the drop happening when a chunk can't be divided any further rather than when the retry budget runs out.

The bound itself is good and I'd keep it — including the 360 default and `TestDefaultRetryBudgetOutlastsARestart`. One thing worth adding: this path drops silently, with no callback and no metric. With a full 1000-item queue and a >30 min outage, every item exhausts its budget in roughly the same tick, so this can emit up to 1000 `Warning` lines at once. A counter and rate-limited logging would make the loss visible.

pkg/containerprofilemanager/v1/queue/containerprofile_queue_errors.go
```diff
@@ -0,0 +1,62 @@
...
+		return failureTerminal, file.ObjectCompletedError
+	}
+
+	return failureRetryable, err
+}
+
+// matchesSentinel reports whether err carries the given storage sentinel, either as a
+// wrapped error or as text in the message relayed by the apiserver.
+func matchesSentinel(err, sentinel error) bool {
+	return errors.Is(err, sentinel) || strings.Contains(err.Error(), sentinel.Error())
```
matthyx: Minor: the substring arm is unanchored, so any error whose message happens to embed `"object is too large"` or `"object is completed"` becomes terminal and ends learning for the container. Low probability, but the blast radius is large for a text match.

The `errors.Is` arm and the prefixed-message case are both well justified — storage really does wrap as `"...size exceeds the limit of %d: %w"`. Could we narrow the substring arm to `*apierrors.StatusError` (where we know the message is a relayed sentinel and `errors.Is` can't work) rather than applying it to every error shape?

pkg/containerprofilemanager/v1/queue/containerprofile_queue_errors_test.go
```diff
@@ -0,0 +1,282 @@
...
+func testProfile() *v1beta1.ContainerProfile {
+	return &v1beta1.ContainerProfile{
+		ObjectMeta: metav1.ObjectMeta{Name: "cattle-cluster-agent", Namespace: "cattle-system"},
+	}
+}
+
+// TestQueueEndsLearningOnHTTP413 is the regression test for the ISO Gruppe report: an
+// oversized profile rejected by storage's QueueManager must end learning instead of being
+// retried every RetryInterval forever.
+func TestQueueEndsLearningOnHTTP413(t *testing.T) {
```
matthyx: This test pins the behaviour I'm asking to change, so it'll need reshaping: a 413 should result in the profile being split and **both halves** delivered, with `OnQueueError` never called and learning still running.

The rest of the file is genuinely good — building the errors the way client-go actually does (`NewGenericServerResponse` for the plain-text 413, a relayed `StatusError` for the sentinels) is the right level of fidelity, and `TestClassifyFailure_413IsNotRequeued` is a nice guard against the old equality check coming back. A couple worth adding alongside the split tests:

- the two halves carry distinct `-<uuid-hex>` name suffixes that `SplitProfileName` maps back to the same base name (a `...-part1` style suffix would silently re-base the aggregate key);
- `Attempts` survives a queue `Close`/reopen — the whole retry bound depends on that counter being persisted, and nothing exercises it today.

matthyx (dismissed):
Thanks for the thorough writeup — the diagnosis in #865 is right, and I verified every load-bearing claim against `kubescape/storage@v0.0.290`: QueueManager really does reject on `Content-Length` with a plain-text 413, `CreateContainerProfileDirect` passes the clientset error through unwrapped, and the exact-string classification on `main` falls through to `requeueImmediate` forever on a disk-persistent queue. The bounded retry and the `errors.Is` + relayed/prefixed `StatusError` matching are real, independent fixes and I'd like to keep both.

My one blocking concern is the `413 -> ObjectTooLargeError -> end learning` hop. I don't think we should stop learning just because a **chunk** is too big — we're sending TS container profiles here, and the two conditions are genuinely different:

- **`ObjectTooLargeError`** is a findings-count check on the **aggregated** profile. In `ContainerProfileProcessor.PreSave`, the TS branch (`ReportSeriesIdMetadataKey != ""`) returns early and never size-checks the chunk at all — it only rejects when the *existing aggregate* is already flagged `too-large`/`completed`. The size rule lives in the non-TS branch and counts findings, not bytes:

  ```go
  size += len(profile.Spec.Execs) + len(profile.Spec.Opens) + len(profile.Spec.Syscalls) +
          len(profile.Spec.Capabilities) + len(profile.Spec.Endpoints) +
          len(profile.Spec.IdentifiedCallStacks) + len(profile.Spec.Ingress) + len(profile.Spec.Egress)
  if size > a.MaxContainerProfileSize {
      // set annotation but don't return an error as we want to save the profile anyway
      profile.Annotations[helpers.StatusMetadataKey] = helpers.TooLarge
  }
  ```

  It doesn't even return an error — it annotates and saves. So `ObjectTooLargeError` on a chunk means "the aggregate is full, stop sending", which *is* a fair reason to end learning.

- **The 413** is a defensive `Content-Length` check on one delta. Mapping it onto the sentinel manufactures the "container is full" verdict out of a transport limit.

### What I'd like instead: split the chunk and send both halves

Since `saveContainerProfile` calls `emptyEvents()` after each enqueue, every queued profile is a delta. Cutting one delta in two is exactly what the existing `ProfileRequiresSplit` path already does (`container_operations.go:38-50` — it's literally "flush early and start a new chunk"), just reactive instead of predictive. I checked the three things that could have broken it:

1. **Naming works, with one trap.** `GetOneTimeSlug` returns `<base-slug>-<uuid-hex>`, and storage's `SplitProfileName` cuts on the **last** hyphen to recover the base name. Each half needs its own fresh `-<uuid-hex>` on the base name. Do *not* suffix the existing chunk name (`...-<uuid>-part1`) — `SplitProfileName` would then return `base-<uuid>` and scatter the halves under a different aggregate key.
2. **The halves can't collide.** `time_series` has `PRIMARY KEY (kind, namespace, name, seriesID, tsSuffix)` and `AfterCreate` writes `tsSuffix` from the per-chunk UUID, so two halves land as distinct rows even with identical `reportTimestamp` / `previousReportTimestamp` / `seriesID`. Nothing needs re-stamping.
3. **The aggregate is unchanged.** Chunks merge by `patchMergeKey` and the count is `len()` after dedupe, so how the delta was cut doesn't affect the findings total.

**What splits vs what duplicates:** partition the lists (`Capabilities`, `Execs`, `Opens`, `Syscalls`, `Endpoints`, `IdentifiedCallStacks`, `Ingress`, `Egress`) and the `PolicyByRuleId` map; copy `Architectures`, `ImageID`, `ImageTag`, `SeccompProfile`, `LabelSelector` and the annotations into both halves.

Halving converges in ceil(log2(bytes/cap)) rounds — 3.23 MB against 2.5 MB is a single round — and costs nothing server-side, since QueueManager rejects before dispatch. Floor case: a chunk holding one element across all lists that still 413s is genuinely unsendable, so drop *that chunk* with a Warning; still not the container.

### Two smaller things

- **The "container is finally flagged too-large" outcome doesn't actually land on the 413 path.** `handleSaveProfileError` sets the status on the in-memory `watchedContainer`, then calls `deleteContainer`, which explicitly skips the termination save when status is `TooLarge` (`lifecycle.go:287-290`). We only write `StatusMetadataKey` inside `saveContainerProfile` (`monitoring.go:181`), so no chunk ever carries `status=too-large` to storage. For the registry sentinel path that's fine — storage annotates the object itself — but for a 413 the request never reaches the registry, so the series is just abandoned and the expiry sweep later marks it `Completed/Partial`. Worth correcting in the description either way.
- **The estimator is why we get a 413 at all.** `maxTsProfileSize` defaults to 2 MiB — already *under* storage's 2.5 MB — yet the observed body was 3.23 MB, because `size` mixes byte counts (`ReportCapability` -> `size.Of(...)`) with element counts (`ReportSyscall` -> `syscalls.Append(...)`, the number of elements added) and never accounts for egress/ingress neighbors being expanded at serialization time after DNS resolution. That's a separate node-agent-side defect from the cap inconsistency tracked in kubescape/storage#350 — worth its own issue. Fix it and the split stays a safety net rather than the normal path.

Verification on my side: `go build ./...` ok, `go test ./pkg/containerprofilemanager/...` all pass, `gofmt`/`vet` clean. The `-race` failure in `TestQueueWithDifferentConfigurations/small-queue` does reproduce on `main`, so that note is accurate. I also confirmed the gob claim independently — `dque` uses gob (`segment.go:138,221`), old->new decodes with `Attempts=0` and new->old ignores the extra field, so the persisted-item change is safe both ways.

pkg/containerprofilemanager/v1/queue/containerprofile_queue.go
```diff
@@ -294,17 +415,101 @@ func (qd *QueueData) requeueImmediate(queuedProfile *QueuedContainerProfile) {
...
+			SplitDepth:  parent.SplitDepth + 1,
+			IsStitch:    false,
 		}
 	}
+
+	qd.mu.Lock()
+	defer qd.mu.Unlock()
+
+	if err := qd.enqueueLocked(half(a)); err != nil {
+		logger.L().Debug("failed to enqueue the first half of a split container profile",
```
matthyx: The two enqueue-failure paths here are asymmetric, and the first one is the worse of the two.

If the **first** half fails, this logs at `Debug` and returns — so `b` is never enqueued either. The parent was already dequeued, so *both* halves are gone. The second-half failure loses one half and correctly logs a `Warning` naming the forked chain; the first-half failure loses the entire chunk and is quieter about it. Neither path increments `chunksDropped` or calls `ReportContainerProfileChunkDropped`, so the new metric can't see either.

`enqueueLocked` returns an error on `!qd.running` — reachable during `Close` while `processAllItems` is still mid-flight — or on a dque disk error. Rare, but the consequence is precisely what this PR exists to prevent: a forked chain, which by the analysis in #871 leaves the profile in `Learning` permanently.

Suggest treating it like the second-half case: `Warning`, a drop-metric bump, and ideally a stitch carrying the parent's original `(previousReportTimestamp, reportTimestamp)` pair — at that point nothing from the parent reached the queue, so the parent's own interval is exactly what needs repairing. That also makes the two branches read the same way, which matters for a path this hard to exercise.

matthyx (commented):
Re-reviewed at `49f5f19`. This addresses everything from my previous review, and it's a considerably more careful change than what I asked for — I've dismissed the stale `CHANGES_REQUESTED` so it isn't blocking. One real finding below, plus one leftover minor.

**Verification on my side:** `go build ./...` ok · `go test ./pkg/containerprofilemanager/... ./pkg/metricsmanager/...` all pass · `gofmt`/`vet` clean · **`-race` on the queue package now passes**, so the `MockProfileCreator` race that fails on `main` got fixed here too — worth calling out in the description, since the previous revision listed it as pre-existing and untouched.

### What landed

- The 413 no longer maps to `ObjectTooLargeError`. The new `failureSplit` kind, and the decision to run both sentinel checks *before* the status-code check so an authoritative aggregate verdict still wins over a bare 413, is exactly the right split of the two meanings.
- Splits inherit `Attempts` rather than consuming the retry budget.
- Drops are counted, labelled with a closed `dropReason` set, and exported — that closes the "silent drop with no metric" gap I raised.
- `TestQueuePersistsSplitDepth` covers `Attempts`/`SplitDepth`/`IsStitch` surviving a `Close`/reopen, which was the test I'd flagged as missing.

I re-derived the storage-side claims the design rests on rather than trusting the comments, and they hold: `mergeContainerProfileTS` does *assign* `SeccompProfile`/`ImageID`/`ImageTag` while appending everything else, and `AfterCreate` does hardcode `hasData=true` — so `stitchChunk` keeping exactly those three scalars is correct, and a naively empty stitch really would have zeroed them on the aggregate.

One thing you can relax: `DeflateContainerProfileSpec` runs `DeflateSortString(Architectures)` and `DeflateLabelSelectorRequirement(MatchExpressions)` before the size count, so the duplication a split introduces in those fields is deduped server-side regardless. Clearing `Architectures` in the stitch is a nice-to-have rather than load-bearing — the `SeccompProfile`/`ImageID`/`ImageTag` part is the part that matters.

### Leftover minor

`matchesSentinel` still uses an unanchored `strings.Contains`. I raised this last time and the check ordering is now deliberate, which makes that substring arm load-bearing for the ordering decision: a 413 whose body happened to contain `"object is too large"` would classify terminal and end learning. QueueManager's own body is `"Request entity too large"` so it doesn't bite today, but a proxy or ingress returning arbitrary text could. Narrowing the substring arm to `*apierrors.StatusError` — where you know the message is a relayed sentinel and `errors.Is` genuinely can't work — would remove the last text-matching hazard without losing the prefixed-sentinel case.

The `chainHalves` shim, its TEMPORARY marker and the pointer to kubescape/storage#352 are well judged, as is deriving the base name from storage's own `file.SplitProfileName` rather than reimplementing the suffix heuristic.

Issue #721, opened 2026-02-13
Absolute paths partially changed to relative paths for OPEN_calls in Rule0002


entlein:
# Description
The OpenEvents are capturing paths that are missing their root nodes, examples are
```"path":"/46/task/46/fd"``` ( /proc is missing)
or 
```{"alertName":"Files Access Anomalies in container","arguments":{"apChecksum":"72b6f0c31555285bc48f30fab4e55093d40ce0f58951dcd07b097b521bca30fa","flags":["O_RDONLY","O_CLOEXEC"],"message":"Unexpected file access detected: runc:[2:INIT] with PID 304615 to ","path":""``` (not sure what path RUNC is reporting here "identifiers":{"process":{"name":"runc:[2:INIT]"},"file":{"name":".","directory":"."}}} 

In the applicationProfile, we now find a mention to a "." or sometimes "/." path with all flags

```
    opens:
    - flags:
      - O_APPEND
      - O_CLOEXEC
      - O_CREAT
      - O_DIRECTORY
      - O_EXCL
      - O_NONBLOCK
      - O_RDONLY
      - O_RDWR
      - O_WRONLY
      path: .
```

I can provide similar instances, of "headless" paths, the easiest is to look in an applicationProfile for e.g. the webapp example.

# Environment
OS: ` ubuntu 24 LTS (but I dont think it matters)` 
Version: ` 1.30.2` 
 
# Steps To Reproduce
helm install kubescape-operator version 1.30.2 on k3s v1.35.0+k3s3 with 
default-rules.yaml
```
    - description: Detects unexpected file access that is not in the baseline
      enabled: true
      expressions:
        message: >-
          'Unexpected file access detected: ' + event.comm + ' with PID ' +
          string(event.pid) + ' to ' + event.path
        ruleExpression:
          - eventType: open
            expression:  >-
              !ap.was_path_opened(event.containerId, event.path)
        uniqueId: event.comm + '_' + event.path
      id: R0002
      isTriggerAlert: true
      mitreTactic: TA0009
      mitreTechnique: T1005
      name: Files Access Anomalies in container
      profileDependency: 0
      severity: 1
      supportPolicy: false
      tags:
        - anomaly
        - file
        - open
        - applicationprofile
```

# Expected behavior
The old behavior, e.g. for /proc was
```"path":"/proc/46/task/46/fd"```

# Actual Behavior
```"path":"/46/task/46/fd"```

# Additional context
[...]

PR #872, opened 2026-08-01
addon fix for the old issue to /proc full paths not being recorded

entlein:
We re seeing again :

 - "/17/setgroups" → "/proc/17/setgroups"
  - "/1/gid_map" → "/proc/1/gid_map", "/1/uid_map" → "/proc/1/uid_map"
  
  This is a band-aid type patch for the resurface of  https://github.com/kubescape/node-agent/issues/721 

In my tests on dragonfly and argocd, it worked, but that list is not exhaustive, so will look if we have a real regression as soon as I have some time
 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of additional process metadata and mapping paths under `/proc`, including namespace, user/group mapping, status, and cgroup paths.
  * Preserved correct behavior for non-process paths that contain numeric path segments.

### DIFF
--- a/pkg/utils/normalize_path_test.go
+++ b/pkg/utils/normalize_path_test.go
@@ -51,6 +51,30 @@ func TestNormalizePath(t *testing.T) {
 			expected: "/proc/46/task",
 		},
 		{
+			// #721 regression: runc:[2:INIT] user-namespace-setup paths outside
+			// the old (task|fd) allowlist previously leaked /proc-less.
+			name:     "headless proc path (setgroups)",
+			input:    "/17/setgroups",
+			expected: "/proc/17/setgroups",
+		},
+		{
+			name:     "headless proc path (gid_map)",
+			input:    "/1/gid_map",
+			expected: "/proc/1/gid_map",
+		},
+		{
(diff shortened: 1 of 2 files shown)


matthyx (approved):
do you need a release with that @entlein ?

entlein:
@matthyx thank you! (I dont need a release for this, maybe once we merge all the `chore: deprecate ap/nn with cp` it would merit a release.
I was using my fork build with this patch stacked on the `chore` PRs, and I didnt find any issues -> other than the CEL rules need a replacement of ap.* -> cp.*  
I need to check if I opened that PR already, otherwise, from a functional PoV , it seems to be working :)

entlein:
Actually, for this issue, I m a bit miffed that we didnt catch the regression sooner ... I need to add to my backlog to write a test, so next time we loose the /proc prefix again, the CTs catch it

matthyx:
> Actually, for this issue, I m a bit miffed that we didnt catch the regression sooner ... I need to add to my backlog to write a test, so next time we loose the /proc prefix again, the CTs catch it

I agree, this is weird

PR #873, opened 2026-08-02
feat(processtree): populate the process start time (SUB-7845)

AlonLiwsky:
## Summary

`armotypes.Process.StartTime` was declared but assigned nowhere in this repo: the periodic `/proc` scan read process stats and skipped the start-time field, so it was a hardcoded zero. This populates it, so later work can tell recycled process ids apart.

**Nothing reads the new values yet.** This is the groundwork for network-stream process attribution (SUB-7786) and pid-reuse hardening (SUB-7846), separated deliberately from the changes that give the value meaning.

Two representations, and the split matters:

- **Boot-relative nanoseconds** — `/proc/<pid>/stat` field 22 (clock ticks since boot) converted once via `ticks × 10⁷` (USER_HZ=100). Lives in a new pid-keyed side map in the process-tree creator, exposed as `GetProcessBootTimeNs`. **This is the sole identity source.**
- **Wall-clock `Process.StartTime`** on tree nodes — derived from `/proc/stat` btime, which has whole-second resolution. **Display only.** Comparing it for identity is wrong by up to the btime skew.

The conversion happens exactly once, on the way in. Nothing downstream rescales: a division at emission would compile, parse, and even join correctly *within* one message while silently breaking cross-message identity by seven orders of magnitude.

### The part that would otherwise silently do nothing

Three separate functions build a process node from an explicit field list and strip anything absent from it. All three omitted `StartTime`, so populating the source alone changes nothing downstream — a change that looks like it worked:

- `buildBranchToShim` — feeds every alert branch and every stream tree
- `CopyProcess` — the alert bulk manager's merged tree
- `EnrichProcess` — the bulk manager's merge of overlapping chains

Each now carries the field and each has its own named test, so a future field-strip regression fails by name rather than showing up as an empty value on the wire.

### Coverage

The periodic scan runs every 30 seconds, which would leave every process shorter than one interval at zero — and short-lived processes are both the bulk of beacon-style connections and the drivers of pid churn. So a fork or exec that **creates** a node also reads field 22 on demand: same kernel source, same semantics, just fetched earlier. One read per node creation, never per event. A failed read (process already gone) leaves zero rather than guessing.

This is **not** the rejected "derive the start time from an exec event" option. That rejection was of event *timestamps* — an exec stamps the exec instant, not creation, so the value would change on re-exec. `convertExecEvent`/`convertForkEvent`/`convertExitEvent` still set `StartTimeNs` from the event wall-clock timestamp (epoch ns — a different clock domain), those assignments are untouched, and two tests pin that the value can never reach the side map.

Known gaps, both accepted: in Kubernetes mode the tree refuses to create nodes for host processes, so pre-existing host daemons stay at zero; and a process that dies before its creation event is processed stays at zero.

### Recycled-pid guard

A fork's pid is newborn, so a surviving side-map entry can only belong to a process the kernel already recycled the pid away from. Exits linger for `exitCleanup.cleanupDelay` (5 minutes by default), so this is readily reachable: A exits on pid 4242, the kernel hands 4242 to B, B's fork event arrives, B reports A's creation time — and if B lives under one scan interval the scan never corrects it. That is worse than the zero it replaced, because a consumer joining on `(pid, startTime)` then concludes A and B are one process.

**This guard is scoped to the side map this PR introduces and does *not* implement pid-reuse hardening.** The shared tree node still carries the dead process's comm, cmdline and path — that is SUB-7846's job, and it remains to be done.

### Cost of the on-demand read

A single `/proc/<pid>/stat` open-read-parse measures **~7.5 µs** (Go benchmark, Linux container, warm). **Every fork event performs one**, because the recycled-pid guard drops any stale value before the read decision is made — so the "skip when already known" shortcut does not apply to the fork path.

Fork runs to thousands per second on a busy node. At 1,000/s that is ~7.5 ms of read time per second; at 3,000/s, ~22 ms/s. Under `pt.mutex` — the tree-wide write lock that every alert type contends on — that would all be hold time blocking readers.

So the fork path reads **before** acquiring the lock and stores the result under it: the same number of reads, none of them holding the lock, and zero added hold time from fork. This is a local reordering inside code this PR introduces, not a change to lock scope or discipline. Exec keeps its read inside `ensureStartTime`, where the skip-when-known check makes it conditional and it fires only on first sight of a pid.

`TestHandleForkEvent_ReadsStartTimeWithoutHoldingTreeLock` pins the property, using `TryRLock` so a regression fails rather than deadlocks.

The tradeoff: reading before the lock widens the window between event and read, so a pid recycled inside it yields the new incarnation's value. That is the same already-accepted race, slightly wider, and the reuse hardening is what detects it.

### Note for reviewers: a live wire value changes

`armotypes.Process.StartTime` is a `time.Time` tagged `omitempty`, and Go's encoder ignores `omitempty` on structs — so alert payloads **already** ship `"startTime":"0001-01-01T00:00:00Z"` today. This changes an existing value from the zero time to a real time rather than adding a new field. Nothing branches on it and there is no hash or equality over `Process` anywhere in the repo, so it is still additive in effect, but it is not invisible to consumers.

Out of scope and untouched: `cmd/host`, `cmd/ecs`, `pkg/hostnetworksensor`, and the Kubernetes-mode streaming gate.

## Testing

Run in a Linux container — this repo cannot be built or tested on macOS, because `inspektor-gadget/pkg/utils/host` excludes darwin and every package under `pkg/processtree` imports it transitively:

```bash
docker run --rm -v "$PWD":/src -v "$(go env GOMODCACHE)":/go/pkg/mod -w /src \
  -e GOFLAGS=-mod=mod golang:1.25 go test ./pkg/processtree/... ./pkg/utils/...
```

Test-driven throughout: every test was watched failing for the right reason before the fix, at value level rather than only at compile level. Where a test was written after its implementation, the regression it claims to catch was verified by reverting the fix and watching it fail.

- Both tick→nanosecond conversions are pinned against an independently read field 22 times a literal `10⁷`, so drift between the feeder's and the creator's constant fails deterministically. This replaced weaker checks: `ns % 10⁷ == 0` only catches a 10× error when `ticks % 10 != 0`, and a wall-clock proximity check loses sensitivity on a freshly booted node.
- `TestManager_GetContainerProcessTree_CarriesStartTime` walks the real production path — procfs event → `ReportEvent` → creator → `buildBranchToShim` → `GetContainerProcessTree` — against a creator-populated tree rather than a hand-built fixture.

**Full-suite baseline.** `go test ./pkg/...` was recorded on a pristine checkout of `origin/main` before any change. The failure set is identical after: **19 failing tests before, 19 after, same names** — all environmental (`pkg/containerwatcher/v2/tracers` needs a `tracers.tar` build artifact; `pkg/validator` needs privileges to mount `/sys/fs/bpf`). `go vet` clean; touched files gofmt-clean.

**On `-race`:** `pkg/processtree/creator` reports 7 data races. They are **pre-existing and not from this PR** — every frame sits in `exit_manager.go` around `exitCleanupStopChan` (`startExitManager`/`stopExitManager`/`exitCleanupLoop`). I ran `-race` on a detached checkout of `origin/main` and got the same 7 at the same code sites. That is SUB-7847, which has a fix in flight on a separate branch. No frame touches code added here.

## AI Review

Reviewed by **Claude Opus 5** in a fresh context with no prior knowledge of the change, against `armosec-shared-rules:code-review-standards`.

**Verdict: APPROVE WITH COMMENTS.**

Findings acted on:

- **The on-demand reader had no test coverage** — every other test on that path injects a fake reader, so the real `/proc` parse and its constant were unexercised. Confirmed by mutating the constant to `10⁶` and watching the whole suite stay green. Both conversions are now pinned deterministically, as described above.
- **A recycled pid inherited the dead process's start time** — mechanism reproduced, guard and regression test added above.
- **Reader cleanups** — `/proc` is now resolved once instead of re-stating the mount point per call, and the two btime/mount failure modes log instead of degrading silently.

## Post-review changes (CodeRabbit)

CodeRabbit reviewed this PR and raised four points; all four are addressed.

- **Real defect, mine.** The recycled-pid guard cleared the boot-ns side-map entry but left the *dead* process's wall-clock `Process.StartTime` on the reused node, because `ensureStartTime` only assigns the display value when it is zero. For exactly the case the guard exists to handle, the identity value and the display value disagreed. My test asserted only the accessor, which is why it passed. Fixed, and the test now asserts both; the reset also moved to where reuse is actually detected, closing a narrower case where the side-map entry is gone but the node survives.
- **`/proc` read under the tree write lock.** Correct, and sharper than when the design accepted it — see *Cost of the on-demand read* above. The fork read is now taken outside the lock.
- **`(pid, startTimeNs)` is not unique.** Documented as a known limitation; the 10 ms tick quantization means a pid recycled inside one tick yields two indistinguishable incarnations. This propagates to consumers building a join key from the tuple, so it is recorded in the feature doc rather than left implicit.
- **Markdownlint MD040.** Fence tagged.

Reviewer confirmed all seven contract points hold, verifying several by mutation: removing `StartTime` from each of the three copy sites, removing the side-map delete in `exitByPid`, and injecting a clock-domain violation into `handleForkEvent` each fail a dedicated test.

Noted and deliberately **not** addressed here, as out of scope: `utils.CreateProcessTree` (`pkg/utils/process.go:129`, reached from the malware manager) is a fourth node builder that still strips `StartTime`, so malware alerts will carry a zero start time while other alert types carry a real one. Tracked separately.

## Ticket

[SUB-7845](https://armosec.atlassian.net/browse/SUB-7845) — Populate the process start time (node-agent).

Parent: [SUB-7784](https://armosec.atlassian.net/browse/SUB-7784). Unblocks [SUB-7786](https://armosec.atlassian.net/browse/SUB-7786) (streaming attribution) and [SUB-7846](https://armosec.atlassian.net/browse/SUB-7846) (pid-reuse hardening).

## Summary by CodeRabbit

* **New Features**
  * Added process creation-time tracking with both precise identity timestamps and display-friendly wall-clock timestamps.
  * Preserved process start times across process trees, container views, event conversion, and process duplication.
  * Added process start-time lookup for known and active processes.

* **Bug Fixes**
  * Prevented stale or recycled process IDs from inheriting previous start-time data.
  * Improved cleanup when processes exit or are removed.

* **Documentation**
  * Documented timestamp semantics, limitations, fallback behavior, and testing requirements.

### DIFF
--- a/docs/features/process-start-time.md
+++ b/docs/features/process-start-time.md
@@ -1,0 +1,149 @@
+# Process start time
+
+Every process in the process tree can carry a creation time. It exists to tell
+apart two processes that share a process id — the kernel recycles ids, and
+without a creation time a recycled id looks like the process that held it before.
+
+There are **two representations, and they are not interchangeable**:
+
+| Value | Where | Unit | Use for |
+|---|---|---|---|
+| `ProcessTreeManager.GetProcessBootTimeNs(pid)` | Creator side map | Nanoseconds since boot (`CLOCK_BOOTTIME`) | **Process identity.** The only value that may be compared. |
+| `armotypes.Process.StartTime` | Tree node, and every alert's process tree | Wall-clock `time.Time` | **Display only.** Never compare it. |
+
+## Why two
+
+The identity value must be exact. The display value is derived through
+`/proc/stat`'s `btime`, which has **whole-second resolution**, so it carries up
+to a second of skew. Two processes created 40 ms apart can share a wall-clock
+start time while their boot-relative values differ correctly.
+
+So: show `Process.StartTime` to a human, compare `GetProcessBootTimeNs` in code.
+Anyone comparing the wall-clock value for identity is wrong by up to the btime
+skew, and will be wrong intermittently rather than consistently.
+
+## Where the value comes from
+
+`/proc/<pid>/stat` field 22 (`starttime`), the process's creation time in clock
+ticks since boot. It is converted **once**, on the way in:
+
+```
+boot-relative nanoseconds = ticks × 10,000,000        (USER_HZ = 100)
+```
+
+Nothing downstream rescales it. A division at emission would compile, parse and
+even join correctly *within* a single message, while silently breaking identity
+across messages by seven orders of magnitude. If you find yourself scaling this
+value on output, that is a bug.
+
+**The unit implies more precision than the value carries.** USER_HZ is 100, so
+every value is an exact multiple of 10 ms. That is fine for distinguishing
+recycled process ids; do not treat it as a timestamp.
+
+**It is never derived from an event timestamp.** `convertExecEvent`,
+`convertForkEvent` and `convertExitEvent` set `ProcessEvent.StartTimeNs` from the
+event's wall-clock timestamp — epoch nanoseconds, a different clock domain, and
+stamping the event instant rather than process creation. That value feeds
+pending-exit sort ordering only and must never reach the identity map. An exec
+event in particular stamps the exec, so a re-exec would change it.
+
+## When it gets populated
+
+Two paths, both reading the same kernel source:
+
+- **The periodic `/proc` scan** (every 30 s) — covers processes that predate the
+  agent, so coverage completes within one interval of startup.
+- **On demand, when a fork or exec creates a tree node** — without this, every
+  process shorter than one scan interval would carry zero, and short-lived
+  processes are both the bulk of beacon-style connections and the drivers of
+  process-id churn.
+
+The on-demand read happens **once per node creation, never per event**. An exec
+on a node that already has a value does not re-read: creation time cannot change
+on exec.
+
+### Zero means unknown
+
+A read fails if the process is already gone. The value stays zero rather than
+being guessed. Callers must treat `0` as "unknown", never as "created at boot".
+
+Two populations keep a zero start time, both accepted:
+
+- Processes that died before their creation event was processed (sub-second
+  queue latency).
+- In Kubernetes mode, pre-existing **host** (non-container) processes — the tree
+  refuses to create nodes for them at all.
+
+### Recycled process ids
+
+A fork's process id is newborn, so a surviving identity entry can only belong to
+a process the kernel already recycled that id away from. The fork handler drops
+the stale entry before reading, so the new process gets its own creation time.
+
+This matters because exits linger: `exitCleanup.cleanupDelay` defaults to five
+minutes, so a dead process's tree node and its identity entry outlive it by that
+long.
+
+> **This guard covers the identity map only.** It is **not** process-id-reuse
+> hardening. During that same window the shared tree node still carries the dead
+> process's `comm`, `cmdline` and `path`, so an alert can still name the wrong
+> command. That is tracked separately.
+
+The entry is deleted at the same point the tree node is deleted, so the map
+cannot outgrow the tree.
+
+## Reading it
+
+```go
+// Identity — the only value safe to compare.
+ns := processTreeManager.GetProcessBootTimeNs(pid)
+if ns == 0 {
+    // Unknown: not yet scanned, died before its creation event, or a host
+    // process in Kubernetes mode. Fall back to process-id-only handling.
+}
+
+// Display — carried on every node of every alert's process tree.
+node.StartTime // wall clock, up to 1s of skew, never compare
+```
+
+`Process.StartTime` reaches consumers only because three separate copy functions
+were taught to carry it. Each builds a node from an **explicit field list** and
+silently drops anything absent from it:
+
+- `containerprocesstree.buildBranchToShim` — every alert branch and every
+  network-stream tree
+- `utils.CopyProcess` — the alert bulk manager's merged tree
+- `utils.EnrichProcess` — the bulk manager's merge of overlapping chains
+
+Each has a dedicated test asserting the field survives. **If you add a field to
+`armotypes.Process` and expect it downstream, you must update all three** —
+otherwise it is populated at the source and silently empty at the consumer.
+
+One builder deliberately still strips it: `utils.CreateProcessTree`
+(`pkg/utils/process.go`), used by the malware manager, so malware alerts carry a
+zero start time. Tracked separately.
+
+## A note on the wire
+
+`armotypes.Process.StartTime` is a `time.Time` tagged `omitempty`, and Go's
+encoder ignores `omitempty` on structs. The field has therefore always been
+serialised — as `"startTime":"0001-01-01T00:00:00Z"` — so populating it changes
+an existing value rather than adding a new one. Nothing branches on it and there
+is no hash or equality over `Process`, but consumers do see the change.
+
+## Testing
+
+This package cannot be built or tested on macOS: `inspektor-gadget/pkg/utils/host`
+excludes darwin and everything under `pkg/processtree` imports it transitively.
+
+```bash
+docker run --rm -v "$PWD":/src -v "$(go env GOMODCACHE)":/go/pkg/mod -w /src \
+  -e GOFLAGS=-mod=mod golang:1.25 go test ./pkg/processtree/... ./pkg/utils/...
+```
+
+The tick conversion is defined once per package (feeder and creator). Both are
+pinned in tests against an independently read field 22 times a literal `10^7`,
+so if the two ever drift, a test fails deterministically. Weaker forms of that
+check are not enough: `ns % 10^7 == 0` only catches a 10× error when
+`ticks % 10 != 0`, and comparing the wall-clock value against "now" loses
+sensitivity on a freshly booted machine.
--- a/docs/features/process-start-time.md
+++ b/docs/features/process-start-time.md
@@ -27,7 +27,7 @@ skew, and will be wrong intermittently rather than consistently.
 `/proc/<pid>/stat` field 22 (`starttime`), the process's creation time in clock
 ticks since boot. It is converted **once**, on the way in:
 
-```
+```text
 boot-relative nanoseconds = ticks × 10,000,000        (USER_HZ = 100)
 ```
 
@@ -37,8 +37,20 @@ across messages by seven orders of magnitude. If you find yourself scaling this
 value on output, that is a bug.
 
 **The unit implies more precision than the value carries.** USER_HZ is 100, so
-every value is an exact multiple of 10 ms. That is fine for distinguishing
-recycled process ids; do not treat it as a timestamp.
+every value is an exact multiple of 10 ms. Do not treat it as a timestamp.
+
+### `(pid, startTimeNs)` is not guaranteed unique
+
+The 10 ms quantization means two processes created within the same tick share a
+start time. For a **join key** built from `(pid, startTimeNs)` — which is what
+the network-stream attribution work uses — that leaves a residual collision: a
+process id recycled and reused inside the same 10 ms window produces two
+incarnations with an identical tuple, and they are indistinguishable.
+
+This is narrow but real, and it is a limitation of the source rather than of
+this code — `/proc` does not expose anything finer. Treat the tuple as a strong
+discriminator, not a unique identifier, and do not build logic that assumes
+two matching tuples must be the same process.
 
 **It is never derived from an event timestamp.** `convertExecEvent`,
 `convertForkEvent` and `convertExitEvent` set `ProcessEvent.StartTimeNs` from the
--- a/docs/features/process-start-time.md
+++ b/docs/features/process-start-time.md
@@ -122,13 +122,31 @@ This matters because exits linger: `exitCleanup.cleanupDelay` defaults to five
 minutes, so a dead process's tree node and its identity entry outlive it by that
 long.
 
-> **This guard covers the identity map only.** It is **not** process-id-reuse
-> hardening. During that same window the shared tree node still carries the dead
-> process's `comm`, `cmdline` and `path`, so an alert can still name the wrong
-> command. That is tracked separately.
+The wipe is gated on the pid actually having a **pending exit**. An existing node
+alone does not prove reuse — some other path may simply have created it first —
+and wiping on that weaker signal would discard a good scan-recorded value
+whenever the on-demand read then fails.
 
-The entry is deleted at the same point the tree node is deleted, so the map
-cannot outgrow the tree.
+> **This guard covers the identity map only.** It is **not** process-id-reuse
+> hardening, and its benefit is temporary.
+>
+> The dead process's entry in `pendingExits` survives the fork. When the delayed
+> cleanup fires — up to `cleanupDelay` after the *original* exit — it runs
+> `exitByPid` on that pid and deletes what is by then the **live** successor's
+> node, along with the start time just read for it. The pid reverts to reading
+> as unknown.
+>
+> During the same window the node also still carries the dead process's `comm`,
+> `cmdline` and `path`, so an alert can name the wrong command.
+>
+> Both are the pre-existing pid-reuse behaviour, tracked separately. The fix is
+> to retire the predecessor when a fork reuses a pid with a pending exit —
+> reparenting its children rather than merely dropping the pending entry, which
+> would leave them to be inherited by the successor.
+
+The entry is deleted at the same point the tree node is deleted. Note that per
+the above, the node deleted may by then belong to a different process than the
+one whose exit scheduled the deletion.
 
 ## Reading it
 
--- a/docs/features/process-start-time.md
+++ b/docs/features/process-start-time.md
@@ -74,6 +74,32 @@ The on-demand read happens **once per node creation, never per event**. An exec
 on a node that already has a value does not re-read: creation time cannot change
 on exec.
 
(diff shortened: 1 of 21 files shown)


pkg/processtree/creator/processtree_creator.go
```diff
@@ -139,11 +167,49 @@ func (pt *processTreeCreatorImpl) UpdatePPID(proc *armotypes.Process, event conv
...
 
+	// A fork's pid is newborn, so a surviving side-map entry can only belong to a
+	// process the kernel already recycled this pid away from. Drop it so
+	// ensureStartTime reads this incarnation's own creation time instead of
+	// inheriting the dead one's — exits linger for exitCleanup.cleanupDelay
+	// (5 minutes by default), so the stale entry is readily reachable.
+	//
+	// Scoped to this side map. This is NOT pid-reuse hardening: the shared tree
+	// node still carries the dead process's comm, cmdline and path.
+	delete(pt.pidStartTimeNs, event.PID)
```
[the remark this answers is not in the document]
AlonLiwsky: Fixed in a093e8ea — this was a real defect and my test missed it.

You are right that `ensureStartTime` only assigns `proc.StartTime` when it is zero, so the reused node kept the dead process's wall-clock value while the side map got the new one. Reproduced it first (the node reported the old time, the accessor the new one), then fixed.

I moved the reset to where reuse is actually detected — an existing node on a fork event — rather than clearing the side map unconditionally at the top of the handler. That also closes a narrower case my original placement missed: if the side-map entry is already gone but the node survives (the reparent-error early return in `exitByPid` leaves both, but the paths are not perfectly symmetric), the display value would still have gone unrestamped.

The test now asserts the node's display value alongside the accessor, so the two cannot drift apart again.

docs/features/process-start-time.md
```diff
@@ -0,0 +1,149 @@
...
+## Why two
+
+The identity value must be exact. The display value is derived through
+`/proc/stat`'s `btime`, which has **whole-second resolution**, so it carries up
+to a second of skew. Two processes created 40 ms apart can share a wall-clock
+start time while their boot-relative values differ correctly.
+
+So: show `Process.StartTime` to a human, compare `GetProcessBootTimeNs` in code.
+Anyone comparing the wall-clock value for identity is wrong by up to the btime
+skew, and will be wrong intermittently rather than consistently.
```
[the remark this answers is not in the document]
AlonLiwsky: Fixed in a093e8ea.

Agreed, and this matters beyond the docs — the network-stream attribution work builds its join key from exactly this tuple, so "strong discriminator" versus "unique identifier" is a distinction its consumers need. Added a section stating that a pid recycled and reused inside the same 10 ms tick yields two incarnations with an identical `(pid, startTimeNs)`, that they are indistinguishable, and that logic must not assume two matching tuples are the same process.

Worth noting the limitation is in the source rather than this code: `/proc` field 22 is denominated in USER_HZ ticks and exposes nothing finer. I have flagged it to the workstream owning the rejoin so the constraint is not rediscovered downstream.

docs/features/process-start-time.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Where the value comes from
+
+`/proc/<pid>/stat` field 22 (`starttime`), the process's creation time in clock
+ticks since boot. It is converted **once**, on the way in:
+
+```
+boot-relative nanoseconds = ticks × 10,000,000        (USER_HZ = 100)
+```
```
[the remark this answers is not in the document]
AlonLiwsky: Fixed in a093e8ea — fence tagged `text`.

pkg/processtree/creator/processtree_creator.go
```diff
@@ -139,11 +167,49 @@ func (pt *processTreeCreatorImpl) UpdatePPID(proc *armotypes.Process, event conv
...
+	}
+	ns, wall := pt.readStartTime(pid)
+	if ns == 0 {
+		return
+	}
+	pt.pidStartTimeNs[pid] = ns
+	if !wall.IsZero() && proc.StartTime.IsZero() {
+		proc.StartTime = wall
+	}
+}
```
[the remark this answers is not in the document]
AlonLiwsky: Holding this one open rather than acting on it, and I want to be straight about why.

The observation is correct: `ensureStartTime` does `/proc` I/O under `pt.mutex`, which is the tree-wide write lock, and fork/exec is the highest-volume path. That cost was analysed and consciously accepted when this was designed — the reads are served from kernel memory with no I/O wait, and the periodic scan already does `Stat` + `Status` + `CmdLine` + `Cwd` + `Executable` + `Namespaces` for every pid on the node every 30 seconds. An independent review of this branch also assessed it as a performance tradeoff rather than a correctness problem: no lock inversion, no deadlock, and `-race` reports nothing in the new code.

Your point did get sharper after the recycled-pid guard landed, though, and I do not want to wave that away. Because the guard now drops the stale entry whenever a fork finds an existing node, the fork path no longer benefits from the "skip when already known" shortcut in the reuse case, so those forks always read.

The reason I am not restructuring it here: this PR is deliberately additive, and the process tree is shared by every alert type, not just the consumer this field was added for. Moving the read outside the lock changes lock discipline on that shared structure and introduces a check-then-act window that needs its own thought — a bigger and riskier change than anything else in this PR, in code paths nothing here otherwise touches. Landing it alongside a field that no consumer reads yet is the wrong order.

It is with the owner to either take as a follow-up ticket or dismiss. I will update this thread with the outcome either way rather than leaving it silently unresolved.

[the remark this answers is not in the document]
AlonLiwsky: Reversing my earlier reply — you were right, and my reasoning was stale. Fixed in 183225aa.

The design did accept this read under `pt.mutex`, but it accepted it on the basis that the read is skipped when the value is already known. The recycled-pid guard added in this same PR drops the entry on **every fork**, which removes that shortcut for the fork path entirely. So the cost that was analysed and the cost that actually existed had diverged, and citing the original analysis at you was answering a question nobody had asked.

I measured it rather than continuing to argue from judgement. A single `/proc/<pid>/stat` open-read-parse is **~7.5 µs** (Go benchmark, Linux container, warm) — several times the "a few microseconds" the design assumed. Fork runs to thousands per second on a busy node, so under the lock that is roughly 7.5 ms of hold time per second at 1,000 forks/s, and ~22 ms/s at 3,000/s, all of it blocking every reader of a tree that every alert type uses.

What changed: the fork path now reads before acquiring the lock and stores the result under it. Because a fork always needs the value, that is the same number of reads with none of them holding the lock — a local reordering, not a change to lock scope or discipline, which is the line I did not want to cross in this PR. Exec keeps its read inside `ensureStartTime`, where the skip-when-known check still makes it conditional and it only fires on first sight of a pid.

`TestHandleForkEvent_ReadsStartTimeWithoutHoldingTreeLock` pins it, using `TryRLock` so a regression fails rather than deadlocking the suite.

One honest tradeoff, noted in the PR body and the feature doc: reading before the lock widens the window between the event and the read, so a pid recycled inside that window yields the new incarnation's value. That is the same race the design already accepted, slightly wider, and the reuse hardening is what detects it.

pkg/processtree/creator/processtree_creator.go
```diff
@@ -139,14 +167,77 @@ func (pt *processTreeCreatorImpl) UpdatePPID(proc *armotypes.Process, event conv
...
+		// A fork's pid is newborn, so an existing node belongs to a process the
+		// kernel already recycled this pid away from. Exits linger for
+		// exitCleanup.cleanupDelay (5 minutes by default), so this is readily
+		// reachable. Drop BOTH halves of the stale start time — the identity
+		// value and the display value must never disagree — and let
+		// ensureStartTime restamp them for this incarnation.
+		//
+		// Scoped to the start time. This is NOT pid-reuse hardening: the node
+		// still carries the dead process's comm, cmdline and path.
+		delete(pt.pidStartTimeNs, event.PID)
```
matthyx, line 245: This drops the stale identity but leaves `pendingExits[pid]` holding the dead process's entry. `cleanupDelay` later (5 min default) `exitByPid` then deletes the **live** recycled process's node along with this freshly-read entry — so the guard's benefit expires within the delay and the process goes back to reading as unknown.

Verified on this head: after `performExitCleanup()` the node is gone and `GetProcessBootTimeNs(4242)` is 0 again.

Deleting the pending exit alone isn't right either — A's children would never be reparented and B would inherit them. Retiring the predecessor here does work, and every test in this PR still passes with it:

```go
if _, pending := pt.pendingExits[event.PID]; pending {
    pt.exitByPid(event.PID) // reparents A's children, removes A's node
    ok = false              // fall through to getOrCreateProcess for B
}
```

Fine to defer to SUB-7846 — but then the feature doc's "the entry is deleted at the same point the tree node is deleted" should say that node may by then belong to a different process.

pkg/processtree/creator/processtree_creator.go
```diff
@@ -139,14 +167,77 @@ func (pt *processTreeCreatorImpl) UpdatePPID(proc *armotypes.Process, event conv
...
+		proc.StartTime = time.Time{}
```
matthyx, line 246: Minor: the wipe runs before the read result is consulted, and `applyStartTime` no-ops on `ns == 0`. So a fork event on an existing node whose on-demand read fails leaves both values zero even when the periodic scan had already recorded a correct one.

Reachability is low (the ordered queue is a min-heap on timestamp, so a fork normally lands before the procfs event) and "unknown" beats "wrong", so the trade-off is defensible. But the comment above asserts more than the code can know — an existing node also just means some other path created it first. Gating the wipe on `pendingExits` presence, per the comment above, makes it precise rather than heuristic.

matthyx (commented):
Two notes on the recycled-pid guard, left inline. Everything else checks out: suite green, and the `-race` failures really are pre-existing (every frame lands on `startExitManager`/`stopExitManager`, none on the new code).

pkg/processtree/creator/processtree_creator.go
AlonLiwsky: Fixed in a163e77e — gated the wipe on `pendingExits` presence, as you suggested.

You were right that the comment asserted more than the code could know. Wiping on "an existing node" meant a fork whose on-demand read then failed discarded a value the periodic scan had already recorded correctly, leaving both halves zero. I reproduced that before fixing it. Gating on a pending exit makes the signal precise — the previous holder actually exited — rather than heuristic, and it makes the comment true.

Your reachability assessment was right, and so was "unknown beats wrong" — but that only holds when the alternative is a wrong value, not when we are also throwing away a correct one. With the gate, that case no longer arises.

New test `TestHandleForkEvent_DoesNotWipeKnownStartTimeWithoutAPendingExit` covers it. The recycled-pid test is unchanged and still passes, since a genuine reuse always has the pending exit present.

pkg/processtree/creator/processtree_creator.go
AlonLiwsky: Confirmed — I reproduced it on this head before deciding. After `performExitCleanup()` the node is gone and `GetProcessBootTimeNs(4242)` reads 0 again, so the guard's benefit does expire within `cleanupDelay`. And you are right that dropping the pending entry alone would be worse: A's children would never be reparented and B would inherit them.

Taking your second option and deferring the fix, for a scope reason rather than a disagreement. Retiring the predecessor means calling `exitByPid` from the fork path — reparenting children and deleting nodes — which is shared process-tree lifecycle that every alert type depends on, in a PR that otherwise only populates a field nothing reads yet. It is also already scoped to SUB-7846, which describes flushing a pending exit when a fork reuses a pid as a layer that needs no start time and is independent of this change. Your snippet is essentially that layer, and I would rather it land there with its own tests for the reparenting behaviour than ride in on an additive change.

Doc updated as you asked, and I went a bit further than the one line since the whole paragraph was overclaiming. It now states that the dead process's `pendingExits` entry survives the fork; that the delayed cleanup runs `exitByPid` on what is by then the live successor's node, taking the freshly-read start time with it; that the pid reverts to reading as unknown; and — on the line you quoted — that the node deleted may belong to a different process than the one whose exit scheduled the deletion. It also records that the fix is to retire the predecessor, reparenting rather than merely dropping the pending entry, so whoever picks up SUB-7846 does not have to rediscover why the simpler version is wrong.

Separately, your other comment led to a real fix in the same commit — the wipe is now gated on `pendingExits`.

pkg/processtree/creator/processtree_creator.go
```diff
@@ -139,14 +167,83 @@ func (pt *processTreeCreatorImpl) UpdatePPID(proc *armotypes.Process, event conv
...
+//
+// A zero ns means the read failed — leave the pre-existing zero, never guess.
+//
+// Must be called with pt.mutex held.
+func (pt *processTreeCreatorImpl) applyStartTime(proc *armotypes.Process, pid uint32, ns uint64, wall time.Time) {
+	if ns == 0 {
+		return
+	}
+	pt.pidStartTimeNs[pid] = ns
+	if !wall.IsZero() && proc.StartTime.IsZero() {
```
matthyx: Creation time can't change for a live pid, and both values come from the same /proc field, so overwriting is always correct for the current incarnation.
```suggestion
	if !wall.IsZero() {
```

AlonLiwsky: Good catch, applied.

matthyx (approved):
thanks!

PR #878, opened 2026-08-04
fix(rulebinding): exact namespace match and skip invalid selectors

### LINKED ISSUE #401: admission/rulebinding: NamespaceSelector match uses strings.Contains on NamespaceList.String()
# Description
Admission rule-binding `NamespaceSelector` matching uses `strings.Contains(namespaces.String(), ns)` in `admission/rulebinding/cache/cache.go`. That substring-checks a debug dump, so e.g. namespace `dev` wrongly matches a binding that selected `devel`.

# Environment
OS: `mac os`
Version: operator (current main)

# Steps To Reproduce
1. Create a `RuntimeAlertRuleBinding` whose `namespaceSelector` matches only `devel`.
2. Create/update an object in namespace `dev` that matches `podSelector`.
3. Binding still applies (false positive).

# Expected behavior
Only exact namespace names from the selected list match.

# Actual Behavior
Substring match on `NamespaceList.String()` → wrong namespaces get admission rules/alerts.

manumathon:
## Overview
Fix rulebinding cache matching in `pkg/rulebindingmanager/cache`
fixes https://github.com/kubescape/operator/issues/401

- Match namespaces by exact name instead of `strings.Contains(namespaces.String(), …)` (avoids `dev` matching `devel`)
- Skip bindings when `LabelSelectorAsSelector` fails instead of calling `Matches` on nil

## How to Test
- Unit: `TestNamespaceListHasName`
- Binding with `namespaceSelector` on `devel` only → pod in `devel` matches, pod in `dev` does not

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved pod and namespace label selector validation.
  - Invalid rule bindings are now skipped with parsing errors logged.
  - Namespace matching now uses exact names, preventing incorrect substring matches.
  - Added safe handling for empty or unavailable namespace lists.

### DIFF
--- a/pkg/rulebindingmanager/cache/cache.go
+++ b/pkg/rulebindingmanager/cache/cache.go
@@ -2,7 +2,6 @@ package cache
 
 import (
 	"context"
-	"strings"
 	"sync"
 	"time"
 
@@ -411,14 +410,22 @@ func (c *RBCache) addPod(ctx context.Context, pod *corev1.Pod) []rulebindingmana
 		rbName := uniqueName(&rb)
 
 		// check pod selectors
-		podSelector, _ := metav1.LabelSelectorAsSelector(&rb.Spec.PodSelector)
+		podSelector, err := metav1.LabelSelectorAsSelector(&rb.Spec.PodSelector)
+		if err != nil {
+			logger.L().Warning("RBCache - failed to parse pod selector", helpers.String("ruleBiding", uniqueName(&rb)), helpers.Error(err))
+			continue
+		}
 		if !podSelector.Matches(labels.Set(pod.GetLabels())) {
 			// pod selectors doesnt match
 			continue
 		}
 
 		// check namespace selectors
-		nsSelector, _ := metav1.LabelSelectorAsSelector(&rb.Spec.NamespaceSelector)
+		nsSelector, err := metav1.LabelSelectorAsSelector(&rb.Spec.NamespaceSelector)
+		if err != nil {
+			logger.L().Warning("RBCache - failed to parse ns selector", helpers.String("ruleBiding", uniqueName(&rb)), helpers.Error(err))
+			continue
+		}
 		nsSelectorStr := nsSelector.String()
 		if len(nsSelectorStr) != 0 {
 			// get related namespaces
@@ -427,7 +434,7 @@ func (c *RBCache) addPod(ctx context.Context, pod *corev1.Pod) []rulebindingmana
 				logger.L().Warning("RBCache - failed to list namespaces", helpers.String("ruleBiding", uniqueName(&rb)), helpers.String("nsSelector", nsSelectorStr), helpers.Error(err))
 				continue
 			}
-			if !strings.Contains(namespaces.String(), pod.GetNamespace()) {
+			if !namespaceListHasName(namespaces, pod.GetNamespace()) {
 				// namespace selectors dont match
 				continue
 			}
--- a/pkg/rulebindingmanager/cache/helpers.go
+++ b/pkg/rulebindingmanager/cache/helpers.go
@@ -10,10 +10,26 @@ import (
 
 	k8sruntime "k8s.io/apimachinery/pkg/runtime"
 
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
+
+// namespaceListHasName reports whether name is an exact member of list.Items.
+// Do not use strings.Contains(list.String(), name): NamespaceList.String() is a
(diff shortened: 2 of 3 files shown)


matthyx (approved):
Reviewed the full diff plus the surrounding `RBCache` flow, and verified it locally. **No blockers — approving.**

## Correctness

**Exact namespace match.** `strings.Contains(namespaces.String(), pod.GetNamespace())` matched against a debug dump of the entire `NamespaceList`, so a pod whose namespace name is a substring of *anything* in that dump — another selected namespace name, but equally a label key/value, an annotation, a UID — falsely matched. `namespaceListHasName` is the right fix.

I wrote a throwaway integration-level test driving the real `addPod` path (fake clientset; ns `devel` labelled `env=devel`, ns `dev` unlabelled, binding scoped to `env=devel`):

- on `main`: **FAILS** — pod in `dev` picks up rule `R0001`
- on this branch: **PASSES**

So the reported behaviour genuinely reproduces and this genuinely fixes it.

**Nil selector guard.** This is a real fix, not defensive noise: `metav1.LabelSelectorAsSelector` returns `(nil, err)` on an invalid requirement (apimachinery v0.35.0, `pkg/apis/meta/v1/helpers.go`), so the old `podSelector, _ := ...` followed by `podSelector.Matches(...)` was a nil-interface dereference — a panic in the pod event handler, not a mismatch.

FYI both new branches are unreachable in the current flow: `addRuleBinding` (`cache.go:284-295`) returns before `c.rbNameToRB.Set(...)` on a parse error, and that is the only writer of `rbNameToRB`. Good belt-and-braces regardless.

**In sync with the operator.** This is a faithful port of the fix already merged in `kubescape/operator` `admission/rulebinding/cache/` — the helper is identical and the call-site shape matches. Nice to have the twins aligned.

## Verified locally (on this branch)

```
go build ./...                        ok
go vet  ./pkg/rulebindingmanager/...  ok
go test ./pkg/rulebindingmanager/...  ok
```

The checklist item *"New and existing unit tests pass locally with my changes"* is still unticked — they do pass.

## Non-blocking suggestions

**1. The test covers the helper, not the bug.** `TestNamespaceListHasName` exercises the new function in isolation — it would pass unchanged even if `addPod` still called `strings.Contains`. The existing harness in `cache_test.go` makes a real regression test cheap. This is verbatim what I ran (fails on `main`, passes here):

```go
func TestAddPodNamespaceIsNotPrefixMatched(t *testing.T) {
	c := NewCacheMock("")
	k8sClient := k8sinterface.NewKubernetesApiMock()
	k8sClient.KubernetesClient = k8sfake.NewClientset(
		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "devel", Labels: map[string]string{"env": "devel"}}},
		&corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: "dev"}},
	)
	c.k8sClient = k8sClient

	rb := typesv1.RuntimeAlertRuleBinding{
		ObjectMeta: metav1.ObjectMeta{Name: "testRB", Namespace: "devel"},
		Spec: typesv1.RuntimeAlertRuleBindingSpec{
			PodSelector:       metav1.LabelSelector{MatchLabels: map[string]string{"app": "p"}},
			NamespaceSelector: metav1.LabelSelector{MatchLabels: map[string]string{"env": "devel"}},
			Rules:             []typesv1.RuntimeAlertRuleBindingRule{{RuleID: "R0001"}},
		},
	}
	c.addRuleBinding(&rb)

	podDev := &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "p", Namespace: "dev", Labels: map[string]string{"app": "p"}}}
	c.addPod(context.Background(), podDev)
	assert.Empty(t, c.ListRulesForPod("dev", "p"), "pod in ns 'dev' must NOT match a binding scoped to ns 'devel'")

	podDevel := &corev1.Pod{ObjectMeta: metav1.ObjectMeta{Name: "p", Namespace: "devel", Labels: map[string]string{"app": "p"}}}
	c.addPod(context.Background(), podDevel)
	assert.Len(t, c.ListRulesForPod("devel", "p"), 1, "pod in ns 'devel' must match")
}
```

**2. Reuse `rbName`.** It is already in scope at `cache.go:410`; the two new log lines recompute `uniqueName(&rb)`.

**3. Log the offending selector.** `addRuleBinding` includes `helpers.Interface("PodSelector", ...)`; without it the new warnings don't say *what* failed to parse.

**4. Pre-existing, out of scope for this PR.** `addPod` issues one `Namespaces().List()` API call per rule binding per pod event and then linearly scans the result — while only the pod's own namespace is ever relevant. A namespace lister plus `nsSelector.Matches(labels.Set(ns.Labels))` would be O(1) with no API call. Worth a follow-up if rule-binding counts grow.

**5. Description nit.** `fixes https://github.com/kubescape/operator/issues/401` points at a file in the *operator* repo, and that issue is already closed by the operator-side fix; a cross-repo `fixes` won't close anything from here. "Ports the fix from kubescape/operator#401" would read more accurately.

PR #880, opened 2026-08-04
fix(sbom): classify sidecar oversized images as TooLarge

### LINKED ISSUE #877: SBOM sidecar path misclassifies oversized images as Incomplete instead of TooLarge
## Description

When an image exceeds `maxImageSize`, the **in-process** SBOM path correctly classifies the failure as terminal, but the **sidecar** path does not. On the sidecar path the same deterministic size failure is retried `maxScanRetries` (3) times, is eventually pinned `Incomplete` rather than `TooLarge`, and is reported to the platform as `ReasonSBOMGenerationFailed` instead of `ReasonImageTooLarge`.

This is the same class of bug as kubescape/kubevuln#391 (fixed in kubescape/kubevuln#392), just in node-agent's sidecar transport rather than its scanner logic.

## Where it breaks

The scanner server correctly detects the condition and signals it with a dedicated gRPC code:

```go
// pkg/sbomscanner/v1/server.go:59-60
if errors.Is(err, syftutil.ErrImageTooLarge) {
    return nil, status.Error(codes.FailedPrecondition, "image size exceeds maximum allowed size")
}
```

But the client only translates `Unavailable` / `Aborted` into a sentinel, and passes everything else through raw:

```go
// pkg/sbomscanner/v1/client.go:92-93
if ok && (st.Code() == codes.Unavailable || st.Code() == codes.Aborted) {
    return nil, fmt.Errorf("%w: %v", ErrScannerCrashed, err)
}
```

So `SbomManager` never recognises the condition and falls into the generic failure branch (`pkg/sbommanager/v1/sbom_manager.go:512-521`):

```go
s.handleGenericFailure(sbomName)
s.reportFailure(notif, imageTag, imageID, scanfailure.ReasonSBOMGenerationFailed, scanErr)
```

`handleGenericFailure` only marks the SBOM after `maxScanRetries` consecutive failures, and marks it `Incomplete`.

`codes.FailedPrecondition` is currently produced in exactly one place and consumed nowhere outside tests:

```
pkg/sbomscanner/v1/server.go:60          <- produced
pkg/sbomscanner/v1/server_test.go:116    <- asserted
pkg/sbomscanner/v1/integration_test.go:135 <- asserted
```

## Expected behaviour

Same as the in-process path already does (`pkg/sbommanager/v1/sbom_manager.go:562-570`):

```go
if errors.Is(srcErr, syftutil.ErrImageTooLarge) {
    if wipSbomHadContent {
        // don't let a content-bearing SBOM reach the TooLarge one-way door; treat
        // it as a generic (retryable, eventually Incomplete) failure instead.
        s.handleGenericFailure(sbomName)
    } else {
        s.markSBOMStatus(sbomName, helpersv1.TooLarge, nil)
    }
    s.reportFailure(notif, imageTag, imageID, scanfailure.ReasonImageTooLarge, srcErr)
}
```

An oversized image should be terminal on the first attempt, marked `TooLarge`, and reported as `ReasonImageTooLarge`.

## Impact

- **Wasted work.** Three full pull attempts per oversized image, each aborting only once the layer stream crosses `maxImageSize`, repeated on every scan cycle that re-enters the retry budget.
- **Wrong user-facing reason.** The platform shows *"Failed to generate software inventory (SBOM) for this image"* instead of *"Image exceeds the maximum size limit for vulnerability scanning"*, so operators have no signal that raising `maxImageSize` is the remedy.
- **Inconsistent status between transports.** The same image yields `TooLarge` in-process and `Incomplete` via the sidecar, so downstream consumers see different terminal states depending on deployment mode.

## Suggested fix

1. Map `codes.FailedPrecondition` to a sentinel in `pkg/sbomscanner/v1/client.go` — e.g. an exported `ErrImageTooLarge` alongside the existing `ErrScannerCrashed`.
2. In `sbom_manager.go`, handle that sentinel next to the existing `ErrScannerCrashed` check (around line 504), reusing the in-process branch's logic — including the `wipSbomHadContent` guard, since `TooLarge` is a one-way door in the storage layer and must never be applied to a content-bearing SBOM.

A test asserting that a `FailedPrecondition` from the sidecar produces `TooLarge` + `ReasonImageTooLarge` (and does not consume the retry budget) would lock this in.

## Environment

- Repo: `kubescape/node-agent` @ `8866b6c` (main)
- Affects the sidecar SBOM-scanner path only; the in-process path is correct.

## Related

- kubescape/kubevuln#391 — the equivalent bug in kubevuln
- kubescape/kubevuln#392 — the kubevuln fix

manumathon:
## Overview

Fix the sidecar SBOM path to classify oversized images the same way as the in-process path. This change maps gRPC `FailedPrecondition` to `ErrImageTooLarge` and updates SBOM handling to mark fresh SBOMs as `TooLarge` while preserving content-bearing SBOMs on the generic `Incomplete` path.

## How to Test

```bash
go test ./pkg/sbomscanner/v1/ -run 'TestIntegration_ImageTooLarge|TestCreateSBOM_ImageTooLarge' -count=1
go test ./pkg/sbommanager/v1/ -run 'TooLargeImmediatelyViaSidecar|PreservesContentOnSidecarTooLarge' -count=1
```

## Related issues/PRs

* Resolved #877

## Summary by CodeRabbit

* **New Features**
  * Images exceeding the maximum allowed size are now explicitly flagged as too large without consuming retry attempts.

* **Bug Fixes**
  * Enhanced error handling to distinguish image size limit violations from other failure conditions, preserving existing SBOM content when applicable.

* **Tests**
  * Added test coverage for oversized image handling scenarios.

### DIFF
--- a/pkg/sbommanager/v1/sbom_manager.go
+++ b/pkg/sbommanager/v1/sbom_manager.go
@@ -509,6 +509,23 @@ func (s *SbomManager) processContainerWithMetadata(notif containercollection.Pub
 				s.handleScannerCrash(sbomName, notif, scanErr, imageTag, imageID, wipSbomHadContent)
 				return
 			}
+			if errors.Is(scanErr, sbomscanner.ErrImageTooLarge) {
+				s.metrics.ReportSBOMScan("error")
+				s.metrics.ObserveSBOMScanDuration("error", scanDuration)
+				logger.L().Ctx(s.ctx).Error("SbomManager - sidecar image too large",
+					helpers.Error(scanErr),
+					helpers.String("namespace", notif.Container.K8s.Namespace),
+					helpers.String("pod", notif.Container.K8s.PodName),
+					helpers.String("container", notif.Container.K8s.ContainerName),
+					helpers.String("sbomName", sbomName))
+				if wipSbomHadContent {
+					s.handleGenericFailure(sbomName)
+				} else {
+					s.markSBOMStatus(sbomName, helpersv1.TooLarge, nil)
+				}
+				s.reportFailure(notif, imageTag, imageID, scanfailure.ReasonImageTooLarge, scanErr)
+				return
+			}
 			s.metrics.ReportSBOMScan("error")
 			s.metrics.ObserveSBOMScanDuration("error", scanDuration)
 			logger.L().Ctx(s.ctx).Error("SbomManager - sidecar scan failed",
--- a/pkg/sbomscanner/v1/client.go
+++ b/pkg/sbomscanner/v1/client.go
@@ -92,6 +92,9 @@ func (c *sbomScannerClient) CreateSBOM(ctx context.Context, req ScanRequest) (*S
 		if ok && (st.Code() == codes.Unavailable || st.Code() == codes.Aborted) {
 			return nil, fmt.Errorf("%w: %v", ErrScannerCrashed, err)
 		}
+		if ok && st.Code() == codes.FailedPrecondition {
+			return nil, fmt.Errorf("%w: %v", ErrImageTooLarge, err)
+		}
 		return nil, err
 	}
 
--- a/pkg/sbomscanner/v1/types.go
+++ b/pkg/sbomscanner/v1/types.go
@@ -11,6 +11,7 @@ import (
 var (
 	ErrScannerCrashed  = errors.New("SBOM scanner sidecar crashed during scan")
 	ErrScannerNotReady = errors.New("SBOM scanner sidecar not ready")
+	ErrImageTooLarge = errors.New("image size exceeds maximum allowed size")
 )
 
 type ScanRequest struct {
--- a/pkg/sbommanager/v1/sbom_manager_reprocessing_test.go
+++ b/pkg/sbommanager/v1/sbom_manager_reprocessing_test.go
@@ -469,6 +469,56 @@ func Test_processContainerWithMetadata_MarksFreshImageTooLargeImmediately(t *tes
 	assert.Equal(t, 1, fake.patchCalls, "a fresh reservation must mark TooLarge on the first occurrence, with no retry budget")
 }
 
+// Test_processContainerWithMetadata_MarksFreshImageTooLargeImmediatelyViaSidecar is the
+// sidecar counterpart to Test_processContainerWithMetadata_MarksFreshImageTooLargeImmediately:
+// a fresh reservation whose sidecar returns sbomscanner.ErrImageTooLarge must still be marked
+// TooLarge immediately, with no retry budget.
+func Test_processContainerWithMetadata_MarksFreshImageTooLargeImmediatelyViaSidecar(t *testing.T) {
+	fake := newFakeSbomClient()
+	notif, imageStatus, imageTag, imageID := testNotifAndImageStatus()
+	sbomName, err := names.ImageInfoToSlug(imageTag, imageID)
+	assert.NoError(t, err)
+
+	mgr := newTestManagerWithScannerErr(fake, "v2.0.0", sbomscanner.ErrImageTooLarge)
+	mgr.processContainerWithMetadata(notif, nil, imageStatus, imageTag, imageID)
+
+	stored := fake.get(sbomName)
+	assert.Equal(t, helpersv1.TooLarge, stored.Annotations[helpersv1.StatusMetadataKey])
+	assert.Equal(t, 1, fake.patchCalls, "sidecar ErrImageTooLarge must mark TooLarge on the first attempt")
+	_, stillBudgeted := mgr.failureRetries.Get(sbomName)
+	assert.False(t, stillBudgeted, "sidecar ErrImageTooLarge must not consume the generic retry budget")
+}
+
+// Test_processContainerWithMetadata_PreservesContentOnSidecarTooLarge is the sidecar
+// counterpart to Test_processContainerWithMetadata_PreservesContentOnTooLarge: TooLarge is a
+// storage-layer one-way door, so a content-bearing SBOM must fall back to generic Incomplete
+// pinning instead.
+func Test_processContainerWithMetadata_PreservesContentOnSidecarTooLarge(t *testing.T) {
+	fake := newFakeSbomClient()
+	notif, imageStatus, imageTag, imageID := testNotifAndImageStatus()
+	sbomName, err := names.ImageInfoToSlug(imageTag, imageID)
+	assert.NoError(t, err)
+
+	good := &v1beta1.SBOMSyft{}
+	good.Name = sbomName
+	good.Annotations = map[string]string{
+		helpersv1.StatusMetadataKey:      helpersv1.Learning,
+		helpersv1.ToolVersionMetadataKey: "v1.0.0",
+	}
+	good.Spec.Syft.Artifacts = make([]v1beta1.SyftPackage, 2)
+	fake.sboms[sbomName] = good
+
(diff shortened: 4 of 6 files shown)


pkg/sbomscanner/v1/client.go
```diff
@@ -92,6 +92,9 @@ func (c *sbomScannerClient) CreateSBOM(ctx context.Context, req ScanRequest) (*S
 		if ok && (st.Code() == codes.Unavailable || st.Code() == codes.Aborted) {
 			return nil, fmt.Errorf("%w: %v", ErrScannerCrashed, err)
 		}
+		if ok && st.Code() == codes.FailedPrecondition {
```
manumathon, line 95: FailedPrecondition is currently the sidecar transport signal for ErrImageTooLarge.

pkg/sbomscanner/v1/types.go
```diff
@@ -11,6 +11,7 @@ import (
 var (
 	ErrScannerCrashed  = errors.New("SBOM scanner sidecar crashed during scan")
 	ErrScannerNotReady = errors.New("SBOM scanner sidecar not ready")
+	ErrImageTooLarge = errors.New("image size exceeds maximum allowed size")
```
matthyx, line 14: **Nit (worth fixing before merge):** this line isn't `gofmt`-clean — the new entry breaks the var block's `=` alignment.

```console
$ gofmt -d pkg/sbomscanner/v1/types.go
-	ErrImageTooLarge = errors.New("image size exceeds maximum allowed size")
+	ErrImageTooLarge   = errors.New("image size exceeds maximum allowed size")
```

No CI gate catches it (`go-basic-tests.yaml` runs build + tests, no `gofmt`/lint step), so this won't turn the PR red — but the next person with format-on-save will produce a spurious diff here. For the record, the `gofmt` findings in `client.go` and `server.go` are import-ordering issues that already exist on `main`; only this one is new.

**Separate, optional:** the repo now has two distinct sentinels with byte-identical messages — `syftutil.ErrImageTooLarge` and this one. `errors.Is` does not relate them, which is exactly why `sbom_manager.go` has to check a different sentinel per branch. That's correct as written, but easy to misread. Since this package already imports `syftutil` (`server.go`), `var ErrImageTooLarge = syftutil.ErrImageTooLarge` would collapse them into one identity and make both branches match the same target. Your call — the current split is defensible if you want the transport sentinel to stay independent of the syft layer.

pkg/sbomscanner/v1/client.go
matthyx, line 95: **Non-blocking, but the consequence is severe enough to flag.** This is correct today — I grepped the repo and `codes.FailedPrecondition` is produced in exactly one place (`server.go:60`), so the mapping is unambiguous.

The concern is what happens when that stops being true. `FailedPrecondition` is a natural code to reach for on a future precondition failure ("layer paths not mounted", "scanner not initialized", ...). The moment a second one is added, this branch silently converts it to `ErrImageTooLarge`, and `sbom_manager.go` writes `TooLarge` — which, per the comment at `sbom_manager.go:381-386`, is a storage-layer one-way door where `GuaranteedUpdate` drops every future write. The SBOM would be permanently frozen and unfixable by any later version. Low probability, unbounded blast radius, and the failure is silent.

Two cheap ways to bound it, pick either:

1. Match code **and** message. You already changed `server.go:60` to send `ErrImageTooLarge.Error()`, but nothing consumes that — the coupling is currently write-only. Both binaries ship from this repo, so message matching is safe here and narrows the check to exactly the intended condition.
2. Keep the code-only check and add a one-line comment at `server.go:60` reserving `FailedPrecondition` for image-too-large, pointing at this branch as the consumer. Cheaper, and enough to stop the next person.

Option 2 alone is fine by me.

pkg/sbommanager/v1/sbom_manager.go
```diff
@@ -509,6 +509,23 @@ func (s *SbomManager) processContainerWithMetadata(notif containercollection.Pub
...
+					helpers.String("namespace", notif.Container.K8s.Namespace),
+					helpers.String("pod", notif.Container.K8s.PodName),
+					helpers.String("container", notif.Container.K8s.ContainerName),
+					helpers.String("sbomName", sbomName))
+				if wipSbomHadContent {
+					s.handleGenericFailure(sbomName)
+				} else {
+					s.markSBOMStatus(sbomName, helpersv1.TooLarge, nil)
+				}
+				s.reportFailure(notif, imageTag, imageID, scanfailure.ReasonImageTooLarge, scanErr)
```
matthyx, line 526: **Test-coverage gap, non-blocking.** #877 asked for a test asserting `TooLarge` **and** `ReasonImageTooLarge` and no retry-budget consumption. The two new tests nail the first and third — but not the reason.

`newTestManagerWithScannerErr` leaves `failureReporter` nil, and `reportFailure` returns immediately on a nil reporter, so this line never executes under test. That matters because the wrong user-facing reason was impact #2 in the issue: operators see *"Failed to generate software inventory (SBOM)"* instead of *"Image exceeds the maximum size limit"*, which is the whole signal that raising `maxImageSize` is the remedy. Right now nothing would catch a regression that swaps this back to `ReasonSBOMGenerationFailed`.

In fairness this is symmetric with existing coverage — the in-process `Test_..._MarksFreshImageTooLargeImmediately` doesn't assert the reason either, and `http_failure_reporter_test.go` only tests the reporter in isolation, not which reason the manager hands it. So this is a pre-existing convention, not a regression you introduced.

A fake reporter capturing `FailureReason` into a slice, wired through `newTestManagerWithScannerErr`, would close it for both paths in a few lines. Happy to see it land as a follow-up rather than hold this PR.

matthyx (approved):
## Approved ✅

This is a faithful, minimal implementation of the fix described in #877. The sidecar branch mirrors the in-process branch semantics exactly — including the `wipSbomHadContent` guard that keeps a content-bearing SBOM away from the `TooLarge` one-way door — and both new tests are the correct counterparts to the existing in-process ones.

### Verification performed locally (at `36de36d`)

| Check | Result |
|---|---|
| `go build ./...` | pass |
| `go vet ./pkg/sbomscanner/... ./pkg/sbommanager/...` | pass |
| `go test ./pkg/sbomscanner/... ./pkg/sbommanager/...` | pass |
| `codes.FailedPrecondition` producers repo-wide | exactly one (`server.go:60`) — the mapping is unambiguous today |

### Correctness review

* `client.go` — the new mapping is placed after the `Unavailable`/`Aborted` → `ErrScannerCrashed` check, so a crashed sidecar still wins. Correct ordering.
* `sbom_manager.go` — the new branch is checked before the generic fallback and after `ErrScannerCrashed`, matching the in-process precedence. It correctly does **not** call `SetSBOMScannerReady(false)`: an oversized image leaves the sidecar healthy.
* `markSBOMStatus(..., TooLarge, nil)` with no `ScannerMemoryLimitAnnotation` is right here — that annotation exists to re-open the OOM-crash path when the memory limit changes, and an image-size limit breach is not memory-driven. Same as the in-process path.
* `Test_..._PreservesContentOnSidecarTooLarge` — I traced the loop: iterations 1–2 only increment the budget, iteration 3 pins `Incomplete`, iterations 4–5 short-circuit on the `Incomplete` + same-tool-version case. `patchCalls == 1` is the correct assertion.
* Wire contract is still covered — the integration test now asserts the sentinel, but `server_test.go:TestCreateSBOM_ImageTooLarge` still asserts `codes.FailedPrecondition` at the raw pb layer, so relaxing the integration assertion loses nothing.
* Backward compatible under sidecar/node-agent version skew in both directions.

CodeRabbit found nothing actionable, and I found no blockers either. Three non-blocking notes inline — the `gofmt` one is worth folding in before merge since it's a one-character fix.

PR #881, opened 2026-08-04
testing a fix for the scramble of the path opens

entlein:
Looks like it could be as simple as this, currently testing for various side effects as tag= `sbob-rc5s-osp` -> Edit: its only half the rent, there is more to the story. This current patch does help, but I havnt had time to check where those `scrambled` lookups were introduced.
More info can be found here:
https://github.com/k8sstormcenter/bob/issues/183

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved path handling when full paths are unresolved or contain stale relative fragments.
  * Prevented invalid path fragments from being treated as valid absolute paths.
  * Added fallback behavior to use the available filename when a resolved full path is unavailable.

* **Tests**
  * Added coverage for relative, absolute, root, and process-related path formats.


entlein:
Closing in favor of #889

PR #882, opened 2026-08-05
fix(node-agent): fix ContainerProfile size accounting in ReportSyscall and ReportNetworkEvent

### LINKED ISSUE #870: containerprofilemanager's pre-send size estimator mixes byte-size and element-count units
### Context

Split off from the review discussion on #866 (https://github.com/kubescape/node-agent/pull/866#pullrequestreview-4794269159). #866 makes node-agent recover gracefully from an HTTP 413 by splitting the oversized chunk, but that only treats the symptom — this issue is the root cause of *reaching* a 413 in the first place.

### Problem

`pkg/containerprofilemanager/v1/config.MaxTsProfileSize` is node-agent's own pre-send threshold, meant to flush a `ContainerProfile` chunk before it grows past what storage's `QueueManager` will accept (`kindQueues.containerprofiles.maxObjectSize`, e.g. 2.5MB). But the running estimate that's compared against this threshold mixes two different units depending on which accumulator is being updated:

- Some accumulators track **byte size** (e.g. `ReportCapability` → `size.Of(...)`).
- Others track **element count** (e.g. `ReportSyscall` → the number of elements appended, not their serialized size).
- Endpoints/network-neighbor fields are also under-counted at estimation time relative to what they expand to at serialization time (e.g. after DNS resolution), so the estimate can miss the actual byte cost of an endpoint entirely.

In production this produced a request body of 3.23MB against a cap of 2.5MB even though `MaxTsProfileSize` defaults to 2MB, i.e. *below* storage's own cap — the flush should have happened well before the 413. It didn't, because the estimator undercounts.

### Fix

Rework `MaxTsProfileSize`'s accounting so every accumulator is measured consistently — either all by estimated serialized byte size, or with a conversion factor applied to element-count fields, and account for endpoint/network-neighbor expansion at estimation time rather than assuming raw event count. Once this is fixed, the splitting mechanism added in #866 should become a rare safety net rather than something regularly exercised in the field.

### Related

- kubescape/storage#350 — the size caps themselves are inconsistent (three caps, three units, ~19% headroom, two not chart-exposed).
- #866 — the client-side recovery mechanism this estimator bug makes necessary.

aryanghai12:
## Overview

- **Current Behavior:** The `MaxTsProfileSize` pre-send threshold undercounted `ContainerProfile` byte sizes due to mixed accumulator units. `ReportSyscall` added `mapset.Set.Append`'s return value (0 or 1) rather than the actual syscall byte size. `ReportNetworkEvent` only measured the raw `NetworkEvent` struct, ignoring post-serialization field expansions (`DNS`, `PodSelector`, `NamespaceSelector`, hashes) added later by `createNetworkNeighbor`. As a result, profiles grew past storage limits instead of flushing early, triggering HTTP 413 payload errors.
- **Future Behavior:** `ReportSyscall` adds `size.Of(syscall)` only when a syscall is newly appended. `ReportNetworkEvent` applies a conservative 256-byte `networkNeighborExpansionEstimate` surcharge on top of `size.Of(networkEvent)` to account for deferred serialization expansions. Profiles now flush accurately before hitting storage caps.

## Additional Information

- **Syscall Accounting:** Fixed the unit mismatch where element-count deltas were being added directly to the byte accumulator.
- **Network Neighbor Expansion:** Because DNS resolution and Service selector lookups are deferred to serialization, exact expansion sizes cannot be calculated at report time; the fixed 256-byte surcharge provides a safe, conservative buffer.

## How to Test

Run the size accounting unit tests:

```bash
go test -v ./... -run 'TestReportSyscallSizeAccounting|TestReportNetworkEventSizeAccounting'
```

## Examples/Screenshots

N/A (backend byte accounting fix)

## Related issues/PRs:

* Resolved #870

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container profile size estimates for network events and syscall reporting.
  * Duplicate syscalls and repeated network reports no longer inflate size calculations.

* **Tests**
  * Added coverage to verify accurate accounting for unique syscalls, network-neighbor data, and duplicate reports.

* **Chores**
  * Updated ignored development artifacts.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ resources/ebpf/falco/*
 node-agent
 __pycache__
 tracers.tar
+LFX_AGENT_SANDBOX_PREP.md
--- a/.gitignore
+++ b/.gitignore
@@ -6,4 +6,3 @@ resources/ebpf/falco/*
 node-agent
 __pycache__
 tracers.tar
-LFX_AGENT_SANDBOX_PREP.md
--- a/pkg/containerprofilemanager/v1/event_reporting.go
+++ b/pkg/containerprofilemanager/v1/event_reporting.go
@@ -19,6 +19,15 @@ import (
 
 var procRegex = regexp.MustCompile(`^/proc/\d+`)
 
+// networkNeighborExpansionEstimate approximates, in bytes, how much a NetworkEvent grows
+// when createNetworkNeighbor() (container_data.go) turns it into a v1beta1.NetworkNeighbor
+// at serialization time: a generated Identifier hash (sha256 hex, 64 bytes), a Type string,
+// a Ports entry, and - when DNS resolution or a Pod/Service selector applies - a DNS name or
+// label-selector map that has no counterpart on the raw event. None of that can be computed
+// exactly at report time, since DNS resolution and the Service selector lookup only happen
+// at serialization, so this is a fixed conservative surcharge rather than a live measurement.
+const networkNeighborExpansionEstimate = 256
+
 // ReportCapability reports a capability event for a container
 func (cpm *ContainerProfileManager) ReportCapability(containerID, capability string) {
 	err := cpm.withContainer(containerID, func(data *containerData) (int, error) {
@@ -273,7 +282,7 @@ func (cpm *ContainerProfileManager) ReportNetworkEvent(containerID string, event
 		}
 
 		data.networks.Add(networkEvent)
-		return size.Of(networkEvent), nil
+		return size.Of(networkEvent) + networkNeighborExpansionEstimate, nil
 	})
 
 	cpm.logEventError(err, "network", containerID)
@@ -299,7 +308,13 @@ func (cpm *ContainerProfileManager) ReportSyscall(containerID string, syscall st
 		if data.syscalls == nil {
 			data.syscalls = mapset.NewSet[string]()
 		}
-		return data.syscalls.Append(syscall), nil
+		// Append returns the number of elements newly added (0 or 1 here), not their
+		// serialized size - using it directly mixed element-count units into a
+		// byte-size accumulator and made syscalls contribute ~0 to MaxTsProfileSize.
+		if data.syscalls.Append(syscall) == 0 {
+			return 0, nil
+		}
+		return size.Of(syscall), nil
 	})
 
 	cpm.logEventError(err, "syscalls", containerID)
--- a/pkg/containerprofilemanager/v1/event_reporting.go
+++ b/pkg/containerprofilemanager/v1/event_reporting.go
@@ -22,65 +22,88 @@ import (
 
 var procRegex = regexp.MustCompile(`^/proc/\d+`)
 
-// networkNeighborExpansionEstimate is the number of bytes a NetworkEvent gains when
-// createNetworkNeighbor() (container_data.go) turns it into a v1beta1.NetworkNeighbor at
-// serialization time. None of these fields exist on the raw event, and none can be computed
-// exactly at report time since DNS resolution and the Service selector lookup are deferred
-// to serialization, so each is sized at its documented worst case instead of guessed:
-//
-//   - Identifier: hex.EncodeToString of a sha256 sum, always exactly 2*sha256.Size bytes.
-//   - Type: the longer of "internal"/"external".
-//   - Ports: createNetworkNeighbor always appends exactly one NetworkPort entry.
-//   - DNS/DNSNames: the longest legal DNS name (RFC 1035 §3.1: 253 bytes), stored once in
-//     DNS and again in DNSNames - previously undercounted entirely (253*2 alone exceeds the
-//     old flat 256-byte guess this replaces).
-//   - NamespaceSelector: getNamespaceMatchLabels always produces exactly one entry keyed
-//     "kubernetes.io/metadata.name", valued with a namespace name (DNS-1123 label, RFC 1123:
-//     63 bytes max) - this bound is exact, not assumed.
-//   - PodSelector: filterLabels forwards whatever label set the destination pod has. Core
-//     Kubernetes caps a label's key (253 bytes, optional DNS-subdomain prefix + "/" + 63-byte
-//     name) and value (63 bytes) but not the number of labels on an object, so no fixed bound
-//     is exact here. maxBudgetedPodLabels labels at that per-label maximum is budgeted as a
-//     documented, deliberately generous headroom for typical (e.g. Helm-templated) workloads;
-//     a pod with more labels than that could still push a profile past MaxTsProfileSize before
-//     this estimator catches it, in which case the queue-level split from #866 is the backstop.
-//
-// A single event never produces DNS and both selectors at once (createNetworkNeighbor takes
-// one branch per Destination.Kind), so summing every component here is deliberately
-// conservative on top of the already-generous per-field bounds.
-var networkNeighborExpansionEstimate = computeNetworkNeighborExpansionEstimate()
+// neighborFixedOverhead is the number of bytes every v1beta1.NetworkNeighbor gains over its
+// source NetworkEvent regardless of Destination.Kind: a generated Identifier hash and a Type
+// string. Both are exactly bounded, so they're measured once rather than guessed:
+// Identifier is hex.EncodeToString of a sha256 sum (always 2*sha256.Size bytes), Type is the
+// longer of "internal"/"external".
+var neighborFixedOverhead = size.Of(strings.Repeat("f", sha256.Size*2)) + size.Of(ExternalTrafficType)
 
-const maxBudgetedPodLabels = 12
-
-func computeNetworkNeighborExpansionEstimate() int {
-	identifier := strings.Repeat("f", sha256.Size*2)
+// maxDNSNameEstimate budgets the one field that's genuinely unknowable at report time on
+// container_data.go's raw/DNS branch: DNS resolution happens at serialization, not when the
+// event is reported. RFC 1035 §3.1 bounds an encoded domain name to 253 bytes;
+// createNetworkNeighbor stores it twice (DNS and DNSNames[0]).
+var maxDNSNameEstimate = func() int {
 	maxDNSName := strings.Repeat("a", 253)
+	return size.Of(maxDNSName) + size.Of([]string{maxDNSName})
+}()
+
+// maxBudgetedServiceLabels bounds maxServiceSelectorEstimate below. Kubernetes Services are
+// conventionally selected on a handful of short labels (e.g. "app: foo"), unlike Pods which
+// can carry many more, so this is deliberately smaller than a Pod label budget would be.
+const maxBudgetedServiceLabels = 6
+
+// maxServiceSelectorEstimate budgets the other field genuinely unknowable at report time:
+// on the Service branch, svc.GetServiceSelector() is fetched from the k8s API at
+// serialization time and has no relationship to anything on the raw event. Each of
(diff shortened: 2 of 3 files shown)


pkg/containerprofilemanager/v1/event_reporting.go
```diff
@@ -15,10 +16,73 @@ import (
...
+		podLabels[maxLabelKey+string(rune('a'+i))] = maxLabelValue
+	}
+	podSelector := &metav1.LabelSelector{MatchLabels: podLabels}
+
+	return size.Of(identifier) +
+		size.Of(ExternalTrafficType) +
+		size.Of(ports) +
+		size.Of(maxDNSName) + size.Of([]string{maxDNSName}) +
+		size.Of(namespaceSelector) +
+		size.Of(podSelector)
```
matthyx: **Blocker.** This sum evaluates to **4571 bytes**, and `size.Of(podSelector)` alone is **3656** of it.

The doc comment above correctly notes that `createNetworkNeighbor` takes exactly one branch per `Destination.Kind`, but the code still adds every branch to every event. Worse, the `PodSelector` budget is not actually unknown at report time: on the `EndpointKindPod` branch the selector is `filterLabels(networkEvent.GetDestinationPodLabels())`, which is parsed from `Destination.PodLabels` — a `string` field on `NetworkEvent` that `size.Of(networkEvent)` already counts, and `filterLabels` only removes entries. So those 3656 bytes are double-counted on the pod branch and entirely fictional on the raw/DNS branch, where `PodSelector` is `nil`.

Only the DNS name (raw branch) and the Service selector (`EndpointKindService`) are genuinely deferred. `Ports` and `NamespaceSelector` are both computable exactly from the event plus the container namespace. Switching on `Destination.Kind` and charging only that branch would keep this conservative without the 6–20x inflation.

pkg/containerprofilemanager/v1/event_reporting.go
```diff
@@ -273,7 +337,7 @@ func (cpm *ContainerProfileManager) ReportNetworkEvent(containerID string, event
 		}
 
 		data.networks.Add(networkEvent)
-		return size.Of(networkEvent), nil
+		return size.Of(networkEvent) + networkNeighborExpansionEstimate, nil
```
matthyx: Concrete effect of the flat surcharge, measured on this branch:

- external IP, no DNS: estimate **4726 B** vs actual `size.Of(*neighbor)` **242 B** → 19.5x
- pod destination with 6 typical Helm labels: estimate **4930 B** vs actual **767 B** → 6.4x

At the default `maxTsProfileSize: 2Mi` that trips `ProfileRequiresSplit` after ~443 unique network events — a real payload of roughly 130–350 KB, i.e. 6–15% of the threshold. Since network neighbors usually dominate a profile, splitting stops being the rare backstop #866 intended and becomes the steady state.

.gitignore
```diff
@@ -5,4 +5,5 @@ temp
 resources/ebpf/falco/*
 node-agent
 __pycache__
-tracers.tar
\ No newline at end of file
+tracers.tar
+LFX_AGENT_SANDBOX_PREP.md
```
matthyx: This one is unrelated to the size-accounting fix (CodeRabbit's out-of-scope check flags it as well), and `LFX_AGENT_SANDBOX_PREP.md` reads as a personal working file rather than a build artifact this repo produces — `~/.config/git/ignore` or `.git/info/exclude` is the better home for it. Please drop it from this PR.

Minor: the file still ends without a trailing newline; adding one here would fix that too.

matthyx (commented):
Thanks for taking this on — the `ReportSyscall` half is exactly right, and the test coverage is genuinely good. One blocker on the network half before this can go in.

## Blocker: the expansion estimate over-counts by ~6–20x, which trades the 413 undercount for constant premature splitting

`networkNeighborExpansionEstimate` evaluates to **4571 bytes**. Component breakdown (measured on `b9ad48c`):

| component | bytes |
|---|---|
| `identifier` | 80 |
| `Type` | 24 |
| `Ports` | 85 |
| `DNS` + `DNSNames` | 562 |
| `NamespaceSelector` | 164 |
| **`PodSelector`** | **3656** |
| **total** | **4571** |

Measured against what `createNetworkNeighbor` actually produces:

| case | estimate (`size.Of(event)` + surcharge) | actual `size.Of(*neighbor)` | ratio |
|---|---|---|---|
| external IP, no DNS | 4726 | 242 | 19.5x |
| pod destination, 6 typical Helm labels | 4930 | 767 | 6.4x |

With the default `maxTsProfileSize: 2Mi`, that means a split fires after ~443 unique network events, whose real serialized weight is ~130–350 KB — roughly 6–15% of the intended threshold. Network neighbors are usually the dominant part of a profile, so in practice this makes `ProfileRequiresSplit` the normal path rather than the safety net #870 asks for ("Once this is fixed, the splitting mechanism added in #866 should become a rare safety net rather than something regularly exercised in the field"). Under-flushing became over-flushing; both miss the target.

The root cause is that the surcharge is unconditional, while `createNetworkNeighbor` takes exactly one branch per `Destination.Kind` — the doc comment says as much, but the code still sums all branches. And critically, most of the budget isn't actually unknown at report time:

- **`PodSelector` for `EndpointKindPod` is already counted.** It comes from `filterLabels(networkEvent.GetDestinationPodLabels())`, which parses `Destination.PodLabels` — a plain `string` field on the `NetworkEvent` that `size.Of(networkEvent)` already includes, and `filterLabels` only ever *removes* entries. So the 3656 bytes are double-counted on the pod branch and pure fiction on the raw/DNS branch (where `PodSelector` is `nil`).
- **`NamespaceSelector` is exactly computable**: the namespaces are both known at report time.
- **`Ports`** is exactly computable from `Port`/`Protocol`.
- Only two things are genuinely deferred: the **DNS name** (raw branch only) and the **Service selector** (`EndpointKindService` only).

Suggested shape — switch on `event.Destination.Kind` and add only that branch's cost:

```go
// fixed for every neighbor
est := neighborFixedOverhead // identifier + Type + Ports

switch networkEvent.Destination.Kind {
case EndpointKindPod:
    // labels already inside size.Of(networkEvent); only the selector wrappers are new
    est += size.Of(namespaceSelectorFor(networkEvent.Destination.Namespace, namespace))
case EndpointKindService:
    est += serviceSelectorBudget // genuinely deferred, but a Service selector is a handful of labels
default:
    est += maxDNSNameBudget // 253*2, raw branch only
}
```

That lands typical events in the few-hundred-bytes range — still conservative, but without turning every profile into a split. Keeping the existing tests as upper-bound assertions works fine with this shape.

## Non-blocking

- **`.gitignore`**: `LFX_AGENT_SANDBOX_PREP.md` is unrelated to this fix (CodeRabbit's out-of-scope check flags it too) and looks like a personal working file — that belongs in your global gitignore or `.git/info/exclude`. Please drop it here. While you're in the file, it'd be nice to end it with a newline.
- **PR description is stale**: it still describes a "conservative 256-byte `networkNeighborExpansionEstimate`", which was `c576f98`; `b9ad48c` replaced it with the computed 4571-byte value. Worth updating before merge so the commit history reads correctly.
- **Other accumulators still over-count** (out of scope for this PR, but #870 asks for *all* accumulators to be consistent, so worth a follow-up): `ReportFileExec` adds `size.Of(exec)` unconditionally even though `data.execs.Set` overwrites the same `execIdentifier` key, so a repeatedly-exec'd binary inflates the estimate without bound. `ReportFileOpen` re-adds `size.Of(path)` when only new flags are appended to an existing path. Both push the same direction as the blocker above.
- **`size.Of` measures in-memory Go size, not the JSON payload storage rejects with a 413.** The new tests assert `estimate >= size.Of(*neighbor)`, which doesn't directly bound the wire size. That's the pre-existing convention in this file so I'm not asking you to change it, but the estimator's relationship to the actual HTTP cap stays approximate.

## Verified locally

`go build ./...` and `go test ./pkg/containerprofilemanager/...` both pass on `b9ad48c`; the numbers above come from instrumenting that build. Happy to re-review as soon as the branch-aware sizing is in.

matthyx (approved):
Blocker resolved — this is the right shape now. Approving.

`fe073b2` replaces the flat surcharge with `networkNeighborIncrement(data, networkEvent)`, charging only the branch `createNetworkNeighbor` will actually take. Re-measured on this commit:

| constant | bytes |
|---|---|
| `neighborFixedOverhead` | 104 |
| `maxDNSNameEstimate` | 562 |
| `maxServiceSelectorEstimate` | 1875 |

| case | estimate | actual `size.Of(*neighbor)` | ratio (was) |
|---|---|---|---|
| external IP, DNS unresolved | 983 | 242 | 4.06x (19.5x) |
| external IP, 253-byte DNS name | 983 | 535 | 1.84x |
| pod dest, 6 Helm labels, cross-ns | 912 | 767 | 1.19x (6.4x) |
| pod dest, 6 Helm labels, same ns | 802 | 682 | 1.18x |
| pod dest, 30 long labels | 4152 | 4007 | 1.04x |

The upper-bound property still holds in every case I tried — no undercounts — and events-until-split at the default `maxTsProfileSize: 2Mi` goes from ~443 back to ~2133 for the external-IP case. The residual 4x on unresolved DNS is inherent (whether `ResolveIPAddress` will hit isn't knowable at report time) and the comment says so; charging the RFC 1035 bound there is the right call.

The `PodSelector` handling is better than what I suggested — subtracting `size.Of(Destination.PodLabels)` and charging only the map-wrapper delta (clamped at zero) is exact rather than budgeted, and it holds up: on the 30-label stress case the estimate lands 145 bytes over a 4007-byte neighbor.

Things I checked before approving:

- **The two namespaces agree.** `networkNeighborIncrement` reads `data.watchedContainerData.Namespace`, which is populated from `container.K8s.Namespace` (`pkg/containerwatcher/v2/containercallback.go:117`) — the same value `monitoring.go:205-206` later passes to `createNetworkNeighbor`. So the estimator and the serializer make the same `getNamespaceMatchLabels` call, not two different ones.
- **No new race.** `watchedContainerData` is read under `entry.mu` inside `withContainer`, the same lock `lifecycle.go:154` writes it under. `go test -race ./pkg/containerprofilemanager/v1/` is clean.
- **Nil `watchedContainerData` errs conservative.** Events reported before shared data is ready see `sourceNamespace == ""`, which over-charges the `NamespaceSelector` rather than under-charging it.
- **Cost is fine.** The increment is ~2.5µs / 42 allocs more than the old constant on the pod branch, but it sits *after* the `data.networks.Contains` early return, so it's paid once per unique neighbor tuple, not per packet. Negligible against a profile's lifetime.
- `.gitignore` now nets out to just adding the missing trailing newline. Thanks.
- `go build ./...`, `go vet`, `gofmt -l`, and `go test ./pkg/containerprofilemanager/...` all clean on `fe073b2`.

Two things to tidy, neither blocking the approval:

1. **Please update the PR description before merging** — it still describes "a conservative 256-byte `networkNeighborExpansionEstimate`", which is two commits stale and describes a symbol that no longer exists. If this lands as a squash merge that text becomes the commit body.
2. **Optional test gaps**: nothing exercises the `EndpointKindService` branch (`maxServiceSelectorEstimate` is the one budget left that's a genuine guess) or the `watchedContainerData == nil` path. Also a theoretical 164-byte undercount if a `EndpointKindPod`/`Service` event ever arrives with an empty `Destination.Namespace` — report time sees `"" == ""` and skips the `NamespaceSelector` while serialization would add it. I don't think that's reachable in practice, so purely FYI.

The follow-ups I mentioned last time on `ReportFileExec` (unconditional `size.Of(exec)` despite `execs.Set` overwriting the same key) and `ReportFileOpen` (re-adding `size.Of(path)` for flag-only updates) are still open, but they're #870's remaining scope rather than this PR's — worth a separate issue.

PR #883, opened 2026-08-05
feat(otel): add pod-level memory metric, harden cgroup scope resolution

matthyx:
## Overview

Adds `node_agent.pod.memory.cgroup_bytes` / `cgroup_limit_bytes`, read from the parent `kubepods-*-pod<UID>.slice` cgroup one level above the container `.scope` dir `findCgroupScopeDir` already resolves. This is what makes third-party sidecars with no OTEL instrumentation of their own (e.g. `clamav`, gated behind `capabilities.malwareDetection`) visible in pod memory accounting — the kernel already aggregates every container in the pod at that level. Additive: the existing `node_agent.process.memory.*` container-scoped metrics are unchanged in name, meaning, and scope.

Also fixes two defects in `findCgroupScopeDir`, found while investigating a live `cgroup_bytes < rss_bytes` anomaly:
- Container-ID matching was an unanchored substring (`strings.Contains`) — now a delimited-segment match, so a longer container ID that happens to contain ours as a substring can no longer be selected.
- The first name-matching `.scope` directory was accepted without verifying it actually contained `memory.current`. On cgroup-v1/hybrid hosts, the directory walk can reach an unrelated controller subtree (`blkio`, `cpu`, ...) before `memory`, silently locking onto a scope with no `memory.current` and caching a `0` read for the process lifetime (`sync.Once`). The resolver now keeps walking until it finds a real match.

Scoped to the **systemd** cgroup driver — `findCgroupScopeDir` requires a `.scope`-suffixed name, which the cgroupfs driver never produces (`kubepods/burstable/pod<uid>/<id>`, no suffix). Hosts on cgroupfs already read `0` today; this PR doesn't change that. Tracked as a follow-up.

## Additional Information

**Per-pod production verification** (SigNoz, prod-eu, 690 live pods, avg over the prior 15 min at investigation time):
- **241/690 (35%)** report `cgroup_bytes == 0` while `rss_bytes > 0` — the signature of the second defect above. ~92% of these are on standard cloud-managed instance types (AWS/Azure node pools), which default to the systemd cgroup driver, so this fix should resolve the large majority of them.
- **13/690 (2%)** show a genuine per-pod inversion (`0 < cgroup_bytes < rss_bytes`), a smaller, more modest gap (45–180MB) consistent with legitimate RSS-vs-cgroup accounting differences (e.g. page cache) rather than a resolution bug — noted honestly as not fully attributed to either defect.

No live cluster currently runs with `malwareDetection` enabled, so multi-container pod-level aggregation is verified via synthetic `t.TempDir()` cgroup fixtures rather than a real sidecar; single-container behavior (today's actual topology) is unchanged and covered by the existing + new tests.

This PR went through a 3-round Planner → Architect → Critic consensus review before implementation (two real defects were caught and fixed during that process: a pod-slice name-matching regex that rejected the guaranteed-QoS shape it claimed to accept, and a cgroup v1/v2 filename mismatch in the `memory.current`-existence check) — happy to share that trail if useful for review.

## How to Test

```
go test ./pkg/metricsmanager/otel/... -v -count=1
go vet ./...
```

Manual: compare the new `node_agent_pod_memory_cgroup_bytes` (Prometheus bridge name) against `kubectl top pod` for a real pod.

- [x] I have performed a self-review of my code (code-reviewer + security-reviewer agent passes, no blockers)

AI-skills: armosec-shared-rules:environment-links,oh-my-claudecode:plan,oh-my-claudecode:autopilot | cmds: /oh-my-claudecode:deep-dive

## Summary by CodeRabbit

* **New Features**
  * Added pod-level memory usage and memory limit metrics, including sidecar workloads.
  * Improved discovery of pod and container memory scopes across cgroup versions and deployment configurations.

* **Bug Fixes**
  * Prevented incorrect scope matches caused by partial identifiers or unavailable memory data.
  * Added validation to ensure pod metrics are associated with the correct pod.
  * Improved handling of unavailable memory limits and unresolved scope lookups.

* **Documentation**
  * Added migration guidance for pod memory metrics and scope-resolution behavior.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -201,7 +201,10 @@ func main() {
 	// when OTEL_METRICS_EXPORTER=prometheus; OTLP push when OTEL_EXPORTER_OTLP_ENDPOINT set).
 	// MUST be constructed after otelsetup.InitProviders() so the global MeterProvider is set.
 	// Always use the OTEL impl — the SDK's own no-op providers handle the "no endpoint" case.
-	metricsProvider := otelmetrics.NewOTELMetricsManager(resolveOwnContainerID(ctx, k8sClient))
+	ownContainerID, ownPodUID := resolveOwnContainerID(ctx, k8sClient)
+	// true: this entrypoint (the Kubernetes DaemonSet) bind-mounts the HOST's
+	// /sys/fs/cgroup over its own — see NewOTELMetricsManager's doc comment.
+	metricsProvider := otelmetrics.NewOTELMetricsManager(ownContainerID, ownPodUID, true)
 
 	// Create watchers
 	dWatcher := dynamicwatcher.NewWatchHandler(k8sClient, storageClient.GetStorageClient(), cfg.SkipNamespace)
@@ -546,11 +549,13 @@ func resolveOwnContainerID(ctx context.Context, k8sClient *k8sinterface.Kubernet
 	const containerName = "node-agent"
 	podName, namespace := os.Getenv("POD_NAME"), os.Getenv("NAMESPACE_NAME")
 	if podName == "" || namespace == "" {
+		logger.L().Warning("resolveOwnContainerID - POD_NAME or NAMESPACE_NAME unset, cgroup memory gauges will report 0",
+			helpers.String("pod", podName), helpers.String("namespace", namespace))
 		return "", ""
 	}
 	pod, err := k8sClient.GetKubernetesClient().CoreV1().Pods(namespace).Get(ctx, podName, metav1.GetOptions{})
 	if err != nil {
-		logger.L().Warning("resolveOwnContainerID - failed to get own pod, cgroup memory gauges will fall back",
+		logger.L().Warning("resolveOwnContainerID - failed to get own pod, cgroup memory gauges will report 0",
 			helpers.Error(err), helpers.String("pod", podName), helpers.String("namespace", namespace))
 		return "", ""
 	}
@@ -563,5 +568,11 @@ func resolveOwnContainerID(ctx context.Context, k8sClient *k8sinterface.Kubernet
 			return cs.ContainerID, string(pod.UID)
 		}
 	}
+	// No ContainerStatuses entry named "node-agent" yet — most likely a
+	// startup race (this pod's own status hasn't propagated yet). This result
+	// is cached for the process lifetime by the caller, so it's worth logging:
+	// otherwise it's silently indistinguishable from a genuine misconfiguration.
+	logger.L().Warning("resolveOwnContainerID - no ContainerStatuses entry named \"node-agent\" yet, cgroup memory gauges will report 0",
+		helpers.String("pod", podName), helpers.String("namespace", namespace))
 	return "", ""
 }
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -535,31 +535,33 @@ func main() {
 	// Return normally so deferred OTEL shutdown flushes traces/metrics/logs.
 }
 
-// resolveOwnContainerID returns node-agent's own container ID via the k8s API
-// (using the Downward-API POD_NAME / NAMESPACE_NAME env), or "" if it cannot be
-// determined. The cgroup memory gauges use it to locate the correct container
-// scope in the host-mounted cgroup tree; "" makes them fall back to /proc-based
-// resolution. Best-effort: a failure here only degrades those gauges.
-func resolveOwnContainerID(ctx context.Context, k8sClient *k8sinterface.KubernetesApi) string {
+// resolveOwnContainerID returns node-agent's own container ID and pod UID via
+// the k8s API (using the Downward-API POD_NAME / NAMESPACE_NAME env), or "", ""
+// if they cannot be determined. The cgroup memory gauges use the container ID
+// to locate the correct container scope in the host-mounted cgroup tree ("" makes
+// them fall back to /proc-based resolution) and the pod UID to verify the parent
+// pod slice the pod-level gauges read from. Best-effort: a failure here only
+// degrades those gauges.
+func resolveOwnContainerID(ctx context.Context, k8sClient *k8sinterface.KubernetesApi) (containerID, podUID string) {
 	const containerName = "node-agent"
 	podName, namespace := os.Getenv("POD_NAME"), os.Getenv("NAMESPACE_NAME")
 	if podName == "" || namespace == "" {
-		return ""
+		return "", ""
 	}
 	pod, err := k8sClient.GetKubernetesClient().CoreV1().Pods(namespace).Get(ctx, podName, metav1.GetOptions{})
 	if err != nil {
 		logger.L().Warning("resolveOwnContainerID - failed to get own pod, cgroup memory gauges will fall back",
 			helpers.Error(err), helpers.String("pod", podName), helpers.String("namespace", namespace))
-		return ""
+		return "", ""
 	}
 	for _, cs := range pod.Status.ContainerStatuses {
 		if cs.Name == containerName {
 			// ContainerID is "<runtime>://<id>", e.g. "containerd://9103ee5f..."
 			if i := strings.LastIndex(cs.ContainerID, "/"); i >= 0 {
-				return cs.ContainerID[i+1:]
+				return cs.ContainerID[i+1:], string(pod.UID)
 			}
-			return cs.ContainerID
+			return cs.ContainerID, string(pod.UID)
 		}
 	}
-	return ""
+	return "", ""
 }
--- a/pkg/metricsmanager/otel/resource_metrics.go
+++ b/pkg/metricsmanager/otel/resource_metrics.go
@@ -6,12 +6,15 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
+	"regexp"
 	"runtime"
 	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
 
+	"github.com/kubescape/go-logger"
+	"github.com/kubescape/go-logger/helpers"
 	"go.opentelemetry.io/otel/metric"
 )
 
@@ -20,10 +23,14 @@ const cgroupRoot = "/sys/fs/cgroup"
 // registerResourceMetrics wires up observable gauges for process-level and
 // host-level resource. Called once from NewOTELMetricsManager; panics on
 // instrument creation failure (same policy as mustCounter/mustGauge).
-func registerResourceMetrics(meter metric.Meter, containerCount *atomic.Int64, ownContainerID string) {
+func registerResourceMetrics(meter metric.Meter, containerCount *atomic.Int64, ownContainerID, ownPodUID string) {
 	// Per-process memory gauges (rss + cgroup usage/limit) — shared with the
 	// sbom-scanner sidecar so both containers report the same memory signals.
 	RegisterProcessMemoryMetrics(meter, ownContainerID)
+
+	// Pod-scoped memory gauges — registered only here (the main agent), never
+	// from the sidecar-shared RegisterProcessMemoryMetrics.
+	RegisterPodMemoryMetrics(meter, ownContainerID, ownPodUID)
 
 	hostMemTotal := readHostMemTotalBytes()
 	hostCPUCount := int64(runtime.NumCPU())
@@ -98,6 +105,42 @@ func RegisterProcessMemoryMetrics(meter metric.Meter, ownContainerID string) {
 	}, rssGauge, cgroupMemGauge, cgroupLimitGauge)
 }
 
+// RegisterPodMemoryMetrics registers pod-scoped memory gauges covering every
+// container in this pod, including sidecars with no instrumentation of their
+// own. Called only from registerResourceMetrics (the main agent), never from
+// the shared RegisterProcessMemoryMetrics — registering it per-container would
+// double-count the pod under any aggregation.
+//
+// ownContainerID == "" (sbom-scanner sidecar, cmd/host, cmd/ecs) registers
+// nothing at all, so no empty series is ever emitted from those topologies.
+func RegisterPodMemoryMetrics(meter metric.Meter, ownContainerID, ownPodUID string) {
+	if ownContainerID == "" {
+		return
+	}
+
+	podMemGauge, err := meter.Int64ObservableGauge("node_agent.pod.memory.cgroup_bytes",
+		metric.WithDescription("Pod memory usage from the parent pod cgroup slice (kubepods-...-pod<UID>.slice) memory.current — covers every container in the pod, including uninstrumented sidecars"),
+		metric.WithUnit("By"),
+	)
+	if err != nil {
+		panic("otelmetrics: gauge node_agent.pod.memory.cgroup_bytes: " + err.Error())
+	}
+	podLimitGauge, err := meter.Int64ObservableGauge("node_agent.pod.memory.cgroup_limit_bytes",
+		metric.WithDescription("Pod memory limit from the parent pod cgroup slice memory.max (0 = unlimited or unresolved). Pair with node_agent.pod.memory.cgroup_bytes for OOM headroom."),
+		metric.WithUnit("By"),
+	)
+	if err != nil {
+		panic("otelmetrics: gauge node_agent.pod.memory.cgroup_limit_bytes: " + err.Error())
+	}
+
+	_, _ = meter.RegisterCallback(func(_ context.Context, o metric.Observer) error {
+		cur, lim := readPodCgroupMem(ownContainerID, ownPodUID)
+		o.ObserveInt64(podMemGauge, cur)
+		o.ObserveInt64(podLimitGauge, lim)
+		return nil
+	}, podMemGauge, podLimitGauge)
+}
+
 func readProcessRSSBytes() int64 {
 	f, err := os.Open("/proc/self/status")
 	if err != nil {
@@ -122,6 +165,7 @@ var (
 	cgroupResolveOnce sync.Once
 	cgroupCurrentPath string // path to memory.current / memory.usage_in_bytes ("" if unresolved)
 	cgroupMaxPath     string // path to memory.max / memory.limit_in_bytes ("" if unresolved)
+	cgroupScopeDir    string // this container's ".scope" dir under cgroupRoot ("" if not found that way); reused by the pod-level resolver
 )
 
 // readCgroupMem returns the process's cgroup memory usage and limit in bytes
@@ -166,6 +210,7 @@ func readCgroupMem(ownContainerID string) (current, limit int64) {
 func resolveCgroupMemoryPaths(ownContainerID string) (current, max string) {
 	if ownContainerID != "" {
 		if dir := findCgroupScopeDir(cgroupRoot, ownContainerID); dir != "" {
+			cgroupScopeDir = dir // published under cgroupResolveOnce for cachedCgroupScopeDir
 			return filepath.Join(dir, "memory.current"), filepath.Join(dir, "memory.max")
 		}
 	}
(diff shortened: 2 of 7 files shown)


pkg/metricsmanager/otel/resource_metrics.go
```diff
@@ -196,20 +362,73 @@ func parseSelfCgroupV2(content string) (string, bool) {
...
+// name-only match: that is what lets the walk skip the decoy controller
+// subtrees and reach the real memory-controller scope.
+//
+// The check deliberately accepts the cgroupv2 filename memory.current *only*,
+// never the cgroupv1 memory.usage_in_bytes: resolveCgroupMemoryPaths returns
+// the v2 filename pair unconditionally once this function succeeds, so
+// accepting a v1 directory would hand back paths that do not exist and
+// reproduce H4 verbatim. Returning "" on a v1 host is strictly better — it lets
+// resolveCgroupMemoryPaths fall through to its later strategies, including the
+// cgroupv1 fixed-mount fallback, instead of caching a dead v2 path.
```
matthyx: **BLOCKER — on cgroup-v1/hybrid hosts this change turns a silent `0` into a *node-wide* number.**

This comment argues that returning `""` on a v1 host is "strictly better ... it lets `resolveCgroupMemoryPaths` fall through to its later strategies, including the cgroupv1 fixed-mount fallback". In node-agent's own topology that fallback is not this container's cgroup:

```go
// resolveCgroupMemoryPaths, strategy 3
if fileExists("/sys/fs/cgroup/memory/memory.usage_in_bytes") {
    return "/sys/fs/cgroup/memory/memory.usage_in_bytes", "/sys/fs/cgroup/memory/memory.limit_in_bytes"
}
```

`resolveCgroupMemoryPaths`' own doc comment states node-agent "bind-mount[s] the host's /sys/fs/cgroup over its own". So `/sys/fs/cgroup/memory/` is the **host's root memory cgroup** — node-wide usage — and `memory.limit_in_bytes` there is the unlimited sentinel `9223372036854771712`.

Walk a pure cgroup-v1 host with the new code:

1. `findCgroupScopeDir` visits `/sys/fs/cgroup/{blkio,cpu,...,memory}/.../cri-containerd-<id>.scope`. **None** of them holds `memory.current` — the v1 memory-controller file is `memory.usage_in_bytes`, which this check deliberately rejects — so every candidate is skipped and it returns `""`.
2. Strategy 2 (`0::` line of `/proc/self/cgroup`) is absent on pure v1, and on hybrid the unified hierarchy carries no memory controller → no match.
3. Strategy 3 fires → node-wide usage + sentinel limit, cached for the whole process lifetime by `cgroupResolveOnce`.

Before this PR the same host name-matched a decoy scope, got nonexistent v2 paths, and read `0`. After it, `node_agent_process_memory_cgroup_bytes` reports the entire node and `node_agent_process_memory_cgroup_limit_bytes` reports ~9.2e18. That is strictly worse than the `0` it replaces: a wrong-but-plausible value silently breaks OOM-headroom dashboards and any absolute-threshold alert on this series, whereas `0` is visibly broken.

Note this is the fixed-mount fallback that already existed — the regression is that this PR newly *routes v1/hybrid main-agent hosts into it*, i.e. exactly the hosts the PR targets.

Either fix works:

- Accept `memory.usage_in_bytes` in `findCgroupScopeDir` too, and have `resolveCgroupMemoryPaths` return the v1 filename pair when the resolved scope dir is v1. This actually fixes v1 rather than skipping it.
- Or gate strategy 3 on `ownContainerID == ""`, so the fixed-mount fallback only serves the own-cgroup-namespace (sidecar) topology it was written for.

Minor, same function: dropping `SkipAll` on a name-only match means that when nothing matches, the walk now traverses the **entire** host cgroup tree (tens of thousands of dirs on a busy node) inside an OTEL collection callback. One-time, but it stalls that collection cycle — worth resolving off the callback path or bounding the walk depth.

pkg/metricsmanager/otel/resource_metrics.go
```diff
@@ -98,6 +105,42 @@ func RegisterProcessMemoryMetrics(meter metric.Meter, ownContainerID string) {
...
+// RegisterPodMemoryMetrics registers pod-scoped memory gauges covering every
+// container in this pod, including sidecars with no instrumentation of their
+// own. Called only from registerResourceMetrics (the main agent), never from
+// the shared RegisterProcessMemoryMetrics — registering it per-container would
+// double-count the pod under any aggregation.
+//
+// ownContainerID == "" (sbom-scanner sidecar, cmd/host, cmd/ecs) registers
+// nothing at all, so no empty series is ever emitted from those topologies.
+func RegisterPodMemoryMetrics(meter metric.Meter, ownContainerID, ownPodUID string) {
+	if ownContainerID == "" {
```
matthyx, line 132: **BLOCKER — the 35% of pods this PR is justified by are likely the exact pods that will emit nothing here.**

The H4 root cause requires per-controller subdirectories under `/sys/fs/cgroup` (`blkio`, `cpu`, ... sorting before `memory`). Those only exist on cgroup **v1/hybrid** hosts. On a pure cgroup-v2 unified host there is no `blkio`/`cpu` subtree at all, so the described walk-order failure cannot occur there.

The PR description attributes 92% of the 241 zero-reporting pods to AWS/Azure managed node pools — but those (EKS AL2023, AKS Ubuntu 22.04+) are overwhelmingly cgroup-v2 unified, where H4 is impossible. Before merging on the stated justification, please confirm the actual v1/v2 split of those 241 pods; "systemd cgroup driver" is not the same claim as "cgroup v1".

There is a much simpler v2 explanation for `cgroup_bytes == 0 && rss_bytes > 0`: `ownContainerID == ""`. That happens when `POD_NAME`/`NAMESPACE_NAME` are unset, the pod `GET` fails, or no container status is named `node-agent`. Then strategy 1 is skipped entirely; strategy 2 resolves to the cgroup-v2 **root**, which exposes no memory-controller interface files (no `memory.current`); strategy 3 finds no `/sys/fs/cgroup/memory` on unified — result `0`, cached forever.

If that is the real cause, this early return means those same pods register **no pod gauge at all** — no series, not even a `0` — and nothing is logged, since `ownContainerID == ""` is deliberately unlogged both here and in `resolvePodCgroupMemoryPathsUnder`. The metric would then be missing precisely where it is needed, with no signal that it is missing.

Ask: emit a one-time `Warning` from the main-agent path when `ownContainerID == ""` (the "by design" silence is only correct for the sidecar / `cmd/host` / `cmd/ecs` topologies, and this constructor cannot tell them apart), and confirm the cgroup version breakdown of the 241 pods.

pkg/metricsmanager/otel/resource_metrics_test.go
```diff
@@ -65,6 +69,270 @@ func TestFindCgroupScopeDir(t *testing.T) {
...
+	parent := filepath.Join(root, parentRel)
+	require.NoError(t, os.MkdirAll(parent, 0o755))
+	for name, content := range parentFiles {
+		require.NoError(t, os.WriteFile(filepath.Join(parent, name), []byte(content), 0o644))
+	}
+	for i, id := range containerIDs {
+		scope := filepath.Join(parent, "cri-containerd-"+id+".scope")
+		require.NoError(t, os.MkdirAll(scope, 0o755))
+		require.NoError(t, os.WriteFile(filepath.Join(scope, "memory.current"),
+			[]byte(strconv.Itoa(100000*(i+2))+"\n"), 0o644))
```
matthyx, line 313: The three child scopes get `200000` / `300000` / `400000`, which sum to exactly `900000` — the pod slice's own `memory.current` in `podFiles`.

`TestResolvePodCgroupMemoryPaths`'s doc comment claims the resolver reads "a value distinct from every child scope **and from their sum**", but an implementation that summed the children would pass the `three child scopes` subtest identically. The property the test is meant to pin is not actually pinned.

Make the sum differ, e.g. `strconv.Itoa(100000*(i+2)+1)`.

(Same finding as CodeRabbit's — co-signing because it defeats the stated purpose of the test rather than being a style nit.)

pkg/metricsmanager/otel/resource_metrics.go
```diff
@@ -184,6 +229,127 @@ func resolveCgroupMemoryPaths(ownContainerID string) (current, max string) {
...
+// segments a custom kubelet --cgroup-root introduces
+// ("kubelet-kubepods-burstable-pod<uid>.slice"). The QoS and root slices
+// ("kubepods.slice", "kubepods-burstable.slice", "kubepods-besteffort.slice",
+// "system.slice") are still rejected by shape since none of them ends in
+// "pod<uid>.slice". Loosening the prefix is safe because the mandatory pod-UID
+// equality check below — not the name shape — is the real correctness guard. A
+// pod UID is 32 hex digits plus 4 separators, which systemd escapes "-" → "_",
+// hence [0-9a-f_]{32,}; that escaping is also why the "-"-delimited prefix
+// segments can never eat into the UID capture.
+var podSliceNameRe = regexp.MustCompile(`^(?:[a-z0-9]+-)*pod([0-9a-f_]{32,})\.slice$`)
```
matthyx, line 317: Minor: the prefix class `[a-z0-9]+-` excludes `_`, but systemd escaping is precisely what turns `-` into `_`.

A kubelet started with `--cgroup-root=/my-root` yields `my_root-kubepods-burstable-pod<uid>.slice`, which this regex rejects → pod memory metrics silently report `0` on those clusters (with only a `Debug` line to explain it).

Since the comment correctly identifies the pod-UID equality check below as the real correctness guard, loosening the prefix to `(?:[a-z0-9_]+-)*` costs nothing and removes the failure mode.

matthyx (commented):
Reviewed the full diff plus the surrounding `resolveCgroupMemoryPaths` strategy chain. The pod-level metric itself is well-scoped and the substring→delimited-segment container-ID fix is a clear improvement. Two blockers, both in the "hardening" half rather than the new metric:

1. **cgroup-v1/hybrid regression** (`resource_metrics.go`, `findCgroupScopeDir`) — requiring `memory.current` makes v1 hosts fall through to the `/sys/fs/cgroup/memory/*` fixed-mount fallback, which under node-agent's host-bind-mounted cgroup tree is the **node-wide** root cgroup. `node_agent_process_memory_cgroup_bytes` goes from `0` to the whole node's usage, and the limit to the `9223372036854771712` sentinel — cached for the process lifetime. Wrong-but-plausible is worse than the `0` it replaces.
2. **Root cause vs. evidence** (`RegisterPodMemoryMetrics`) — H4 is only reachable on v1/hybrid, but the cited 92% of affected pods are on managed node pools that are predominantly cgroup-v2 unified, where the described failure mode cannot happen. A simpler v2 explanation for the 35% is `ownContainerID == ""`, and in that case the new pod gauge registers *nothing at all*, silently. Please confirm the v1/v2 split of the 241 pods before merging on the stated justification.

Non-blocking, also worth addressing:

3. Test fixture in `resource_metrics_test.go` doesn't pin the property it documents (child values sum to the pod value).
4. `podSliceNameRe` prefix class rejects the `_` that systemd escaping produces for a custom `--cgroup-root`.
5. `findCgroupScopeDir` no longer short-circuits on a name-only match, so a no-match host now walks the entire cgroup tree inside an OTEL collection callback.
6. The `docs/metrics-migration.md` claim that the two defects account for the 35% inherits blocker 2 — please soften it to what the telemetry actually supports until the cgroup-version split is confirmed.
7. CI: **DCO is failing** — commit `aacd803` has no `Signed-off-by:` trailer. `git commit --amend -s && git push --force-with-lease` will clear it.

Agreed with CodeRabbit on raising the `reject` log from `Debug` to `Warning` — it resolves once per process, so there's no volume concern, and it is the only signal that this silently degraded.

matthyx:
Both blockers addressed in cbf2661d:

1. **cgroup-v1 node-wide fallback**: confirmed — `resolveCgroupMemoryPaths` did fall through to the unscoped strategies whenever the scoped lookup failed, regardless of whether `ownContainerID` was known. Fixed: those strategies now only run when `ownContainerID == ""` (the only topology where the cgroup root itself is a valid proxy for "this container's memory" — the sbom-scanner sidecar's own namespaced root). A known container ID that can't be scope-resolved now returns `"", ""`, never a plausible-looking wrong number. Went further than a minimal revert: `findCgroupScopeDir` now also accepts a verified v1 directory (`memory.usage_in_bytes`) alongside v2, so v1 hosts get correctly *container-scoped* real numbers instead of a bare `0` — plus a sentinel-value fix so a v1 "unlimited" limit reports as `0` instead of a huge misleading number.

2. **Root-cause attribution**: you're right that I can't confirm the v1/v2 split from SigNoz alone — no cgroup-version attribute exists in current telemetry, and I don't have live per-node access from here to check kubelet cgroup driver directly. Softened `docs/metrics-migration.md` to say what's actually confirmed (35% of pods report `cgroup_bytes == 0` while `rss_bytes > 0`) versus what's a hypothesis (that the v1/hybrid defect specifically explains it — an empty-`ownContainerID` startup race is equally consistent with the same data). This PR now fixes both candidate mechanisms regardless of which dominates, and — since v1 hosts will report real numbers post-deploy instead of a silent `0` — the actual split becomes observable rather than asserted.

New pinning tests: `TestResolveCgroupMemoryPathsUnder` (5 subtests, including the specific regression case: known container ID + no verifiable scope + a v1 fixed-mount file present at root → must return `"", ""`, not read it) and `TestParseCgroupMemValue_V1UnlimitedSentinel`. `go build`, `go test ./pkg/metricsmanager/otel/... -v -count=1`, `go vet ./...` all clean.

Also applied the CodeRabbit nitpick you agreed with (Debug → Warning on the pod-level rejection log).

Not yet addressed (non-blocking, from CodeRabbit): the `resolveOwnContainerID` unbounded-context timeout, the test-fixture child-sum coincidence, and the full-tree-walk-on-no-match perf note — happy to take those in a follow-up if you'd like them in this PR instead.

pkg/metricsmanager/otel/resource_metrics.go
```diff
@@ -150,40 +194,196 @@ func readCgroupMem(ownContainerID string) (current, limit int64) {
...
 //     path) and a container's own namespaced /sys/fs/cgroup mount, where
 //     /proc/self/cgroup is "0::/" and the namespace root (cgroupRoot itself) is
-//     the container's own cgroup — the sbom-scanner sidecar topology. The main
-//     agent overrides /sys/fs/cgroup with the host tree, whose root has no
-//     memory.current, so this path no-ops there and strategy 1 wins.
+//     the container's own cgroup — the sbom-scanner sidecar topology.
+//
 //  3. cgroupv1 fixed mount layout.
+//
+// Strategies 2 and 3 only run when ownContainerID == "" — i.e. the caller
```
matthyx: **Residual blocker — the node-wide read is still reachable, through the `ownContainerID == ""` door instead of the "scope not found" door.**

This gate treats two different topologies as one: *"the caller either doesn't know which container it is, **or** mounts its own namespaced cgroup root"*. Only the second makes strategies 2/3 sound. The main agent bind-mounts the **host's** cgroup tree and can still land in `ownContainerID == ""` — and when it does, strategy 3 reads `<root>/memory/memory.usage_in_bytes`, which is the host's root memory cgroup.

`resolveOwnContainerID` in `cmd/main.go` has three ways to return `""`, and two are silent:

```go
if podName == "" || namespace == "" { return "", "" }          // silent
... Get(...); if err != nil { logger.L().Warning(...); ... }    // logged
for _, cs := range pod.Status.ContainerStatuses { ... }
return "", ""                                                   // silent — no status named "node-agent" yet
```

That last one is exactly the *"`ownContainerID` resolving to empty on a permanently-cached early-startup race"* this commit's own message names as the leading alternative hypothesis for the 35%. So the path most likely to fire in production is the one still uncovered.

Confirmed against this branch (`cbf2661d`) with a throwaway test in the package — a v1/hybrid host root, main-agent topology, private cgroup namespace:

```go
root := t.TempDir()  // stands in for the bind-mounted HOST tree
os.MkdirAll(filepath.Join(root, "memory"), 0o755)
os.WriteFile(filepath.Join(root, "memory", "memory.usage_in_bytes"), []byte("8589934592
"), 0o644) // 8 GiB = whole node

cur, max := resolveCgroupMemoryPathsUnder(root, "", "0::/\n")
// cur = <root>/memory/memory.usage_in_bytes
// parseCgroupMemValue(readFileString(cur)) == 8589934592
```

It passes — i.e. the container reports the whole node's 8 GiB as its own. `TestResolveCgroupMemoryPathsUnder/empty container ID falls through to the v1 fixed-mount strategy` already pins this as intended behavior.

To be fair: this is **pre-existing**, not introduced by this PR — but it's the same failure mode the fix commit claims to have closed, and `docs/metrics-migration.md` now asserts *"a container ID that can't be scope-resolved now reports `0`, never a plausible-looking wrong number."* That's true for "found no scope" and false for "never had an ID".

Suggested fix: stop inferring topology from `ownContainerID == ""` and pass it explicitly — `cmd/main.go` already knows it is the host-tree caller, and the sidecar / `cmd/host` / `cmd/ecs` entrypoints know they mount their own namespaced root. Then strategies 2/3 run only for the topology they were written for, regardless of whether the container ID resolved.

pkg/metricsmanager/otel/resource_metrics.go
matthyx, line 132: **The observability half of the earlier blocker isn't closed yet.**

The `Debug` → `Warning` change only covers the `reject` path — but `reject` is unreachable when `ownContainerID == ""`, because `resolvePodCgroupMemoryPathsUnder` returns at its own `ownContainerID == ""` guard before `reject` is ever consulted.

So for the empty-container-ID case the result is total silence: no pod gauge registered here, no log there, and two of the three `""` returns in `resolveOwnContainerID` don't log either. `docs/metrics-migration.md` now promises that *"the actual v1-vs-empty-container-ID split will become directly observable post-deploy rather than inferred"* — but the empty-container-ID half emits no signal at all, so the split still can't be read off the fleet.

Two one-line `Warning`s would close it and make the doc's claim true:
- in `resolveOwnContainerID`, on the `POD_NAME`/`NAMESPACE_NAME`-unset and no-matching-container-status returns;
- here, before the early return — this constructor is only called from `registerResourceMetrics` (main agent), so `ownContainerID == ""` here is always the anomalous case, never the by-design sidecar one.

pkg/metricsmanager/otel/resource_metrics.go
```diff
@@ -150,40 +194,196 @@ func readCgroupMem(ownContainerID string) (current, limit int64) {
...
+		return reject("parent slice name is not a pod slice: " + filepath.Base(parent))
+	}
+	// Mandatory verification against the pod UID when it is known; the regex
+	// already constrains the captured segment to [0-9a-f_], so no case folding
+	// is involved. An unknown UID falls back to the name-shape guard alone.
+	if ownPodUID != "" && m[1] != strings.ReplaceAll(ownPodUID, "-", "_") {
+		return reject("parent slice belongs to a different pod UID: " + filepath.Base(parent))
+	}
+	if !fileExists(filepath.Join(parent, "memory.current")) {
+		return reject("parent slice has no memory.current (cgroup v1?): " + parent)
```
matthyx: Minor, new inconsistency introduced by the v1 support: `findCgroupScopeDir` and `resolveCgroupMemoryPaths` now handle cgroup v1, but pod-level resolution still hard-requires `memory.current`, so v1 hosts get correct **container** metrics and a permanent `0` for **pod** metrics.

Verified on this branch — a v1 pod slice with `memory.usage_in_bytes`/`memory.limit_in_bytes` and a v1 child scope:

```
container level: <podSlice>/cri-containerd-<id>.scope/memory.usage_in_bytes   ✅ resolves
pod level:       "" , ""                                                       ❌ rejected
[warning] otelmetrics: pod-level cgroup memory unresolved ... rejected: parent slice has no memory.current (cgroup v1?)
```

It's now `Warning`-logged, so it is at least visible — no objection to shipping it as-is. But `docs/metrics-migration.md` currently lists only the cgroupfs driver as the pod-level gap; cgroup v1 is now a second one and deserves a line there (the same `hasV2 || hasV1` + filename-pair pattern used for the container level would fix it outright).

pkg/metricsmanager/otel/resource_metrics.go
```diff
@@ -150,40 +194,196 @@ func readCgroupMem(ownContainerID string) (current, limit int64) {
...
+// read of /proc/self/cgroup, so tests can exercise every strategy against a
+// synthetic tree instead of the real host filesystem.
+func resolveCgroupMemoryPathsUnder(root, ownContainerID, selfCgroupContent string) (current, max string) {
 	if ownContainerID != "" {
-		if dir := findCgroupScopeDir(cgroupRoot, ownContainerID); dir != "" {
+		dir := findCgroupScopeDir(root, ownContainerID)
+		if dir == "" {
+			return "", ""
+		}
+		cgroupScopeDir = dir // published under cgroupResolveOnce for cachedCgroupScopeDir
```
matthyx: Nit: this writes the package-level `cgroupScopeDir` from the testable core, and `TestResolveCgroupMemoryPathsUnder` now calls that core directly with `t.TempDir()` roots — so a temp path that no longer exists is left in the global after those subtests.

Nothing currently reads it in a way that breaks (the pod tests go through `cachedCgroupScopeDir` with a non-`cgroupRoot` root, which walks directly), so this is latent rather than live. Moving the assignment up into the `resolveCgroupMemoryPaths` wrapper keeps the global tied to the real-root path it documents.

matthyx (commented):
Re-reviewed at `cbf2661d`. The main blocker is genuinely fixed, and the v1 support goes further than I asked — nice.

**Verified fixed:**
- A known container ID that can't be scope-resolved now returns `""`, `""` instead of falling through. `TestResolveCgroupMemoryPathsUnder/known container ID, no scope found anywhere...` pins it with a node-wide file deliberately present in the fixture, so it fails the old way. Good test.
- cgroup v1 is now actually supported rather than skipped — `findCgroupScopeDir` accepts either filename and `resolveCgroupMemoryPaths` reads whichever pair is present, so v1 hosts get real container-scoped numbers instead of `0`. This is a better outcome than either option I suggested.
- `v1UnlimitedThreshold` at `1<<62` is comfortably below the `9223372036854771712` sentinel and far above any real limit — correct, and pinned.
- Pod-resolver log raised to `Warning`.
- `docs/metrics-migration.md` honestly walks back the causal attribution of the 35% instead of defending it. That's the right call.
- DCO now passes; both commits signed off.

Ran locally on this branch: `go test ./pkg/metricsmanager/otel/... -count=1` passes, `go vet` clean. CI is 28 pass / 1 fail — `component-tests (Test_13_MergingNetworkNeighborhoodTest)`, which is nowhere near this change and looks unrelated, worth one re-run to confirm.

**One residual blocker**, narrower than the original: the same node-wide misattribution is still reachable when `ownContainerID == ""` on a v1/hybrid host. The new gate infers topology from "do we know our container ID", but the main agent can fail to resolve one (silently, in two of three paths) while still having the *host* tree mounted — and the commit message itself names that empty-container-ID race as the leading hypothesis for the 35%. Reproduced on this branch; details inline.

Also inline: the observability half of the earlier ask isn't closed (the `Warning` is unreachable for exactly the empty-container-ID case the docs promise to make observable), plus a new v1 gap at the pod level and a test-pollution nit.

Still open from the first pass, both non-blocking and both fine to defer: the pod-tree fixture's child values still sum to exactly `900000`, and `podSliceNameRe`'s `[a-z0-9]+-` still rejects the `_` systemd escaping produces.

matthyx:
All four points from this round addressed in 4ec9a4c8:

**Residual blocker — confirmed and fixed.** You're right that the gate was inferring topology from `ownContainerID == ""`, which conflates "don't know my container ID" with "my own root is my own cgroup" — only the second makes the unscoped strategies safe, and the main agent can be in the first state while still bind-mounting the host tree. Fixed properly this time: `hostCgroupMounted bool` is now passed explicitly (not inferred) from `NewOTELMetricsManager` down through the whole chain — `cmd/main.go` passes `true`, the sbom-scanner sidecar's direct call passes `false`. When `true`, only the scoped-by-ID lookup ever runs; any failure returns `"", ""` unconditionally, no unscoped door left open. Added your exact PoC as a pinning test (empty container ID + host-mounted + a node-wide v1 file present at root → must stay `"", ""`).

**Observability gap — closed.** Both silent `resolveOwnContainerID` returns (missing `POD_NAME`/`NAMESPACE_NAME`, and no matching `ContainerStatuses` entry) now log `Warning`. Also added a `Warning` in `RegisterPodMemoryMetrics` specifically for the DaemonSet-topology + empty-container-ID combination — scoped to `hostCgroupMounted` so it stays silent for `cmd/host`/`cmd/ecs`, which have no pod concept at all and shouldn't get noise from an by-design no-op.

**Pod-level v1 gap — closed**, not just documented. Applied the same `hasV2 || hasV1` + correct-filename-pair pattern to the pod-level resolver. cgroupfs remains the only declared non-goal now; cgroup v1 is fully supported at both container and pod level. Flipped the existing "cgroupv1 pod slice" test from expecting rejection to expecting correct resolution.

**Test-pollution nit — fixed.** Moved the `cgroupScopeDir` package-var publish out of the testable `..Under` core into the real `resolveCgroupMemoryPaths` wrapper only, so synthetic test roots no longer leak into global state.

`go build ./...`, `go test ./pkg/metricsmanager/otel/... ./pkg/sbomscanner/... -v -count=1`, `go vet ./...` all clean. DCO signed off.

Still open, both non-blocking per your last review: the pod-tree fixture's child-sum coincidence, and `podSliceNameRe`'s `_` rejection under custom `--cgroup-root`. Let me know if you want those folded in too, or if this is good to land as-is.

cmd/main.go
```diff
@@ -201,7 +201,10 @@ func main() {
 	// when OTEL_METRICS_EXPORTER=prometheus; OTLP push when OTEL_EXPORTER_OTLP_ENDPOINT set).
 	// MUST be constructed after otelsetup.InitProviders() so the global MeterProvider is set.
 	// Always use the OTEL impl — the SDK's own no-op providers handle the "no endpoint" case.
-	metricsProvider := otelmetrics.NewOTELMetricsManager(resolveOwnContainerID(ctx, k8sClient))
+	ownContainerID, ownPodUID := resolveOwnContainerID(ctx, k8sClient)
```
matthyx, line 204: Non-blocking, but the most valuable remaining follow-up: this condition is now **visible**, yet still **permanent**.

`resolveOwnContainerID` runs exactly once, at startup, and `cmd/main.go` hands the result straight into the constructor, where it's captured in the gauge callbacks for the process lifetime. So a transient startup race — this pod's own `ContainerStatuses` entry not having propagated to the API server yet, which is precisely what this branch describes — permanently zeroes **both** the container-level and the pod-level cgroup metrics until the pod restarts. Nothing retries.

That matters because this branch is the leading hypothesis in your own commit message for the 35%. If that hypothesis is right, this PR makes the population *countable* (good — that was the ask) but doesn't shrink it. A pod that logs this Warning reports `cgroup_bytes == 0` forever.

Cheap ways to close it, roughly in order of effort:
- resolve lazily on the first gauge callback instead of at construction — by then kubelet has certainly published the status, and the existing `sync.Once` caching still applies;
- or keep the startup call and add a short bounded retry (a few attempts over ~5s) before giving up;
- or re-resolve once on the first collection that finds an empty container ID.

Any of these turns "permanently degraded" into "degraded for one scrape interval". Fine as a follow-up issue rather than in this PR — but worth deciding before you read the post-deploy split, since a persistent count here would be self-inflicted rather than informative.

Separately, still open from CodeRabbit's first pass: the `Pods(namespace).Get(ctx, ...)` below uses the unbounded `ctx`. The `context.WithTimeout` added at line 118 is the OTEL shutdown path, not this one. A slow or unreachable API server blocks node-agent startup here.

pkg/metricsmanager/otel/resource_metrics_test.go
```diff
@@ -65,6 +69,393 @@ func TestFindCgroupScopeDir(t *testing.T) {
...
```
matthyx, line 313: Still open from round one (non-blocking, and fine to leave): the three child scopes get `200000` / `300000` / `400000`, which sum to exactly `900000` — the pod slice's own `memory.current` in `podFiles`.

`TestResolvePodCgroupMemoryPaths`' doc comment claims the value read is "distinct from every child scope **and from their sum**", but an implementation that summed the children would pass the `three child scopes` subtest identically. `strconv.Itoa(100000*(i+2)+1)` makes the claim true.

pkg/metricsmanager/otel/resource_metrics.go
```diff
@@ -146,44 +208,217 @@ func readCgroupMem(ownContainerID string) (current, limit int64) {
...
+// segments a custom kubelet --cgroup-root introduces
+// ("kubelet-kubepods-burstable-pod<uid>.slice"). The QoS and root slices
+// ("kubepods.slice", "kubepods-burstable.slice", "kubepods-besteffort.slice",
+// "system.slice") are still rejected by shape since none of them ends in
+// "pod<uid>.slice". Loosening the prefix is safe because the mandatory pod-UID
+// equality check below — not the name shape — is the real correctness guard. A
+// pod UID is 32 hex digits plus 4 separators, which systemd escapes "-" → "_",
+// hence [0-9a-f_]{32,}; that escaping is also why the "-"-delimited prefix
+// segments can never eat into the UID capture.
+var podSliceNameRe = regexp.MustCompile(`^(?:[a-z0-9]+-)*pod([0-9a-f_]{32,})\.slice$`)
```
matthyx, line 317: Still open from round one (non-blocking): `[a-z0-9]+-` excludes `_`, but systemd escaping is exactly what produces `_`. A kubelet run with `--cgroup-root=/my-root` yields `my_root-kubepods-burstable-pod<uid>.slice`, which this rejects — pod metrics then sit at `0` on those clusters, now with a `Warning` to explain it.

Since the pod-UID equality check below is the real correctness guard, `(?:[a-z0-9_]+-)*` costs nothing.

matthyx (commented):
Re-reviewed at `4ec9a4c8`. **All blockers are cleared.** Threading `hostCgroupMounted` explicitly instead of inferring topology from `ownContainerID == ""` is the right fix, and it's threaded consistently through every layer.

Verified on this branch, re-running my round-2 PoC and its neighbours as throwaway package tests:

| scenario | `resolveCgroupMemoryPathsUnder` | |
|---|---|---|
| host tree, empty container ID, v1 root file present (round-2 PoC) | `"", ""` | ✅ no longer node-wide |
| host tree, container ID known, no scope anywhere | `"", ""` | ✅ no fall-through |
| sidecar topology (`hostCgroupMounted=false`) | resolves `memory.current`/`memory.max` | ✅ unaffected |

Both callers are correct: `cmd/main.go` passes `true` (host-mounted DaemonSet), `pkg/sbomscanner/v1/run.go` passes `false` (own namespaced root). Those are the only two call sites in the repo.

Also confirmed fixed from round two:
- **Observability gap closed.** All three `""` returns in `resolveOwnContainerID` now log `Warning`, and `RegisterPodMemoryMetrics` logs when an unresolved container ID reaches the DaemonSet topology specifically — while staying correctly silent for the sidecar / `cmd/host` / `cmd/ecs` topologies that have no pod concept. The empty-container-ID population is now countable, which is what makes the doc's post-deploy-split claim honest.
- **Pod-level cgroup v1 gap closed** — I only flagged this as a docs/follow-up item, but you fixed it outright with the same `memory.current` → `memory.usage_in_bytes` fallback used at the container level. Verified: a v1 pod slice now resolves to `memory.usage_in_bytes`/`memory.limit_in_bytes` instead of being rejected. `cgroupfs` is now the only declared non-goal, and the docs say so.
- The `cgroupScopeDir` global-write nit is moot now that the assignment sits behind the scoped branch.

Local on `4ec9a4c8`: `go build ./...` clean, `go test ./pkg/metricsmanager/otel/... -count=1` passes, `go vet ./pkg/metricsmanager/otel/... ./cmd/...` clean. CI at time of review: 4 pass / 3 pending, no failures — the `Test_13_MergingNetworkNeighborhoodTest` failure from the previous round is gone.

No objection to merging. Three non-blocking follow-ups inline: the startup race is now logged but still permanently cached (the one I'd actually act on, since it's your own leading hypothesis for the 35%), plus the two round-one nits that are still open.

PR #884, opened 2026-08-05
fix(containerprofilemanager): repair timestamp chain on LRU eviction and retry exhaustion

### LINKED ISSUE #871: A MaxAttempts exhaustion drop or LRU eviction can permanently break a container's report chain, pre-existing on main
### Context

Found as a side effect of the review process on #866 (https://github.com/kubescape/node-agent/pull/866#pullrequestreview-4794269159), while verifying that #866's chunk-splitting design correctly preserves storage's time-series report chain. This bug is **independent of #866** — it exists on `main` today and does not appear to have been reported before.

### Problem

Storage reconstructs a container's profile lifecycle by walking a strictly linear chain of `previousReportTimestamp → reportTimestamp` links across a series' `time_series` rows (`consolidateContinuousTimeSeries`, `kubescape/storage@v0.0.290` `pkg/registry/file/containerprofile_processor.go:631-659`). Critically, `updateProfileStatus` only reaches its `Completed`/expired branches when the consolidated chain has exactly one row **and** that row's `previousReportTimestamp` is the zero time — i.e. the chain must trace all the way back to the container's very first report (`containerprofile_processor.go:672`).

Two existing node-agent paths can drop a queued `ContainerProfile` chunk without any replacement:
1. **`MaxAttempts` exhaustion** in `pkg/containerprofilemanager/v1/queue` — after the configured retry budget is exhausted, the chunk is dropped with a `Warning` and no further action.
2. **LRU eviction** in `enforceMaxSize` — evicts from the queue's head (oldest items) while new items are appended to the tail, so an unrelated chunk can be silently evicted while its neighbors survive.

Either path removes one link from a container's report-timestamp chain. Because a missing link is never regenerated (nothing re-derives or re-sends a dropped chunk's timestamps), that container's profile can then **never** complete — it stays in `Learning` forever, indistinguishable from a container that's still legitimately being profiled. This makes the bug essentially unreportable in the field: the symptom looks identical to normal, ongoing learning.

### Fix

Make `enforceMaxSize` and the `MaxAttempts` drop path lineage-aware — e.g. emit the same kind of chain-preserving "stitch chunk" that #866 introduces for its own split-chunk drop paths, so any node-agent-initiated drop of a chunk repairs the timestamp chain rather than silently forking it.

### Related

- #866 introduces the "stitch chunk" concept (a metadata-and-scalars-only replacement that preserves a dropped chunk's `(previousReportTimestamp, reportTimestamp)` pair) for its own new drop paths. The same mechanism is the natural fix here.

aryanghai12:
## Overview

- **Current Behavior:** `enforceMaxSize` (LRU eviction) and `MaxAttempts` retry exhaustion dropped queued chunks without replacement, breaking the container's `(previousReportTimestamp, reportTimestamp)` chain. Storage requires an unbroken chain tracing back to the container's first report to complete a profile; dropping a chunk caused the affected container to remain stuck in `Learning` mode indefinitely with no visible errors.
- **Future Behavior:** Both paths now use the stitch-repair mechanism introduced in #866 to preserve timestamp continuity. `MaxAttempts` exhaustion triggers `dropChunk(dropReasonMaxAttemptsExhausted)`, and LRU evictions in `enforceMaxSize` enqueue a metadata-only stitch chunk. Stitch creation is bounded to 10% of `MaxQueueSize` (`maxStitchBacklogFor`) to prevent infinite queue growth during eviction, falling back to `dropReasonLRUBacklogExhausted` when full.

## Additional Information

- **Stitch Refactoring:** Factored out duplicated stitch-construction logic across `dropChunk`, `requeueSplit`, and `enforceMaxSize` into a shared `newStitchFor` helper function.
- **Metrics & Observability:** Introduced metric labels (`dropReasonMaxAttemptsExhausted`, `dropReasonLRUEvicted`, `dropReasonLRUBacklogExhausted`) and exposed `stitchBacklog` / `maxStitchBacklog` in `GetQueueStats`.
- **Documentation:** Updated `docs/features/container-profile-split-on-413.md` to reflect the resolution of this known limitation.

## How to Test

Run the full package validation and race detection test suite:

```bash
go build ./...
go vet ./pkg/containerprofilemanager/...
go test ./pkg/containerprofilemanager/... -race
```

*Covered by existing rewritten tests (`TestQueueLRUEviction`, `TestQueueDropsProfileAfterMaxAttempts`) and new unit tests (`TestEnforceMaxSize_EvictionEnqueuesStitch`, `TestEnforceMaxSize_BacklogBoundsRepairCost`).*

## Examples/Screenshots

N/A (backend queue and timestamp continuity repair)

## Related issues/PRs:

* Resolved #871

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved report-chain recovery when profiles are dropped after retry exhaustion or queue eviction.
  * Added bounded repair handling so replacement data cannot bypass queue capacity limits.
  * Added clear drop handling when the repair backlog is exhausted.
  * Clarified that chunks rejected with HTTP 413 are dropped without re-stitching.

* **Documentation**
  * Documented drop reasons, repair behavior, backlog limits, restart behavior, and known limitations.


pkg/containerprofilemanager/v1/queue/containerprofile_queue.go
```diff
@@ -103,10 +138,25 @@ type QueueData struct {
...
+	// yet dequeued for processing or eviction). Every enqueue/dequeue of the underlying dque
+	// funnels through enqueueLocked/enqueueStitchNoEvict and processAllItems/enforceMaxSize
+	// respectively, which keep this in step; see enforceMaxSize for what it's used for.
+	//
+	// It starts at zero on every process start and is never reconciled against stitches the
+	// disk queue may already hold from a prior run (dque has no cheap way to count a
+	// predicate over its contents without draining it). This can only ever under-count, so the
+	// worst case is enforceMaxSize allowing a brief burst of extra replacements right after a
+	// restart before it settles - never an incorrect refusal to repair a chain.
+	stitchBacklog atomic.Int64
```
matthyx, line 170: **Blocker.** This comment's claim — *"This can only ever under-count, so the worst case is enforceMaxSize allowing a brief burst of extra replacements right after a restart before it settles"* — doesn't hold.

The counter starts at 0 while the disk queue may already hold stitches from the prior run. Each of those decrements it on the way out (`:387`, `:489`), so it goes negative and stays there: nothing restores it except enqueueing an equal number of new stitches, and if the persisted stitches simply succeed, the deficit is permanent for the process lifetime.

The drift is bounded by the stitch count at startup, i.e. up to `maxQueueSize`. At defaults (1000 / 100) the effective budget can reach ~1100 — exactly the "convert the whole queue in one call" pathology `enforceMaxSize`'s doc says this bound prevents, and it'd do it as ~1000 disk enqueues under `qd.mu`.

Reproduced with a throwaway test: after one restart over a single persisted stitch, `stitchBacklog=-1` against `bound=1`, so the budget is 2.

Suggest clamping the decrement at zero via a small CAS helper used at both sites, plus a test that reopens a queue over a persisted stitch and asserts `stitchBacklog >= 0`.

pkg/containerprofilemanager/v1/queue/containerprofile_queue.go
```diff
@@ -520,14 +690,19 @@ func (qd *QueueData) dropChunk(dropped *QueuedContainerProfile, reason dropReaso
...
-		Attempts:    dropped.Attempts,
-		SplitDepth:  qd.maxSplitDepth,
-		IsStitch:    true,
+	if qd.stitchBacklogFull() {
+		logger.L().Warning("stitch backlog is exhausted, dropping container profile chunk without a replacement, its report chain is now forked",
+			helpers.String("name", dropped.Profile.Name),
+			helpers.String("namespace", dropped.Profile.Namespace),
+			helpers.String("containerID", dropped.ContainerID))
+
+		qd.chunksDropped.Add(1)
```
matthyx: **Blocker.** This is the second `chunksDropped.Add(1)` for the *same* chunk — `:686` already counted the drop. Verified: one `dropChunk` call with the backlog at its bound produces `chunksDropped=2, reasons=[max-attempts-exhausted stitch-backlog-exhausted]`.

Same double-count at `:620`+`:629` and `:407`+`:418`. But `enforceMaxSize`'s own backlog-full path (`:396-404`) counts the identical logical event exactly once — so the two paths disagree.

Net effect: `chunksDropped` over-reports by up to 2× precisely under overload, and `sum(chunk_dropped_total) by (reason)` stops equalling the number of chunks dropped, so any alert on it silently double-counts.

Either increment `chunksDropped` once and keep both reason labels (documenting that one chunk can emit two samples), or split "repair refused" onto its own counter. Whichever you pick, please apply it at all four sites — and tighten the test, which currently only asserts `Contains` and so passes either way.

pkg/containerprofilemanager/v1/queue/containerprofile_queue.go
```diff
@@ -249,23 +300,130 @@ func (qd *QueueData) enqueueLocked(item *QueuedContainerProfile) error {
...
+// newStitchFor builds the metadata-only replacement enqueued in place of dropped, so the
+// container's report-timestamp chain (see stitchChunk) survives the drop. SplitDepth is set to
+// qd.maxSplitDepth so a stitch that later fails and needs its own drop reason goes straight to
+// dropReasonStitchRejected rather than treating a chunk that has nothing left to give as
+// splittable.
+func (qd *QueueData) newStitchFor(dropped *QueuedContainerProfile) *QueuedContainerProfile {
+	return &QueuedContainerProfile{
+		Profile:     stitchChunk(dropped.Profile),
+		ContainerID: dropped.ContainerID,
+		Attempts:    dropped.Attempts,
```
matthyx: **Worth discussing before merge.** Reached from `dropReasonMaxAttemptsExhausted`, `dropped.Attempts` is already `>= maxAttempts`, so this stitch gets exactly one attempt before being dropped un-restitched — your own test pins it: `assert.Equal(t, maxAttempts+1, creator.callCount(), "the original gets its full budget, the stitch exactly one try")`.

Applied to #871's motivating scenario: storage is down, an item burns ~30 minutes of retries, `dropChunk` fires, the stitch is tried once ~5s later against the same down storage, fails, chain forks anyway. For a sustained outage the `MaxAttempts` repair is a no-op.

It does help the other case — an item-specific persistent rejection, where a metadata-only stitch lands immediately — and that's worth keeping. But inheriting `Attempts` is load-bearing for the *413* lineage (it bounds the split/stitch cascade), and that reasoning doesn't transfer here: a fresh budget for the retry-exhaustion path is still bounded, since a stitch is never re-stitched.

Either give this path its own budget, or state the limitation plainly in the `dropReasonMaxAttemptsExhausted` comment and the docs' Known limitations — as written they read as though the path is fixed outright.

pkg/containerprofilemanager/v1/queue/containerprofile_queue.go
```diff
@@ -249,23 +300,130 @@ func (qd *QueueData) enqueueLocked(item *QueuedContainerProfile) error {
...
+		qd.stitchBacklog.Add(1)
+	}
+
+	return nil
+}
 
+// enqueueStitchNoEvict adds a repair stitch directly to the tail without invoking
+// enforceMaxSize.
+//
+// It exists only for enforceMaxSize's own use: enforceMaxSize always runs under qd.mu (it's
```
matthyx: Minor: this invariant isn't true as stated. `NewQueueData:250` calls `qd.enforceMaxSize()` directly, without holding `qd.mu` — so it is *not* "only ever called from enqueueLocked."

It's safe in practice (no other goroutine exists at construction), but this file's correctness rests on lock discipline and the comment asserts something the code doesn't hold. Either note the constructor exception here or take the lock in `NewQueueData`.

Related, and confirmed in a probe: that startup sweep now *enqueues stitches*. Reopening a full queue took it 10 → 9, converting one real chunk into a stitch (no net progress) and forking a second — so a restart at capacity now sheds two deltas where it used to shed one, with no new item waiting for the room. Worth a docs note, or skip the repair when `enforceMaxSize` runs outside an enqueue.

matthyx (commented):
## Review

Verified locally on `ad06ebf` (fresh clone of the PR head):

```
go build ./...                                  ok
go vet ./pkg/containerprofilemanager/...        ok
go test ./pkg/containerprofilemanager/... -race ok (4.99s)
```

The direction is right and the `newStitchFor` extraction is a faithful refactor — the struct literals it replaces were byte-for-byte identical at all three sites. The docs update is thorough and CodeRabbit's one actionable comment is addressed.

I'm not approving yet: **two of the findings below are defects against invariants this PR itself states**, and both have small, mechanical fixes. The third is a design question that decides whether the `MaxAttempts` half of this fix actually does anything in the scenario it targets.

---

### 🔴 Blocker 1 — `stitchBacklog` drifts negative across a restart, permanently loosening the bound

Reproduced (probe test, since removed):

```
run1: size=10 stitchBacklog=1                  # one eviction repaired with a stitch
--- restart (same queue dir) ---
run2 after processAllItems: stitchBacklog=-1   # bound=1
CONFIRMED: effective budget is now 2 instead of 1
```

The counter starts at 0 every process start, but the *disk* queue can already hold stitches from the previous run. Every one of those decrements the counter as it leaves (`:387` in `enforceMaxSize`, `:489` in `processAllItems`), driving it negative — and nothing ever pulls it back except enqueueing an equal number of fresh stitches. If the persisted stitch succeeds on its first try, the deficit is permanent for the life of the process.

The field comment says the worst case is *"a brief burst of extra replacements right after a restart before it settles."* It doesn't settle on its own, and the burst isn't bounded by anything small: the drift is bounded by the number of stitches resident at startup, i.e. up to `maxQueueSize`. At the defaults (`MaxQueueSize=1000`, `maxStitchBacklog=100`), a process that starts after a heavy-overload run can end up with an effective budget near 1100 — which is precisely the pathology `enforceMaxSize`'s own doc comment says the bound exists to prevent: *"a single call here walk[ing] through and convert[ing] the whole queue before ever making room for the new item that triggered it"* — 1000 disk enqueues in one `enforceMaxSize` call, all under `qd.mu`.

Cheapest fix is to make the release monotonic at the floor:

```go
// releaseStitch decrements the resident-stitch count without letting it go below zero:
// the counter starts at zero each process start while the disk queue may already hold
// stitches from a prior run, and an unclamped decrement would leave a permanent deficit
// that silently widens maxStitchBacklog.
func (qd *QueueData) releaseStitch() {
	for {
		cur := qd.stitchBacklog.Load()
		if cur <= 0 {
			return
		}
		if qd.stitchBacklog.CompareAndSwap(cur, cur-1) {
			return
		}
	}
}
```

...called from both decrement sites, plus a test that reopens a queue over a persisted stitch and asserts `stitchBacklog >= 0`. (Reconciling the real count at startup would be more accurate, but you already note dque can't count a predicate cheaply — clamping is enough to keep the bound honest.)

### 🔴 Blocker 2 — `chunksDropped` counts one chunk twice on the backlog-exhausted and enqueue-failed paths

Reproduced: a single `dropChunk` call with the backlog at its bound yields

```
chunksDropped=2, reasons=[max-attempts-exhausted stitch-backlog-exhausted]
```

`dropChunk` increments at `:686` for the drop itself, then again at `:699` when the repair is refused. Same shape at `:620`+`:629` (`requeueSplit`) and `:407`+`:418` (`enforceMaxSize`'s failed enqueue). Meanwhile `enforceMaxSize`'s *own* backlog-full path at `:396-404` increments exactly once for the same logical event.

So `chunksDropped` is not a count of dropped chunks — it over-reports by up to 2× exactly when things are going worst, and it's internally inconsistent between two paths that describe the same outcome. The drop-reason counter has the same problem: `sum(chunk_dropped_total) by (reason)` no longer equals chunks dropped, so any dashboard or alert built on it silently double-counts overload.

Two coherent options — either is fine, but please pick one and apply it at all four sites:

1. **One increment per chunk.** Keep the double `ReportContainerProfileChunkDropped` (the two labels do carry distinct information: *why dropped* and *why not repaired*) but increment `chunksDropped` only once. Then document on the metric that a chunk emits one or two reason samples, so nobody sums them as a chunk count.
2. **Separate the concerns.** Keep `chunksDropped` as a pure chunk count and move "repair refused" onto its own counter/metric (`stitchesRefused`), so both series stay individually summable.

`TestDropChunk_StitchBacklogExhaustedForksWithoutReplacement` currently asserts only `Contains` on the reasons, so it passes under either — worth tightening it to pin the counter value once you've decided.

### 🟠 Discuss — the `MaxAttempts` repair stitch inherits an exhausted budget, so it gets exactly one try

`newStitchFor` copies `Attempts: dropped.Attempts`. Coming from `dropReasonMaxAttemptsExhausted` that value is already `>= maxAttempts`, so the stitch is attempted once, fails, and is dropped un-restitched. Your own test pins this: `assert.Equal(t, maxAttempts+1, creator.callCount(), "the original gets its full budget, the stitch exactly one try")`.

Now walk the motivating scenario from #871: storage is unreachable, an item burns 360 retries over ~30 minutes, `dropChunk` fires, the stitch is tried once ~5s later against the *same* unreachable storage, fails, and the chain forks anyway. **For a sustained outage — the case the retry budget exists to survive — the `MaxAttempts` half of this fix changes nothing but the log line.** It does genuinely help the other case (an item-specific persistent rejection, where a metadata-only stitch succeeds immediately), and that's worth having; I just don't think the PR description's framing holds for the outage case.

Inheriting `Attempts` is clearly correct for the 413 lineage — it's what bounds the split/stitch cascade. For the retry-exhaustion path the trade looks different: a fresh budget is still bounded (a stitch is never re-stitched, so total exposure is 30 min + 30 min, not unbounded), and a stitch is a couple of KB. The cost is queue *slots*, not bytes — which is real, but it's the same cost you already accepted by enqueueing the stitch at all.

If you keep the current behaviour, could you say so explicitly in the `dropReasonMaxAttemptsExhausted` comment and in the docs' "Known limitations" — something to the effect that the repair is best-effort and does not survive the outage that caused the drop? Right now the docs read as though this path is fixed outright.

### 🟡 Minor

1. **`enqueueStitchNoEvict`'s lock-discipline claim is wrong** (`:313`): *"enforceMaxSize always runs under qd.mu (it's only ever called from enqueueLocked, whose callers must already hold the lock)"*. `NewQueueData:250` calls `enforceMaxSize()` directly, without the lock. It's safe there — no other goroutine exists yet — but this is a concurrency-critical file and the comment asserts an invariant the code doesn't hold. Please either note the constructor exception or move the startup sweep under the lock.

2. **A restart at capacity now sheds two real chunks instead of one.** Confirmed in the same probe: reopening a full queue took it 10 → 9, evicting one real chunk into a stitch (no net progress) and then forking a second. That startup sweep runs with no new item waiting for room, so the second eviction buys nothing. Pre-fix this cost one delta per restart; it now costs two, every restart, on every node. Worth a line in the docs at minimum — or skipping the repair entirely when `enforceMaxSize` is called outside an enqueue.

3. **`GetQueueStats` can report a negative `stitchBacklog`** — falls out of Blocker 1, and it'll look like a bug to whoever reads the stats endpoint. Fixed by the clamp.

4. **`stitchBacklogFull()` is a soft bound in `dropChunk`** (`:693` checks outside `qd.mu`, `:706` enqueues inside). A `dropChunk` on the processor goroutine can race an `enforceMaxSize` under the lock and overshoot by one. Harmless in practice, but the doc comment reads as a hard invariant — a one-line "advisory, may overshoot by one under concurrent admission" would set expectations correctly.

### 👍 Confirmed correct

- `enforceMaxSize` terminates: every iteration either shrinks the queue or spends one unit of a bounded budget, so the loop is bounded at `maxStitchBacklog + 1` iterations — modulo Blocker 1 inflating that bound.
- I traced `TestQueueLRUEviction`'s rewritten expectations by hand against `enqueueStitchNoEvict` appending at the tail; `[stitch(0), profile-3, profile-4]` with `chunksDropped == 3` is right.
- The new drop reasons are compile-time constants from a closed set, so Prometheus/OTEL label cardinality stays bounded.
- `EmptyQueue`'s `stitchBacklog.Store(0)` correctly resets the counter alongside the drain.

Happy to re-review as soon as 1 and 2 are in — they're both small, and I'd merge on your answer to the third.

matthyx (approved):
## Re-review of `39bcdf8` — all findings addressed, approving ✅

Re-ran everything on the new head:

```
go build ./...                                  ok
go vet ./pkg/containerprofilemanager/...        ok
go test ./pkg/containerprofilemanager/... -race ok (4.96s)
```

I re-ran my two original repro probes **unchanged** against the new code rather than reading the commit message, and added two more to check the fixes didn't just move the problem.

### Blocker 1 — `stitchBacklog` negative drift → fixed

Same probe as before, same restart-over-a-persisted-stitch setup:

```
before (ad06ebf):  stitchBacklog=-1  bound=1   → effective budget 2
after  (39bcdf8):  stitchBacklog=0   bound=1   → full=false, budget intact
```

`releaseStitch`'s CAS loop is correct — the `cur <= 0` early return means a lost race retries against a fresh load, and the loop can't spin indefinitely since every competing operation makes progress.

I also checked the direction of the residual error, since clamping trades one inaccuracy for another. Interleaving untracked (prior-run) releases with tracked enqueues, the counter never exceeded the true resident count — so it can under-count and permit an extra repair, but it can never over-count and *refuse* one. That matches what the new field comment claims, and it's the safe direction: the failure mode is a slightly generous budget for a bounded window, not a silently unrepaired chain.

### Blocker 2 — `chunksDropped` double-count → fixed

```
before (ad06ebf):  one dropChunk → chunksDropped=2
after  (39bcdf8):  one dropChunk → chunksDropped=1, reasons=[max-attempts-exhausted stitch-backlog-exhausted]
```

Option 1, which is the one I'd have picked: one increment per lost chunk, both reason labels retained since they answer different questions. I audited all five increment sites in the final file — `:455`, `:465`, `:682`, `:716`, `:746` — and every follow-up "repair also failed" branch (`:478`, `:691`, `:703`, `:762`, `:781`) now reports a reason without touching the counter. The rule holds uniformly, including on the path that was already correct.

Tightening the assertions from `Contains` to exact-slice equality in `TestDropChunk_StitchBacklogExhaustedForksWithoutReplacement` was the right call — that test would otherwise have passed under either scheme.

### Discussion item — fresh retry budget → adopted, with a consequence worth naming

`newStitchFor(dropped, freshAttempts)` with `dropChunk` passing `reason == dropReasonMaxAttemptsExhausted` is exactly the narrow application I was hoping for: the 413/split lineage keeps inheriting `Attempts` (which is what bounds the split/stitch cascade), and only the retry-exhaustion repair gets a fresh window. Since a stitch is never re-stitched, total exposure is two retry windows — I confirmed that bound holds by draining a queue against a permanently-failing creator and seeing exactly N+1 drops for N originals, never N+2.

One behavioural consequence to be aware of, not a blocker: because a fresh-budget stitch now occupies a backlog slot for a full retry window (~30 min at defaults) instead of ~5s, **`maxStitchBacklog` has become a cap on how many chains can be repaired concurrently during a sustained outage** — roughly 100 at defaults. Probed with `maxStitchBacklog=1` and 5 dropping containers: exactly one got a repair stitch, the other four reported `stitch-backlog-exhausted`.

I think that's the right trade — the repairs it displaces were the one-shot kind that couldn't have succeeded anyway — and it's bounded at 10% of the queue and directly observable through the drop-reason metric. Just worth knowing that during a long outage the `stitch-backlog-exhausted` counter is expected to be non-zero and isn't by itself a sign of a bug.

### Minors — all four addressed

- `NewQueueData` now takes `qd.mu` around the startup sweep, so `enqueueStitchNoEvict`'s invariant is true as written rather than documented-around. Good call taking the lock instead of carving out an exception.
- The two-deltas-per-restart-at-capacity cost is now documented in Known limitations, and the doc correctly notes that exactly-`MaxQueueSize` already triggered an eviction pre-change.
- Negative `stitchBacklog` can no longer surface in `GetQueueStats`; the new test asserts it via the stats map, which is the right place to pin it.
- `stitchBacklogFull`'s advisory nature is now stated on the function.

### Also checked, no issues

- No deadlock from the new `NewQueueData` locking — `enforceMaxSize` reaches only `enqueueStitchNoEvict`, which doesn't take `qd.mu`.
- A stitch that exhausts its own fresh budget still hits the `dropped.IsStitch` early return in `dropChunk`, so it's dropped un-restitched; no loop.
- Queue-occupancy cost of longer-lived stitches stays capped at 10% of `MaxQueueSize` by the existing bound.
- `requeueSplit`'s second-half-failure path was already single-counting and was correctly left alone.

Nice work — the fixes are minimal, the reasoning is in the comments rather than the commit message, and each one landed with a test that would actually catch the regression. LGTM.

PR #886, opened 2026-08-09
test(utils): attribute missing /proc prefixes to tracer or normalization layer

### LINKED ISSUE #874: Add test to catch when /proc prefixes go missing and assert cause
Current tests didnt catch that for some cases (and why for these cases) we lost the /proc prefix.

@entlein must RCA this and add a test, ideally one that tells us whether it broke at the kernel level or the regex level

Also must check if or not this is related to https://github.com/kubescape/node-agent/issues/721

aryanghai12:
## Overview

- **Current Behavior:** Unrooted `/proc` entries (e.g., `root`, `exe`, `cwd`) and truncated tracer basenames (e.g., `ocal.sh`) leak into ContainerProfiles as invalid root-level paths (`/ocal.sh`), causing false positives in `opens`.
- **Future Behavior:** Adds a test-only suite (`normalize_path_attribution_test.go`) that acts as a drift guard for `NormalizePath` and accurately attributes path corruption to either the normalization layer or the tracer layer before evidence is destroyed.

## Additional Information

- **RCA - Normalization Layer:** `NormalizePath` uses an explicit allowlist to re-root `/proc/<pid>/<entry>` paths. Allowlist drift causes newly unhandled entries to leak into profiles without a `/proc` prefix.
- **RCA - Tracer Layer:** When `fpath` is empty, `GetFullPath()` falls back to `fname` (a basename). `NormalizePath` prepends `/` to bare strings, fabricating fictional root paths (e.g., `ocal.sh` -> `/ocal.sh`).
- **Scope:** Test-only PR; `pkg/utils/path.go` remains untouched. Uncovered `/proc` entries are pinned as currently leaking to keep main CI green while exposing drift.

## How to Test

Run the newly added unit test suite:

```bash
go test ./pkg/utils/ -count=1 -v -run 'TestNormalizePath|TestClassifyRawTracerPath|TestOpenExecPathAsymmetry'
```

## Examples/Screenshots

N/A (test-only suite and RCA)

## Related issues/PRs:

* Resolved #874
* Follow-up to #721
* Follow-up to #872

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive regression coverage for path attribution.
  * Improved validation of numeric-directory handling and root path generation.
  * Added checks for known historical samples and `/proc` path coverage.
  * Documented expected differences between open and execute path attribution.

### DIFF
--- a/pkg/utils/normalize_path_attribution_test.go
+++ b/pkg/utils/normalize_path_attribution_test.go
@@ -1,0 +1,309 @@
+package utils
+
+import (
+	"strings"
+	"testing"
+)
+
+// This file is the regression net for #874 ("catch when /proc prefixes go
+// missing and assert cause"), the follow-up to #721 and to the allowlist
+// widening in #872.
+//
+// The point is not only to catch a missing /proc prefix, but to say *which
+// layer* dropped it. Two very different defects reach the profile looking
+// alike once they are written out as an `opens` stanza:
+//
+//   1. normalization layer — the tracer delivered a structurally intact
+//      /proc/<pid>/<entry> path that merely lost its /proc root. The original
+//      is reconstructable from the string alone, so a miss here is a gap in
+//      headlessProcRegex (node-agent's fault, fixable in this repo).
+//
+//   2. tracer layer — the tracer delivered only a *basename*, because the
+//      open event's `fpath` field was empty and DatasourceEvent.GetPath /
+//      GetFullPath fall back to `fname`. Nothing in this repo can reconstruct
+//      the directory: the information never arrived.
+//
+// Telling them apart matters because NormalizePath actively destroys the
+// evidence for case (2) — see TestNormalizePath_FabricatesRootFromBasename.
+
+// pathDefectLayer is the layer a malformed tracer path is attributable to.
+type pathDefectLayer int
+
+const (
+	// layerNone: a well-formed absolute path, nothing to attribute.
+	layerNone pathDefectLayer = iota
+	// layerNormalization: recoverable in this repo by re-rooting under /proc.
+	layerNormalization
+	// layerTracer: unrecoverable here — the full path never reached us.
+	layerTracer
+)
+
+func (l pathDefectLayer) String() string {
+	switch l {
+	case layerNormalization:
+		return "normalization"
+	case layerTracer:
+		return "tracer"
+	default:
+		return "none"
+	}
+}
+
+// classifyRawTracerPath attributes a *raw* tracer path — the value read from
+// the `fpath`/`fname`/`exepath` field, before NormalizePath runs — to the layer
+// responsible for any defect in it.
+//
+// It must be fed the raw field value. Running it on NormalizePath's output is
+// meaningless, because NormalizePath rewrites a bare basename into something
+// that is indistinguishable from a genuine root-level file.
+func classifyRawTracerPath(raw string) pathDefectLayer {
+	// Nothing delivered at all, or runc's "." / "/." placeholder from #721.
+	if raw == "" || raw == "." || raw == "/." {
+		return layerTracer
+	}
+	// No separator anywhere: this is a basename, so `fpath` was empty and the
+	// `fname` fallback supplied it. The directory is simply not in the event.
+	if !strings.Contains(strings.TrimPrefix(raw, "/"), "/") && !strings.HasPrefix(raw, "/") {
+		return layerTracer
+	}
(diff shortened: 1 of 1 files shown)


pkg/utils/normalize_path_attribution_test.go
```diff
@@ -0,0 +1,309 @@
...
+	}
+	// No separator anywhere: this is a basename, so `fpath` was empty and the
+	// `fname` fallback supplied it. The directory is simply not in the event.
+	if !strings.Contains(strings.TrimPrefix(raw, "/"), "/") && !strings.HasPrefix(raw, "/") {
+		return layerTracer
+	}
+	// Shaped like a /proc/<pid>/... path that lost its root. Structurally
+	// intact, so recovering it is normalization's job.
+	if headlessProcRegex.MatchString(raw) || isHeadlessProcShaped(raw) {
+		return layerNormalization
```
matthyx, line 77: **Blocker**: `isHeadlessProcShaped` misclassifies valid top-level numeric directory paths as normalization layer defects.

In `classifyRawTracerPath(raw)`:
```go
if headlessProcRegex.MatchString(raw) || isHeadlessProcShaped(raw) {
    return layerNormalization
}
```
`isHeadlessProcShaped(raw)` checks if `raw` starts with `/<digits>/`. For paths like `"/2024/logs"`, `isHeadlessProcShaped("/2024/logs")` returns `true`, causing `classifyRawTracerPath("/2024/logs")` to return `layerNormalization`. However, `TestNormalizePath_NumericDirIsNotAPID` defines `"/2024/logs"` as a valid numeric directory that should not be re-rooted under `/proc`.

`classifyRawTracerPath` should either check against known `/proc` entries or return an ambiguous/unclassified layer for paths outside `headlessProcRegex`, preserving valid numeric directory handling.

pkg/utils/normalize_path_attribution_test.go
```diff
@@ -0,0 +1,309 @@
...
+	openFallbackShapes := []string{"redis-server", "sh", "runc"}
+	for _, p := range openFallbackShapes {
+		if got := NormalizePath(p); got != "/"+p {
+			t.Errorf("NormalizePath(%q) = %q, want %q", p, got, "/"+p)
+		}
+		if layer := classifyRawTracerPath(p); layer != layerTracer {
+			t.Errorf("classifyRawTracerPath(%q) = %v, want tracer", p, layer)
+		}
+	}
+}
```
matthyx: **Blocker**: `TestOpenExecPathAsymmetry` and `TestNormalizePath_FabricatesRootFromBasename` test string literals directly against `NormalizePath` rather than testing `DatasourceEvent` methods.

The test comments note that the asymmetry is in `DatasourceEvent.GetPath`/`GetFullPath` falling back from empty `fpath` to `fname` while `GetExePath` does not. However, the tests call `NormalizePath` directly with raw strings like `"redis-server"` and `"local.sh"`. If `DatasourceEvent` fallback logic regresses, these tests will still pass because `DatasourceEvent` is never constructed or asserted.

`TestOpenExecPathAsymmetry` and `TestNormalizePath_FabricatesRootFromBasename` should instantiate synthetic `DatasourceEvent` instances with empty `fpath` and basename `fname`, asserting the behavior of `GetPath`/`GetFullPath` vs `GetExePath` before passing the result to `NormalizePath`.

matthyx (commented):
Reviewed PR #886. Found 2 blockers in `pkg/utils/normalize_path_attribution_test.go` that should be addressed:

1. **Numeric directory misclassification in `classifyRawTracerPath`**: `isHeadlessProcShaped` classifies valid numeric directories (e.g. `/2024/logs`) as `layerNormalization` defects.
2. **Missing `DatasourceEvent` fallback testing**: `TestOpenExecPathAsymmetry` and `TestNormalizePath_FabricatesRootFromBasename` test `NormalizePath` directly with raw string literals instead of verifying `DatasourceEvent.GetPath` / `GetFullPath` fallback behavior.

matthyx (approved):
Approved! Both blockers have been fully resolved in commit `2564202fc9916d36aacbed26b820cf3e3830f330`:

1. **Numeric directory classification**: `isHeadlessProcShaped` now verifies entry names against `knownProcEntryNames`, properly attributing valid numeric directories (e.g. `/2024/logs`) to `layerNone`.
2. **`DatasourceEvent` fallback coverage**: Added synthetic `Open` and `Exec` event generators to test `GetPath`, `GetFullPath`, and `GetExePath` fallback logic directly.

The PR is clean and ready to merge.

PR #887, opened 2026-08-11
fix(networkstream): key stream entities by container ID outside Kubernetes

AlonLiwsky:
## Summary

Outside Kubernetes the network stream folded every container's traffic onto the single node entity, so no per-container entity existed and per-workload attribution was impossible. This keys entities by container ID instead — collapsing onto the node only for traffic that is not a container's — and creates the entity on first sight where nothing announces containers. The Kubernetes path is unchanged.

## Overview

`handleNetworkEvent` and `handleDnsEvent` decide which stream entity a connection belongs to. Today the fallback onto the node entity fires whenever `ns.k8sObjectCache == nil`:

```go
if entityId == "" || ns.k8sObjectCache == nil {
    entityId = ns.nodeName
}
```

That cache is a Kubernetes API cache. In this repo's binary it is always present, but embedders that run the sensor outside a cluster construct the stream with a nil cache — and there *every* event, including every container's, is folded onto the single node entity. No per-container entity is ever created, so every container on the machine is indistinguishable in the stream and per-workload attribution downstream is impossible.

After this change the collapse happens only for traffic that is not a container's: an empty container ID, or the `"host"` sentinel that `procfs.go` assigns to processes outside every tracked mount namespace. Everything else keys by container ID.

Keying alone is not enough, because an event for an entity that does not exist is dropped, and outside Kubernetes nothing subscribes this package's `ContainerCallback` — so no container entity is ever announced. When `kubernetesMode` is off, the entity is therefore created on the container's first event, and pruned once it goes an interval without traffic, so a machine with container churn does not accumulate dead ones.

**In Kubernetes it keeps dropping, and that is deliberate.** A missing entity in a cluster does not mean "never announced"; it means the container was removed or is ignored. Removal is not quiet, either: the container collection publishes `EventTypeRemoveContainer` *before* the container leaves the collection and then holds it in a 2-second cache for enrichers, while the watcher dispatches this callback onto a worker pool. So events for a container that is already gone keep arriving for a while afterwards. Creating entities for those would put a container carrying no pod or workload identity into the payload, for a pod that no longer exists, on every pod termination.

Two smaller fixes ride along:

- **Traffic recorded before a container is announced is no longer thrown away.** `containerCallback` submits its subscribers to a worker pool while events flow straight through, so a container's first outbound connection routinely lands before its `AddContainer` reaches this package. The announcement now fills identity in over the event maps already there instead of replacing them. The node entity is excluded — it is not announced, so it has no such window, and excluding it keeps that path byte-identical.
- **The `k8sInventory` lookups in `buildNetworkEvent` are nil-guarded.** The inventory is only built when `kubernetesMode` is set, so a pod- or service-kind endpoint arriving anywhere else would dereference a nil interface.

## Additional Information

**Effect on Kubernetes: none.** The removed `k8sObjectCache == nil` disjunct never fired there, and every new behaviour is either behind `!cfg.KubernetesMode` or scoped away from the node entity. `TestHandleNetworkEvent_KeyingIgnoresK8sObjectCache` and `TestHandleNetworkEvent_KubernetesStillDropsUnknownEntities` pin that.

One known gap is deliberately **not** addressed here. `ContainerCallback` deletes a stopped container's entity along with any traffic it had not yet flushed, losing up to one interval — which matters most for a short-lived container that makes a single outbound connection. That behaviour is pre-existing and untouched by this PR: it is neither introduced nor widened here. But per-container entities make it matter more, so I will follow up with a separate PR that retires the entity at the next flush instead of deleting it. I am keeping it separate because that one *is* a Kubernetes behaviour change and deserves its own review rather than riding inside a change whose point is that Kubernetes is unaffected.

Docs: adds `docs/features/network-stream-entity-keying.md`. Also drops `hostNetworkSensorEnabled` and `hostMalwareSensorEnabled` from `docs/CONFIGURATION.md` — neither key exists anywhere in the codebase, while `hostMonitoringEnabled`, which does, was missing from the table.

## How to Test

This package does not build on macOS, since inspektor-gadget is Linux-only. On a Mac:

```
docker run --rm -v "$PWD":/src -w /src -v "$(go env GOMODCACHE)":/go/pkg/mod \
  golang:1.25 bash -c 'go test -race ./pkg/networkstream/...'
```

On Linux, `go test -race ./pkg/networkstream/...`.

15 new tests (18 cases with subtests) cover the keying, the platform gate, entity creation and pruning, the announcement merge in both arrival orders, node-entity restoration, and the nil-inventory guard. I checked each one earns its place by reverting the corresponding behaviour individually and confirming that test — and only that test — fails. The rest of the unit suite is unaffected.

## Summary by CodeRabbit

* **New Features**
  * Network and DNS activity is now attributed to separate container entities.
  * Non-Kubernetes environments can create temporary container entities from observed traffic.
  * Host activity is consistently associated with the node entity.
  * Container metadata and activity recorded before announcement are preserved.
  * Network endpoint enrichment safely supports pod and service details when inventory is available.

* **Bug Fixes**
  * Unknown Kubernetes entities continue to be excluded.
  * Idle, unannounced container entities are automatically removed.
  * Snapshot and shutdown handling now safely isolate and flush collected data.

* **Documentation**
  * Updated the host monitoring configuration reference.
  * Added documentation for network-stream entity keying and lifecycle behavior.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -154,8 +154,7 @@ correlation is an upgrade, not a dependency.
 | `nodeProfileServiceEnabled` | bool | `false` | Enable node profiling |
 | `fimEnabled` | bool | `false` | Enable File Integrity Monitoring |
 | `httpDetectionEnabled` | bool | `false` | Enable HTTP traffic parsing |
-| `hostMalwareSensorEnabled` | bool | `false` | Enable host-level malware sensor |
-| `hostNetworkSensorEnabled` | bool | `false` | Enable host-level network sensor |
+| `hostMonitoringEnabled` | bool | `false` | Monitor the host's own namespace as a synthetic container |
 | `prometheusExporterEnabled` | bool | `false` | Enable Prometheus metrics |
 | `fullPathTracingEnabled` | bool | `true` | Include full executable paths |
 | `enableEmbeddedSBOMs` | bool | `false` | Use embedded SBOMs from images |
--- a/docs/features/network-stream-entity-keying.md
+++ b/docs/features/network-stream-entity-keying.md
@@ -1,0 +1,102 @@
+# Network-stream entity keying
+
+The network stream groups a flush interval's connections into **entities**: one per
+container, plus one for the node itself. Downstream, the entity is the unit a finding
+is attributed to, so two workloads sharing an entity are indistinguishable to every
+consumer of the stream.
+
+This document covers how an event picks its entity, in `pkg/networkstream/v1`. The
+per-connection process identity carried inside an entity is a separate concern — see
+[network-stream-process-attribution.md](network-stream-process-attribution.md).
+
+## The rule
+
+```go
+entityID := event.GetContainerID()
+if entityID == "" || entityID == armotypes.HostContainerID {
+    entityID = nodeName          // the node's own traffic
+}
+```
+
+Only traffic that is not a container's collapses onto the node entity: an event with
+no container ID, or with the `"host"` sentinel that `procfs.go` assigns to processes
+outside every tracked mount namespace. Everything else keys by container ID.
+
+## What this replaced
+
+The node fallback used to fire on `ns.k8sObjectCache == nil` as well:
+
+```go
+if entityId == "" || ns.k8sObjectCache == nil {
+    entityId = ns.nodeName
+}
+```
+
+The object cache is a Kubernetes API cache. In this repo's binary it is always
+present, but embedders that run the sensor outside a cluster — on a host, or on an
+AWS ECS instance — construct the stream with a nil cache. There *every* event,
+including every container's, was folded onto the single node entity: an ECS task's
+containers all reported as the instance, and per-workload attribution downstream was
+impossible.
+
+The keying on Kubernetes is unchanged. There the cache is always non-nil, so the
+removed disjunct was never the reason an event moved to the node.
+
+## Outside Kubernetes, entities are created on first sight
+
+An entity is normally announced by `ContainerCallback`. An event for an entity that
+does not exist is logged and dropped.
+
+That drop is why the keying fix alone is not enough outside Kubernetes: the embedders
+above do not subscribe this package's `ContainerCallback` at all, so no container
+entity is ever announced and every container event would key to an ID that is not in
+the map. When `kubernetesMode` is off, `entityForEventLocked` therefore creates a
+container entity the first time it sees one, carrying the container ID and nothing
+else — the pod and workload fields stay empty until something announces the container.
+
+**In Kubernetes it keeps dropping**, which is what makes this change inert there. A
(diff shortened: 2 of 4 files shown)


matthyx (commented):
## Review

Went through `entityForEventLocked`, the `unannouncedEntities` prune path, the `ContainerCallback` merge, and the `k8sInventory` nil-guard in `buildNetworkEvent`, plus the doc changes. Checked out `fix/networkstream-ecs-entity-keying` locally and ran the package suite (`go test -race -v ./pkg/networkstream/...`): all 15 new tests + the existing suite pass, no races. `go build ./...`, `gofmt -l`, and `golangci-lint run ./pkg/networkstream/...` are clean on the touched files (the only lint hits are pre-existing `SA1019` deprecation warnings in `wire.go`/`wire_test.go`, untouched by this PR).

Logic checks out:
- The node-collapse condition (`entityID == "" || entityID == armotypes.HostContainerID`) correctly drops the `k8sObjectCache == nil` disjunct, and `cfg.KubernetesMode` is the right signal to gate creation-on-first-sight vs. drop-on-miss — it's the same flag that already gates building `k8sInventory` in the constructor, so it's consistent with an existing invariant rather than a new assumption.
- `unannouncedEntities` bookkeeping is correct: cleared on announce (both directions of the announce/first-event race), cleared on remove, and the prune loop deletes from the map it's ranging over, which Go permits.
- `entityForEventLocked` restoring the node entity as `Host` (not inventing a container keyed by the node name) closes a latent gap in the old code too — previously, if the host container's `RemoveContainer` fired, host traffic was silently dropped until the next flush interval recreated the node entity.
- `docs/CONFIGURATION.md`: confirmed `hostMalwareSensorEnabled`/`hostNetworkSensorEnabled` don't exist anywhere in the tree, and `hostMonitoringEnabled` does (`pkg/config/config.go`, `pkg/containerwatcher/v2/container_watcher_collection.go`), so that table edit is accurate.

No code-correctness blockers found. Two things stand in the way of merging, unrelated to the change itself:

1. **DCO check is failing** (`action_required`) — the commit needs a `Signed-off-by` trailer (`git commit --amend -s`).
2. **This PR is still a draft** and `build-and-push-image` is red — the log shows `Username and password required` on the `quay.io` login step, which looks like registry credentials not being available to a fork-sourced workflow run rather than anything introduced by this diff.

I'll hold off on approving until it's marked ready for review and the DCO check is green — happy to re-review immediately once those are resolved.

matthyx (approved):
Re-checked: DCO now passes and the PR is out of draft. Diffed the current head (`bc687e2`) against what I reviewed before — identical, so this was just the sign-off amend, no code changes to re-review.

`build-and-push-image` is still red on the same `quay.io` login step (`Username and password required`), but that's a registry-credentials gap that looks generic to fork-sourced PRs in this repo, not something this diff affects — branch protection doesn't require that check, and it's orthogonal to the change under review.

Approving.

PR #889, opened 2026-08-12
Chore/retire headless proc regex in favor of opens-gadget fix -- remove band-aid from last week

entlein:
# Paths revisited

THIS PR  was tested stacked on https://github.com/kubescape/node-agent/pull/864 - thus **only the last two commits are relevant** and could be cherry picked.
However the test evidence was built on the CPs (and possibly a few commits behind current main).

The gadget d3fc71c1343f6db5af38833681fec11ccf7c045c itself **should be** standalone 
Once merged, this 135eafd414f672f0b7bdf8b9dccf4cf3d8161e33 retires the **band-aid** regex fix.

## Side-by-side: the same postgres deploy, three node-agents and their definition of `path`

Comparison on fresh 2-node k3s (kernel 6.1.167): `example/postgres/distros/deploy-distros.sh oss` into a clean namespace, one 50k-row `INSERT` + `CHECKPOINT`, learn to completion, read the learned ContainerProfile.

| leg | node-agent | learned opens | **fabricated roots** | data-dir paths |
|---|---|---|---|---|
| **A — upstream stock** (chart 1.40.3, quay) | `v0.3.158` lineage | 253 | **26** | `/base/16384/16423_fsm`, `/backup_label`, `/base/1/⋯` … |
| **B — fork current** (`v0.3.190`, stock gadget) | ghcr `v0.3.190` | 227 | **22** | same shapes |
| **C — fork + this PR's gadget** | same `v0.3.190` binary, fixed `tracers.tar` | 221 | **0** | `/var/lib/postgresql/data/pgdata/base/16384/⋯`, `…/pgdata/backup_label` |

The same file, three ways:

```
A/B:  /base/16384/16423_fsm            <- fabricated root (relative name + "/" prefix)
C:    /var/lib/postgresql/data/pgdata/base/16384/⋯    <- resolved against the dirfd/cwd
A/B:  /backup_label
C:    /var/lib/postgresql/data/pgdata/backup_label
```

Leg B vs A also shows what the fork's `IsResolvedFullPath` guard  already fixed — the *scrambled-fragment* class (`/ocal.sh`, wildcard-root collapse) that upstream still has over longer windows; the fabricated-relative class is untouched by that guard and only this PR removes it.

CI: `Test_43_RelativeOpenPathResolution` (added here) learns a chdir'd relative-open workload in kind and asserts no fabricated roots and that even the FAILED open resolves — passing:

```
learned opens (34): [/data/reldir/absent.txt /data/reldir/present.txt /proc/⋯/task/1/fd ...]
--- PASS: Test_43_RelativeOpenPathResolution (185.66s)
```

## Summary by CodeRabbit

* **New Features**
  * Added tracing for `open` and `openat` operations, including filenames, results, metadata, optional stack information, and resolved paths.
  * Added relative path resolution using directory context.
  * Added build and run support for the tracing capability.

* **Bug Fixes**
  * Preserved valid numeric-leading paths and prevented incorrect `/proc` prefixing.
  * Corrected attribution of relative opens to their absolute working directory.

* **Tests**
  * Added end-to-end coverage for successful and failed relative file opens.


entlein:
This supersedes #881

tests/component_test.go
```diff
@@ -3595,3 +3595,58 @@ func Test_35_ExecTTYFieldTest(t *testing.T) {
...
+		}
+		for _, o := range c.Opens {
+			opens = append(opens, o.Path)
+		}
+	}
+	t.Logf("learned opens (%d): %v", len(opens), opens)
+
+	matchesResolved := func(name string) bool {
+		for _, p := range opens {
+			if p == "/data/reldir/"+name || strings.HasPrefix(p, "/data/reldir/") {
```
matthyx: **Blocker: `matchesResolved` doesn't actually check what it claims to.**

```go
if p == "/data/reldir/"+name || strings.HasPrefix(p, "/data/reldir/") {
    return true
}
```

The `strings.HasPrefix(p, "/data/reldir/")` clause makes the exact-name check meaningless: as soon as *any* path under `/data/reldir/` is learned (e.g. `present.txt`), `matchesResolved("absent.txt")` also returns `true` even if `absent.txt` was never resolved at all. So the two asserts at lines 3648/3650 can't independently fail — the whole point of this test (proving the *failed* open resolves too, not just the successful one) isn't actually enforced.

Since this test is the main evidence backing the PR's central claim, suggest tightening it to the exact match only:

```go
matchesResolved := func(name string) bool {
    for _, p := range opens {
        if p == "/data/reldir/"+name {
            return true
        }
    }
    return false
}
```

pkg/ebpf/gadgets/trace_open/program.bpf.c
```diff
@@ -0,0 +1,176 @@
...
+	u32 pid = (u32)pid_tgid;
+
+	ap = bpf_map_lookup_elem(&start, &pid);
+	if (!ap)
+		return 0; /* missed entry */
+	ret = ctx->ret;
+	if (targ_failed && ret >= 0)
+		goto cleanup; /* want failed only */
+
+	event = gadget_reserve_buf(&events, sizeof(*event));
```
matthyx, line 110: **Suggested fix: `event->fpath` can be submitted uninitialized when `paths` is false.**

`event` comes from `gadget_reserve_buf` (a ring-buffer reservation), which is not zeroed. When `paths == false` (the gadget's own default, `gadget.yaml` line ~52), none of the branches at lines 119/127/131 touch `event->fpath`, so the full 512-byte `fpath` array is submitted at line ~151 containing whatever was previously in that ring-buffer slot.

Today this looks unreachable in practice: `OpenTracer` in `pkg/containerwatcher/v2/tracers/open.go` always drives the eBPF `paths` param and the userspace `FullPathTracing` flag from the same `cfg.EnableFullPathTracing`, and userspace only reads `fpath` when `FullPathTracing` is true. But that coupling lives entirely outside this file, and any other consumer subscribed to this datasource (raw `ig` CLI usage, other operators, debug dumps) would see stale ring-buffer bytes in `fpath` with no indication they're garbage.

Cheap defense-in-depth fix: initialize `event->fpath[0] = '\0'` right after the `gadget_reserve_buf` succeeds (line 111), before the `paths`-gated branches.

pkg/ebpf/gadgets/trace_open/filesystem_patched.h
```diff
@@ -0,0 +1,256 @@
...
+		if (!f)
+			return -1;
+		base = BPF_CORE_READ(f, f_path);
+	}
+
+	char *bstr = get_path_str(&base);
+	if (!bstr)
+		return -1;
+
+	long n = bpf_probe_read_kernel_str(buf, GADGET_PATH_MAX, bstr);
```
matthyx, line 237: Nit: `read_full_path_of_dfd_rel` takes a `buf_len` parameter but ignores it, hardcoding `GADGET_PATH_MAX` here and at lines 244-245 instead. Not a live bug — the only call site (`program.bpf.c`) always passes `sizeof(event->fpath)`, which is exactly `GADGET_PATH_MAX` — but it's a landmine if this helper ever gets reused with a differently-sized buffer, since the masking arithmetic (`& (GADGET_PATH_MAX - 1)`) would silently wrap into the caller's buffer using the wrong size instead of bounding against the real `buf_len`. Worth using `buf_len` (and rejecting if it doesn't fit) to make the signature honest.

pkg/ebpf/gadgets/trace_open/filesystem_patched.h
```diff
@@ -0,0 +1,256 @@
...
+	struct path base;
+
+	if (dfd == AT_FDCWD) {
+		struct task_struct *task = (struct task_struct *) bpf_get_current_task();
+		if (!task)
+			return -1;
+		struct fs_struct *fs = BPF_CORE_READ(task, fs);
+		if (!fs)
+			return -1;
+		bpf_probe_read_kernel(&base, sizeof(base), &fs->pwd);
```
matthyx: Nit: `bpf_probe_read_kernel(&base, sizeof(base), &fs->pwd)` reads `fs->pwd` at a compile-time offset instead of through a CO-RE relocation, unlike the `f_path` read a few lines down (`base = BPF_CORE_READ(f, f_path)`) and the rest of this file. If `fs_struct`'s layout differs on a running kernel from the BTF this was built against, this read silently returns garbage instead of failing. Suggest `BPF_CORE_READ_INTO(&base, task, fs, pwd)` for consistency and portability.

matthyx (changes requested):
Nice, well-evidenced PR — the side-by-side fabricated-root comparison in the description is convincing, and `Test_43_RelativeOpenPathResolution` is exactly the right shape of regression test for this. `pkg/utils/path.go` correctly gets simpler by removing the headless-proc-regex band-aid now that the tracer resolves paths itself.

Two things I'd like fixed before merge:

1. **`tests/component_test.go` (line 3628)** — `matchesResolved`'s `strings.HasPrefix` fallback makes the test unable to distinguish "the failed open resolved" from "some other open under the same dir resolved." Since this test is the evidence the fix works, it should assert the exact path.
2. **`pkg/ebpf/gadgets/trace_open/program.bpf.c` (line 110)** — `event->fpath` can be submitted with uninitialized ring-buffer bytes when `paths` is false. Currently masked by `OpenTracer` always keeping the eBPF `paths` param and userspace `FullPathTracing` in sync, but that's an invariant enforced outside this file, and it's a one-line fix to close off entirely.

Also left two non-blocking nits on `filesystem_patched.h` (unused `buf_len` parameter, and a non-CO-RE read inconsistent with the rest of the file).

Separately, FYI (not something to fix in this PR): `build-and-push-image` is failing on both CI runs with "Username and password required" logging into `quay.io/kubescape` — that's the registry-secrets gate that external-contributor PRs don't get access to, not a code issue here.

Happy to re-review once the two blockers above are addressed — the core approach and evidence look solid.

pkg/ebpf/gadgets/trace_open/program.bpf.c
entlein: Thanks, done that in line 116, now retesting the comparison

ConstanzeTU: Addressed — `event->fpath[0] = '\0'` is now set unconditionally right after `gadget_reserve_buf`, before the `paths`-gated branches, so a failed open (or `paths` disabled) never emits stale ring-buffer bytes.

ConstanzeTU: Addressed — `event->fpath[0] = '\0'` is now set unconditionally right after `gadget_reserve_buf`, before the `paths`-gated branches, so a failed open (or `paths` disabled) never emits stale ring-buffer bytes.

pkg/ebpf/gadgets/trace_open/program.bpf.c
```diff
@@ -0,0 +1,176 @@
...
+			long r = read_full_path_of_open_file_fd(
+				fd, (char *)event->fpath, sizeof(event->fpath));
+			if (r <= 0)
+				event->fpath[0] = '\0';
+		}
+	} else {
+		errval = -ret;
+		if (paths)
+			event->fpath[0] = '\0';
+	}
```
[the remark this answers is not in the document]
ConstanzeTU: Addressed — `event->fpath[0] = '\0'` is now set unconditionally right after `gadget_reserve_buf`, before the `paths`-gated branches, so a failed open (or `paths` disabled) never emits stale ring-buffer bytes.

tests/component_test.go
ConstanzeTU: Addressed — `matchesResolved` asserts the exact `/data/reldir/<name>` path; the prefix fallback is removed, so the failed-open (`absent.txt`) assertion can fail independently of the successful open.

tests/component_test.go
```diff
@@ -3595,3 +3595,58 @@ func Test_35_ExecTTYFieldTest(t *testing.T) {
...
+				return true
+			}
+		}
+		return false
+	}
```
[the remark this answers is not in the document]
ConstanzeTU: Addressed — `matchesResolved` asserts the exact `/data/reldir/<name>` path; the prefix fallback is removed, so the failed-open (`absent.txt`) assertion can fail independently of the successful open.

pkg/ebpf/gadgets/trace_open/filesystem_patched.h
```diff
@@ -0,0 +1,256 @@
...
+	if (off > GADGET_PATH_MAX - REL_NAME_MAX)
+		off = GADGET_PATH_MAX - REL_NAME_MAX;
+	buf[off & (GADGET_PATH_MAX - 1)] = '/';
+	long m = bpf_probe_read_user_str(&buf[(off + 1) & (GADGET_PATH_MAX - 1)],
+					 REL_NAME_MAX - 1, user_fname);
+	if (m <= 1) {
+		buf[off & (GADGET_PATH_MAX - 1)] = '\0';
+		return -1;
+	}
+	return off + m;
```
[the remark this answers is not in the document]
ConstanzeTU: Addressed — the base-too-long case now fails closed (`return -1`, buffer cleared) instead of clamping and appending, so no truncated plausible-but-wrong absolute path is emitted.

pkg/ebpf/gadgets/trace_open/filesystem_patched.h
ConstanzeTU: Addressed — `pwd` is now read via `BPF_CORE_READ(task, fs, pwd)`, matching the `f_path` read below.

pkg/ebpf/gadgets/trace_open/filesystem_patched.h
ConstanzeTU: Resolved by dropping the parameter: `buf` is always the `GADGET_PATH_MAX`-sized `fpath`, and wiring a runtime length into the masked append is rejected by the verifier (`invalid access to memory, mem_size=912 off=911 size=255`) — the masks need a compile-time-constant bound. Using the true buffer size (`GADGET_PATH_MAX`) throughout keeps the signature honest without the dead parameter.

ConstanzeTU:
All review points addressed:
- `event->fpath` is initialised unconditionally after the buffer reservation, so a failed open (or `paths` disabled) never emits stale ring-buffer bytes.
- `read_full_path_of_dfd_rel` reads `pwd` via CO-RE and fails closed when the base cannot fit the appended name within `GADGET_PATH_MAX` (no truncated path).
- `matchesResolved` asserts the exact resolved path.
- Added the relative numeric-leading case to the normalize test.
- On `buf_len`: wiring a runtime length into the masked append is rejected by the verifier (`invalid access to memory, mem_size=912 off=911 size=255`), so the buffer size stays the compile-time `GADGET_PATH_MAX` and the unused parameter is dropped.

The relative-open resolution is exercised by a component test that deploys a workload doing `chdir`'d relative opens of a present and a missing file and asserts both resolve to `/data/reldir/<name>` with no fabricated roots — full green run: https://github.com/k8sstormcenter/node-agent/actions/runs/31722304957

matthyx (approved):
all good now, let's upstream the gadget fix when you have time

PR #890, opened 2026-08-13
Feat/symmetric network rules for CEL R0011 and R0012

entlein:
Not sure how to open a stacked PR , which is why you see the entire https://github.com/kubescape/node-agent/pull/864 

The relevant diff is in the last 3 commits.

Note that this assumes that we depreciated the AP and NN completely and substituted the rules themselves. So all these three PRs are stacked on that change


entlein:
Closing cause the diff is getting too large, the test-equivalence is lost ATM

PR #892, opened 2026-08-15
fix rule failure runtime alert path handling

khuswant18:
## Description

Fixes incorrect host path resolution in `setBaseRuntimeAlert`.

The existing implementation checked `err` after `GetPathFromPid`, even though the function already returned when `err != nil`. As a result, that error branch was unreachable.

When `ProcessTree.Path` was already available, the local `path` variable remained empty and could be used to construct an incorrect `hostPath`.

## Changes

- Remove the unreachable `err != nil` branch.
- Use `ProcessTree.Path` directly when it is already available.
- Resolve the path using `GetPathFromPid` only when `ProcessTree.Path` is empty.
- Use the container PID when available, with the process PID as a fallback.
- Scope error variables closer to where they are used.

## Testing

- Ran `gofmt` and `git diff --check`.
- Verified Linux compilation with:

  `GOOS=linux GOARCH=amd64 go build ./pkg/rulemanager/ruleadapters/...`

- Ran the relevant Go tests in a Linux Docker environment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved executable path resolution for runtime alerts involving containerized processes.
  * Alerts now use the correct host path when a container-to-process mapping is available.
  * Added fallback handling when container mappings or process paths are unavailable.
  * Corrected file-size error handling to improve alert reliability.

### DIFF
--- a/pkg/rulemanager/rule_manager.go
+++ b/pkg/rulemanager/rule_manager.go
@@ -27,6 +27,7 @@ import (
 	"github.com/kubescape/node-agent/pkg/metricsmanager"
 	"github.com/kubescape/node-agent/pkg/objectcache"
 	"github.com/kubescape/node-agent/pkg/objectcache/containerprofilecache"
+	"github.com/kubescape/node-agent/pkg/otelsetup"
 	"github.com/kubescape/node-agent/pkg/processtree"
 	bindingcache "github.com/kubescape/node-agent/pkg/rulebindingmanager"
 	"github.com/kubescape/node-agent/pkg/rulemanager/cel"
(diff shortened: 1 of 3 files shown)


matthyx (approved):
Reviewed the fix and the underlying bug it addresses.

**Bug confirmed real:** in the old code the `if err != nil` branch after `GetPathFromPid` was indeed dead (the function already returns on error), so whenever `ProcessTree.Path` was already set, the local `path` var was empty and `hostPath` was built as `/proc/<pid>/root/` — silently truncating the path.

**Fix looks correct:**
- `getHostPath` now cleanly separates the two cases: resolve via `GetPathFromPid` only when `ProcessTree.Path` is empty, otherwise use `ProcessTree.Path` directly.
- Preferring the container's tracked PID (`containerIdToPid`) over the process's own PID when resolving `/proc/<pid>/root/...`, with fallback to the process PID, matches the existing pattern already used in `malwaremanager/v1/malware_manager.go` for the same host-path-resolution problem — so this isn't a new pattern, it's consistent with the rest of the codebase.
- The `containerIdToPid` map is correctly threaded through: `RuleManager` owns the live `maps.SafeMap`, passes `&r.containerIdToPid` into `NewRuleFailureCreator`, and it's populated/cleared in `containercallbacks.go` — verified this wiring compiles and the map is populated before use.
- The `err` shadowing fix in `setBaseRuntimeAlert` (local `var err error` for the `GetFileSize` call) is a nice cleanup, no functional issue there.

**Verified locally:**
- `go build ./...` — clean
- `go vet ./pkg/rulemanager/...` — clean
- `go test ./pkg/rulemanager/...` — all pass, including the two new `getHostPath` unit tests (mapped + unmapped container ID)

No blockers. CodeRabbit's earlier actionable comment (wiring `containerIdToPid` through) was already addressed in a follow-up commit, and the latest review pass shows no remaining actionable comments.

Approving.

matthyx (approved):

PR #904, opened 2026-08-18
fix(hostsensormanager): stop matching decoy release files as os-release

magic-peach:
## Overview

`getOsReleaseFile` picks the /etc entry whose name ends in `os-release`. On RHEL-family hosts that also match `centos-release`, `redhat-release`, etc, since they end with the same substring. `Readdirnames` doesn't guarantee ordering, so which file wins can vary between runs, and those distro files aren't KEY=VALUE formatted like the real os-release, so whatever consumes `OsReleaseFileSpec.Content` downstream ends up parsing the wrong thing.

Switched to an exact filename match, since there's only one file we actually want here.

## How to Test

Added a test that seeds a temp /etc with both `centos-release` and `os-release` present and asserts the real one wins, plus a case with only the decoy present. Confirmed the first test fails against the old `HasSuffix` check and passes with the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OS release detection to select only the exact `os-release` file.
  * Prevented similarly named files from being incorrectly identified as the operating system release file.
  * Added error handling when no valid `os-release` file is available.

### DIFF
--- a/pkg/hostsensormanager/sensor_osrelease.go
+++ b/pkg/hostsensormanager/sensor_osrelease.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"os"
 	"path"
-	"strings"
 
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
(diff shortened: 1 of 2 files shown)


matthyx (approved):
LGTM — the exact-match fix correctly addresses the decoy-file bug: `readdirnames` order was unspecified, so `centos-release`/`redhat-release` etc. could win over the real `os-release` since they share the suffix, and those files aren't `KEY=VALUE` formatted. Switching to `==` fixes that, and `hostPath(etcDirName)` joins with basenames from `Readdirnames`, so the exact match is correct there.

Tests look solid: cover both the decoy-present-with-real-file case and the decoy-only-not-found case, and correctly restore `hostFSPrefix` via `defer`.

Nit (non-blocking): `osReleaseFileSuffix` is now a slightly stale name since it's used for exact equality rather than a suffix check — worth a rename to something like `osReleaseFileName` in a follow-up, but not worth blocking on.

Approving.

PR #906, opened 2026-08-19
fix(tracers): bounded retry for DNSTracer gadget startup

matthyx:
## Problem

`DNSTracer.Start()` only logs a failed `RunGadget` call and never retries:

```go
err := dt.runtime.RunGadget(dt.gadgetCtx, nil, params)
if err != nil {
    logger.L().Error("Error running gadget", ...)
}
```

A single transient failure permanently stops DNS event collection
(`utils.DnsEventType`) until the node-agent pod restarts.

That failure class is inherently transient. `armosec/private-node-agent#511`
root-caused an intermittent

```
apply CO-RE relocations: load BTF for kmod kvm_intel: rebase split spec: raw BTF differs
```

coming from a cache race in the pinned `cilium/ebpf` fork's CO-RE/BTF loading
layer (unrelated to `trace_dns` itself — `trace_dns` is an unmodified
upstream OCI gadget image). A fix for that race is proposed upstream:
matthyx/ebpf#1. But any gadget startup failure in this class — CO-RE
relocation racing other concurrently-loading gadgets — is timing-dependent
by nature, so a bounded retry is worthwhile independent of that fix landing.

## Fix

Wrap the existing background `RunGadget` call in a bounded retry
(`dnsStartMaxRetries = 5`), using this repo's existing
`github.com/cenkalti/backoff` convention (see
`containercallback.go`'s `setSharedWatchedContainerData` for the established
pattern). Each attempt gets its own `GadgetContext` (the previous one is
left terminated after a failed run, and it's also what `Stop()` will
cancel), and retries stop promptly once the tracer's context is canceled.

## Scope note

Only `dns.go` is touched. `armosec/private-node-agent#511`'s suggested
secondary fix also mentions `http.go`, `gotls.go`, `ssl.go`, and
`pkg/tracermanager/tailcalls.go` as "sibling tracers" — checking this repo's
actual `pkg/containerwatcher/v2/tracers/` directory, `gotls.go`/`ssl.go`/
`tailcalls.go` don't exist here at all (they're `armosec/private-node-agent`'s
own custom TLS tracers, using a manual uprobe-attach mechanism that never
goes through this CO-RE/BTF loading path, so they aren't exposed to this
particular race). This repo's own `http.go` exists but wasn't part of
#511's evidence, so it's left alone to keep this PR scoped to the actual
reported failure.

## Testing

`dns_retry_test.go` adds a minimal `runtime.Runtime` fake and three tests:

- `TestDNSTracerStartRetriesOnTransientFailure`: fails every attempt but the
  last, confirms the tracer retries through to success.
- `TestDNSTracerStartGivesUpAfterMaxRetries`: always fails, confirms exactly
  `dnsStartMaxRetries` attempts happen and no more.
- `TestDNSTracerStartStopsRetryingWhenContextCanceled`: confirms canceling
  the tracer's context stops further retries promptly.

```
--- PASS: TestDNSTracerStartRetriesOnTransientFailure (3.48s)
--- PASS: TestDNSTracerStartGivesUpAfterMaxRetries (4.54s)
--- PASS: TestDNSTracerStartStopsRetryingWhenContextCanceled (1.01s)
```

`go build ./...` and `go vet ./pkg/containerwatcher/...` are clean. The rest
of the `tracers` package's pre-existing `*Fields` tests fail in this
checkout for an unrelated reason (missing `tracers.tar` fixture, not present
in a shallow clone — every `*Fields` test across every tracer file fails
identically, confirming it's fixture-wide and not caused by this change).

AI-skills: none | cmds: /clear,/oh-my-claudecode:autopilot

## Summary by CodeRabbit

* **Bug Fixes**
  * DNS monitoring now automatically retries startup after temporary failures.
  * Retries use bounded backoff and stop after a maximum number of attempts.
  * Stopping or canceling monitoring now safely halts pending retries, including during backoff.
  * Improved reliability when starting DNS tracing after transient runtime or environment issues.
  * Prevented unnecessary restart attempts after monitoring has been stopped or canceled.

### DIFF
--- a/pkg/containerwatcher/v2/tracers/dns.go
+++ b/pkg/containerwatcher/v2/tracers/dns.go
@@ -2,7 +2,10 @@ package tracers
 
 import (
 	"context"
+	"sync"
+	"time"
 
+	"github.com/cenkalti/backoff"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/datasource"
 	gadgetcontext "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-context"
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/operators"
@@ -21,14 +24,24 @@ import (
 const (
 	dnsImageName = "ghcr.io/inspektor-gadget/gadget/trace_dns:v0.48.1"
 	dnsTraceName = "trace_dns"
+
+	// dnsStartMaxRetries bounds retries of a failed gadget start. Startup
+	// failures in this class are transient/timing-dependent (see
+	// armosec/private-node-agent#511: a CO-RE/BTF cache race in the pinned
+	// cilium/ebpf fork can abort relocation for any kernel module with
+	// split BTF present at that moment), so a single failed attempt does
+	// not mean the gadget can never start.
+	dnsStartMaxRetries = 5
 )
 
 var _ containerwatcher.TracerInterface = (*DNSTracer)(nil)
 
 // DNSTracer implements TracerInterface for events
 type DNSTracer struct {
-	eventCallback      containerwatcher.ResultCallback
-	gadgetCtx          *gadgetcontext.GadgetContext
+	eventCallback containerwatcher.ResultCallback
+	gadgetCtxMu   sync.Mutex
+	gadgetCtx     *gadgetcontext.GadgetContext
+
 	kubeManager        operators.DataOperator
 	ociStore           *orasoci.ReadOnlyStore
 	runtime            runtime.Runtime
@@ -55,9 +68,11 @@ func NewDNSTracer(
 	}
 }
 
-// Start initializes and starts the tracer
-func (dt *DNSTracer) Start(ctx context.Context) error {
-	dt.gadgetCtx = gadgetcontext.New(
+// newGadgetContext builds a fresh GadgetContext for a single run attempt and
+// records it as the tracer's current one, so Stop() always cancels whichever
+// attempt is in flight.
+func (dt *DNSTracer) newGadgetContext(ctx context.Context) *gadgetcontext.GadgetContext {
+	gadgetCtx := gadgetcontext.New(
 		ctx,
 		// This is the image that contains the gadget we want to run.
 		dnsImageName,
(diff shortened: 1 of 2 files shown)


pkg/containerwatcher/v2/tracers/dns.go
```diff
@@ -89,8 +131,13 @@ func (dt *DNSTracer) Stop() error {
...
+
+	dt.gadgetCtxMu.Lock()
+	gadgetCtx := dt.gadgetCtx
+	dt.gadgetCtxMu.Unlock()
+
+	if gadgetCtx != nil {
+		gadgetCtx.Cancel()
 	}
 	return nil
 }
```
matthyx, line 162: **Blocker: `Stop()` doesn't reliably stop retries.**

`backoff.WithContext(bo, ctx)` on line 108 is scoped to the `ctx` passed into `Start`, not to anything `Stop()` touches. `Stop()` only cancels `dt.gadgetCtx` — the child `GadgetContext` of whichever attempt is currently in flight (or about to be replaced).

Trace it through: if `Stop()` is called while an attempt is running, canceling `gadgetCtx` makes `RunGadget` return an error, which `backoff.RetryNotify` treats as a normal transient failure — since the outer `ctx` is still alive, it schedules and runs *another* attempt via a brand-new `dt.newGadgetContext(ctx)` (line 114), one that `Stop()` never touches (it already returned). So a `Stop()` call can leave the tracer starting the DNS gadget again after the caller believes it's stopped — exactly backwards from what a "stop" is supposed to guarantee, and worse than the pre-PR behavior (which had no retries to race with `Stop()` at all).

There's also no test covering this path — all three new tests only exercise cancellation of the `Start(ctx)` argument, never a `Stop()` call made mid-retry/mid-run.

Suggested fix: give `DNSTracer` its own cancelable context (e.g. store a `context.CancelFunc` derived from the `Start` ctx), use that context — not the raw `Start` ctx — for `backoff.WithContext` and for every `newGadgetContext` call, and have `Stop()` call that cancel func in addition to canceling the current `gadgetCtx`. Then add a regression test that calls `Stop()` while `RunGadget` is failing/retrying and asserts no further `RunGadget` calls happen afterward.

(Minor, non-blocking nit: `Start()` also creates a `GadgetContext` on line 100 that's immediately discarded/overwritten by the first attempt's own `newGadgetContext` call on line 114 — dead work, though harmless since it's never passed to `RunGadget`.)

matthyx (commented):
Good, well-scoped fix for the transient CO-RE/BTF startup race, and the write-up/tests are thorough. But there's a real correctness bug in the retry/cancellation interplay that needs fixing before merge — see the inline comment on `Start`/`Stop`. Short version: `Stop()` can race with the retry loop and end up letting the tracer start the gadget again after it was supposedly stopped, since the retry loop is scoped to `Start`'s `ctx` rather than anything `Stop()` cancels. CodeRabbit's automated review flagged the same issue independently, which corroborates it.

Separately, the DCO check is currently failing (`action_required`) — that'll need a signed-off commit before this can merge regardless of the code review outcome.

Not approving yet (and note: GitHub won't let this identity formally request changes on its own PR anyway) — this is a blocker, please treat it as such. Happy to take another look once the cancellation path is fixed and DCO is green.

pkg/containerwatcher/v2/tracers/dns.go
matthyx: Fixed in 81ad939: gave Start its own cancelable context (derived from the caller's), used it — not the raw Start ctx — for `backoff.WithContext` and every `newGadgetContext` call, and had Stop cancel it before canceling the current GadgetContext. That closes the gap: after the in-flight attempt's `RunGadget` returns, `RetryNotify` now sees the canceled context and stops instead of scheduling another attempt.

Also dropped the dead `newGadgetContext` call you flagged in the minor nit — with the retry-loop context canceled synchronously inside Stop, any GadgetContext built in that same race window already inherits a canceled context regardless of whether that call ran, so it had no remaining purpose.

Added `TestDNSTracerStopStopsRetryingMidBackoff`, which calls Stop while an attempt is failing/retrying and asserts no further RunGadget calls happen afterward — confirmed it fails on the pre-fix code (a second call happens after Stop returns) and passes with the fix.

matthyx:
Re-reviewed after commit 81ad939. The fix is correct: `Start` now derives its own `context.WithCancel(ctx)` and uses that (not the raw caller ctx) for both `backoff.WithContext` and every `newGadgetContext` call, and `Stop` calls that `cancel()` before canceling the in-flight `GadgetContext`. Since `contextBackOff.NextBackOff()` checks `ctx.Done()` before returning a further wait, canceling that context makes `RetryNotify` stop instead of scheduling another attempt — closing the race whether `Stop()` lands mid-run or mid-backoff. `TestDNSTracerStopStopsRetryingMidBackoff` covers exactly that path. Resolving my thread.

Remaining item before merge: the DCO check is still `action_required` — needs a signed-off commit. Once that's green this looks good to merge from a code-correctness standpoint.

PR #911, opened 2026-08-21
feat(rulemanager): add rule filtering based on context

khuswant18:
## Summary

Implements context-based rule filtering so that rules are only evaluated against events from matching contexts (Kubernetes, Host, Standalone, ECS, Container), reducing unnecessary rule processing.

Previously, all registered rules were evaluated for every event regardless of context. This PR introduces a `context:` tag convention on rules and filters them at both the retrieval and per-rule evaluation stages.

Resolves the `TODO: rule filtering based on context` in `rule_manager.go`.

### Core matching logic
- **New file: `rulecreator/context_match.go`** — Adds `RuleMatchesContext()` as the single source of truth for context-based matching. Rules are matched via `context:<name>` tags with the following semantics:
  - Exact match: `context:host` matches Host, `context:kubernetes` matches Kubernetes, etc.
  - Meta-tag: `context:container` matches all container-type contexts (Kubernetes, Standalone, ECS, Container) but not Host.
  - Backward compatibility: rules with **no** context tags default to Kubernetes-only.
  - Multi-tag: rules can have multiple `context:` tags.

### Integration in rule_manager.go
- Non-K8s contexts now retrieve rules via `CreateRulesForContext()` instead of pod-based bindings.
- Per-rule `RuleAppliesToContext()` check added inside the evaluation loop as a safety net.

### Interface & plumbing
- **`ruleengine_interface.go`** — Added `CreateRulesForContext(ctx)` to the `RuleCreator` interface.
- **`factory.go`** — Implemented `CreateRulesForContext()` on `RuleCreatorImpl`, filtering via `RuleMatchesContext` and initializing prefilters.
- **`ruleengine_mock.go`** — Mock implementation using the same `RuleMatchesContext` logic.
- **`rulepolicy.go`** — Simplified `RuleAppliesToContext()` to a thin bridge: resolves `nil` context to Kubernetes, then delegates to `RuleMatchesContext`.

### Tests (4 new test files, ~40 test cases)
- **`context_match_test.go`** — 20 cases covering exact match, meta-tag, backward compat, multi-tag, mixed tags.
- **`factory_context_test.go`** — Table-driven tests for `CreateRulesForContext` across all 5 contexts, plus edge cases (empty rules, no match, prefilter init, consistency with `RuleMatchesContext`).
- **`rulepolicy_test.go`** — 11 cases for `RuleAppliesToContext` including `nil` contextInfo handling.

## Summary by CodeRabbit

* **Bug Fixes**
  * Rules are now selected according to the event source context, improving behavior across Kubernetes, host, standalone, ECS, and container environments.
  * Untagged rules continue to apply only to Kubernetes events.
  * Container-context rules now match all supported container-related event sources consistently.

* **Tests**
  * Added comprehensive coverage for context-based rule matching and filtering across supported environments.

### DIFF
--- a/pkg/rulemanager/rule_manager.go
+++ b/pkg/rulemanager/rule_manager.go
@@ -297,8 +297,7 @@ func (rm *RuleManager) ReportEnrichedEvent(enrichedEvent *events.EnrichedEvent) 
 	if enrichedEvent.SourceContext == nil || enrichedEvent.SourceContext.Context() == contextdetection.Kubernetes {
 		rules = rm.ruleBindingCache.ListRulesForPod(namespace, pod)
 	} else {
-		// TODO: rule filtering based on context
-		rules = rm.ruleBindingCache.GetRuleCreator().CreateAllRules()
+		rules = rm.ruleBindingCache.GetRuleCreator().CreateRulesForContext(enrichedEvent.SourceContext.Context())
 	}
 
 	if len(rules) == 0 {
--- a/pkg/rulemanager/rulecreator/context_match.go
+++ b/pkg/rulemanager/rulecreator/context_match.go
@@ -1,0 +1,35 @@
+package rulecreator
+
+import (
+	"strings"
+
+	"github.com/kubescape/node-agent/pkg/contextdetection"
+	typesv1 "github.com/kubescape/node-agent/pkg/rulemanager/types/v1"
+)
+
+// RuleMatchesContext is the single source of truth for context-based rule matching.
+func RuleMatchesContext(rule *typesv1.Rule, currentContext contextdetection.EventSourceContext) bool {
+	isContainerContext := currentContext == contextdetection.Kubernetes ||
+		currentContext == contextdetection.Standalone ||
+		currentContext == contextdetection.Container ||
+		currentContext == contextdetection.ECS
+
+	var hasContextTags bool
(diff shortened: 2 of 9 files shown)


matthyx (approved):
Reviewed the context-based rule filtering change.

- `RuleMatchesContext` correctly consolidates the exact/meta-tag/backward-compat matching logic that previously lived only in `rulepolicy.go`; `rulepolicy.go` and the new `CreateRulesForContext` both delegate to it, so there's a single source of truth.
- Verified the matching semantics against `contextdetection.EventSourceContext` (kubernetes/host/standalone/container/ecs): exact tag match, `context:container` meta-tag covering all container-type contexts but not host, and untagged rules defaulting to Kubernetes-only — all consistent with the pre-existing behavior in `rulepolicy.go` and well covered by the new table-driven tests.
- The K8s hot path (`ListRulesForPod`) is untouched; only non-K8s contexts now go through `CreateRulesForContext`, so this doesn't add lock contention to the primary event path.
- The data race CodeRabbit flagged (`Prefilter` lazily written without holding `r.mutex` in `CreateRulesForContext`/`CreateAllRules`) was already fixed in commit 6d7406d — both methods now take `r.mutex.Lock()` consistent with the other mutators in `factory.go`.
- CI (CodeQL + full component-tests suite) is green.

No blockers. Approving.

PR #915, opened 2026-08-23
feat(cel/network): serviceRef/serviceSelector/host neighbor resolution

entlein:
## NOT READY FOR REVIEW
assumed stacked on #902 or even part of it, needs a storage change though whereas #902 is NA only.

## Why this is added
Cause in-cluster health-probes and internal chatter must be allowlistable to avoid FPs . Design not stable yet. Testing with 4 real applications on different protocoals.

Lets a `ContainerProfile` allowlist cluster-infrastructure egress/ingress **by Service name** (`serviceRef` / `serviceSelector`) or the reserved **`host`** entity, instead of a broad `ipAddresses` serviceCIDR — which allowlists every ClusterIP on the listed ports and blinds R0011/R0012 to lateral movement.

Single commit. `pkg/networkpeer`:
- **resolve.go** — `serviceRef`/`serviceSelector`/`host` → `(IP,port,proto)` tuples via a `Lister` (informer-backed in prod, fake in tests). Empty selector **fails closed** (never a cluster-wide match-all).
- **expand.go** — storage `NetworkNeighbor` → selector-free `ipAddresses` neighbors (`ExpandServiceNeighbors`) + `WithResolvedServiceNeighbors` CP wrapper; fails closed on `MatchExpressions` / empty `matchLabels`; never mutates input.
- **lister.go** — `InformerLister` over Service/EndpointSlice/Node listers; ClusterIP ∪ endpoints; `host` = the agent's **own** node InternalIP + CNI gateway.

Wiring: `ContainerProfileCache.SetServiceLister` + `WithResolvedServiceNeighbors` before both `Apply()` projection sites; `cmd/main.go` builds the informer factory (bounded `WaitForCacheSync`) and installs the lister. **nil lister = no-op**, so behavior is unchanged until a profile uses the new fields. 18 unit tests.

## Dependencies / relationship
- **Depends on the storage `NetworkNeighbor` schema** (`ServiceRef`/`ServiceSelector`/`Entity`). `go.mod` temporarily `replace`s `kubescape/storage` with the schema commit until that lands upstream.
- Pairs with the peer-selector CEL work (**#902**). Reuses the existing address surface, so it degrades to address-only without the port-aware matcher (**#85**); serviceRef ports are honored once that lands.

## Validation
Unit: `pkg/networkpeer` 18 tests. End-to-end: a component test (`Test_50`) binds a client to a `serviceRef default/kubernetes` profile → **0 R0011** for apiserver egress, while an **uncovered egress fires R0011** (narrow allowlist, detection preserved) — passing in fork CI on the built image.

## Summary by CodeRabbit

* **New Features**
  * Added Kubernetes-aware network neighbor resolution for services, selectors, entities, endpoints, and host nodes.
  * Network policies can now resolve service references into concrete IP addresses, ports, and protocols.
  * Container profiles automatically include resolved network neighbors when cluster information is available.
  * Added fail-closed handling for invalid, empty, or unresolved selectors.


entlein:
Thanks @coderabbitai — addressed 4 of 6, with rationale on the other two:

**Fixed (commit `6d50aff`):**
- **namespaceSelector broadening** — `specFromNeighbor` now fails closed unless the namespaceSelector is exactly `kubernetes.io/metadata.name=<ns>` (MatchExpressions / extra keys / other keys / empty are rejected, not silently dropped). Added the 4 edge-case tests.
- **gatewayIP bounds** — returns `""` when the incremented address falls outside the PodCIDR (`/31`, `/32`, `.255` overflow); added `10.42.0.5/32` and `255.255.255.255/32` cases.
- **cluster-wide Node informer** — Nodes now use a second factory field-selected to `metadata.name=cfg.NodeName`; Services/EndpointSlices stay cluster-wide (a profile may reference any namespace's Service). `HostIPs` keeps its client-side filter as belt-and-suspenders.
- **`fake.NewClientset`** — renamed.

Also, unprompted but related: the informer setup no longer **blocks** startup on cache sync — that bounded wait was delaying node-agent's core startup after a daemonset restart and tripping `Test_22`'s tight completion budget.

**Deferred — re-project on informer changes (endpoint staleness):** real and acknowledged, but it's design-level (a generation counter on the lister, bumped by debounced event handlers, threaded into the reconciler's fast-skip + `nudge`). Scoped to a follow-up PR rather than a review-round patch. The 30s reconcile tick bounds *CP-driven* staleness but not this, so I won't claim otherwise.

**Not applied — Ready-filtering endpoints:** this is an **allowlist** for anomaly detection, not a load-balancer. An endpoint that exists but isn't Ready yet (readiness-probe traffic, a pod draining during rollout, direct pod-to-pod) is still a legitimate peer of the referenced Service; filtering it **narrows** the allowlist → false positives. It also interacts badly with the frozen-at-projection-time behavior above (a momentarily-unready endpoint would be permanently missing). The permissive behavior is the safer failure mode here.

entlein:
@coderabbitai — implemented the lister generation rather than deferring it (`73d5316f`). No follow-up issue needed.

- `networkpeer.Lister` now exposes `Generation()`; `InformerLister` bumps an atomic counter from Service/EndpointSlice/Node event handlers (wired in `cmd/main.go`) — one atomic increment per event.
- `HasServiceNeighbors(cp)` flags profiles that declare serviceRef/serviceSelector/entity, keyed on the **raw** fields, so a profile projected before the caches synced (the timed-out-sync case you flagged) still re-projects once they fill.
- `CachedContainerProfile` records `UsesServiceResolution` + the `ListerGen` it was resolved against; the reconciler fast-skip now re-projects such entries when the generation advances. Non-resolving profiles keep the cheap RV/spec fast-skip; re-projection is coalesced onto the reconcile tick (no per-event work, no nudge hot-loop).

So both sub-cases you raised are covered: endpoint churn *and* caches that fill after the non-blocking startup.

tests/resources/containerprofile-serviceref-network.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: spdx.softwarecomposition.kubescape.io/v1beta1
+kind: ContainerProfile
+metadata:
+  name: serviceref-overlay
+spec:
+  execs:
+  - path: /bin/sleep
+  - path: /usr/bin/curl
+  syscalls:
```
entlein: chicken, we dont need those syscalls here, this is a network test and they wont do anything but FP

tests/resources/serviceref-client-deployment.yaml
```diff
@@ -0,0 +1,21 @@
...
+      app: serviceref-client
+  template:
+    metadata:
+      labels:
+        app: serviceref-client
+        kubescape.io/user-defined-profile: serviceref-overlay
+    spec:
+      containers:
+      - name: curl
+        image: docker.io/curlimages/curl@sha256:08e466006f0860e54fc299378de998935333e0e130a15f6f98482e9f8dab3058
```
entlein: chicken, we said Flux or Argo IRL test... now you give us another curl

tests/component_test.go
```diff
@@ -3906,3 +3906,92 @@ func Test_49_EphemeralContainerFullTreatment(t *testing.T) {
 		return countRuleAlerts(t, ns.Name, "R0001", "ephcon", "id") > 0
 	}, 2*time.Minute, 10*time.Second, "id was not in the ephemeral container's learned profile — it must fire R0001 (detected + alerted like any other container)")
 }
+
+// Test_50_ServiceRefNetworkNeighbor validates the serviceRef selector end to
```
entlein: big words chicken, for what that test really covers... how about we remove the blubber and write a better test?

entlein:
## Performance measurement on a live cluster

Measured rather than reasoned, because two bench-derived conclusions turned out to be wrong (details at the bottom).

### Setup

Single-node kind cluster, held constant across every arm — only the node-agent image/flag changed:

- real Flux `source-controller` reconciling 2 `HelmRepository` CRs every 30s
- k6, 5 VUs against the repo Services
- an EndpointSlice churn driver (10 slices patched every 5s) so the lister-generation path is continuously exercised
- **319 Services / 319 EndpointSlices** cluster-wide, so the cluster-wide informer cost is representative rather than a bare-kind artifact
- a ContainerProfile using `serviceRef` (apiserver, kube-dns) **and** a `serviceSelector` fanning over 300 Services (~900 endpoint IPs) — deliberately harsher than any realistic profile

Method: 150s settle, 240s sample, Prometheus/cAdvisor, arms **alternated** (`B → C → B → C`) so cluster drift is shared rather than attributed to one arm. Feature-on vs feature-off is always compared **within the same image**, never across builds.

### Result: feature costs +4.3 mCPU (+7%) and ~+6 MiB

| | OFF (default) | ON | Δ |
|---|---|---|---|
| node-agent CPU | 60.5 mCPU | 64.8 mCPU | **+4.3 (+7.0%)** |
| node-agent WSS | 265.2 MiB | 271.0 MiB | **+5.7** |
| node-agent RSS | 98.0 MiB | 103.8 MiB | **+5.8** |
| storage CPU / mem | 9.6–9.9 mCPU / 53 MiB | 9.8–9.9 / 53 MiB | none |
| apiserver | 5.4 req/s | 5.4–5.5 req/s | none |

Paired deltas across two alternating repetitions: **+4.2 and +4.3 mCPU** — tighter than run-to-run noise.

Decomposed:

| Component | Cost |
|---|---|
| Informers + cache, gate on but **no profile using the feature** | ~0, indistinguishable from OFF |
| Resolution + re-projection (300-Service selector under churn) | all of the +4.3 mCPU |

So the informer gating and the `SetTransform` that strips `managedFields`/annotations do their job: standing cost is nil, and the feature is **off by default** (`networkServiceResolutionEnabled`). An agent that doesn't use it measures identical to not having it.

### A rejected optimization, reported because the negative result is the useful part

The obvious follow-up was: a generation bump only says *something* moved in the cluster, not that *this* profile's addresses did — so fingerprint the resolved set and skip the rebuild when it's unchanged. Implemented, measured, **dropped**:

| Arm | CPU |
|---|---|
| feature on, no optimization | 69.7, 68.1 |
| feature on, fingerprint skip | 63.9, 68.4 |

Paired deltas **−5.8 and +0.3** — signs disagree, so no reliable win. Mechanism: the expensive half is the *resolution* (two `ServicesByLabels` scans over 319 Services plus 300 endpoint lookups), not the DeepCopy/Apply the skip avoids. Fingerprinting still resolves, so it targets the minority of the cost. Not worth the extra state on the cache entry.

A first version of that optimization was actively **worse** (+11.5 mCPU vs +4.4 for the feature alone) because `InformerLister.ServicesByLabels` returns items in informer map order, so the hash differed on every call and never matched — paying a full resolution on top of the rebuild it failed to avoid. Worth flagging for anyone who revisits this: the unit-test fake sorts its results while the real lister does not, so ordering bugs of this shape pass unit tests and only show up on a cluster.

### Caveats

- n=2 per arm, single node. node-agent is a DaemonSet, so cache cost scales per node.
- **319 EndpointSlices does not exercise the memory concern.** Microbenchmarks project ~15–19 MiB of trimmed cache at 5k slices; that is unmeasured here.
- The load is harsher than realistic — a 300-Service selector re-resolved under 5s churn. A profile naming 1–5 Services would cost a fraction.
- Two earlier readings in this investigation were wrong and were corrected by re-running: an initial "+26% CPU" was a startup artifact from an arm that hadn't finished becoming ready, and a first "no measurable cost" reading was invalid because the workload pod carried the profile label while the ContainerProfile itself had never been created — so the arm did zero resolution work.

go.mod
```diff
@@ -479,3 +479,5 @@ replace github.com/anchore/syft => github.com/kubescape/syft v1.32.0-ks.2
 replace github.com/anchore/stereoscope => github.com/anchore/stereoscope v0.1.9
 
 replace github.com/opencontainers/runtime-spec => github.com/opencontainers/runtime-spec v1.2.1
+
+replace github.com/kubescape/storage => github.com/k8sstormcenter/storage v0.0.240-0.20260823123818-6f3a2ee6385d
```
matthyx, line 483: Blocker: `replace github.com/kubescape/storage => github.com/k8sstormcenter/storage v0.0.240-...` pins to a fork commit, not the upstream `kubescape/storage` module. Per the PR description this is intentional until the `NetworkNeighbor` schema change (`ServiceRef`/`ServiceSelector`/`Entity`) lands upstream — flagging it explicitly so it isn't merged accidentally while still pointing at the fork. This needs to be removed (pointing back at a released `kubescape/storage` tag) before merge.

tests/component_test.go
matthyx, line 1: Blocker: CI is red on this exact commit (8d5a019), and it's this PR's own new test failing — `Test_50_ServiceRefNetworkNeighbor/selector_allowed_no_alert` asserts 0 R0011 alerts for the serviceRef/serviceSelector-allowlisted apiserver/DNS/helm-repo egress, but got 9 (log at component_test.go:4066, run https://github.com/kubescape/node-agent/actions/runs/32690468406/job/97324007809). That means the resolved allowlist isn't actually suppressing alerts for legitimate service egress in this run — the core behavior this PR adds doesn't hold end-to-end yet. `Test_22_AlertOnPartialNetworkProfileTest` also failed on the same run (timed out waiting for the network neighborhood to complete, component_test.go:1345) — worth checking whether the reconciler's new generation-based re-projection path is adding latency/instability there.

matthyx (commented):
Reviewed the `pkg/networkpeer` resolution/expansion/lister code plus the `containerprofilecache`/`cmd/main.go` wiring. The design is sound and CodeRabbit's earlier rounds already caught and got fixes for the main correctness risks (empty-selector/namespaceSelector fail-closed, `gatewayIP` CIDR-containment check, Node informer scoping, non-blocking startup, generation-based re-projection for endpoint churn) — nothing further to add there.

Not approving yet, for reasons independent of that code quality:

1. **This PR is a draft explicitly marked "NOT READY FOR REVIEW"** by the author, stacked on #902.
2. **CI is red on the current head (8d5a019)**: this PR's own new test, `Test_50_ServiceRefNetworkNeighbor/selector_allowed_no_alert`, fails — it expects 0 R0011 alerts for serviceRef/serviceSelector-allowlisted egress but observed 9 — meaning the allowlist isn't actually suppressing alerts for legitimate service traffic in this run. `Test_22_AlertOnPartialNetworkProfileTest` also failed (timeout waiting for network neighborhood completion). See inline comments for run links.
3. **`go.mod` temporarily replaces `kubescape/storage` with a fork commit** (`k8sstormcenter/storage`) pending the upstream schema PR — by design per the description, but it's a hard merge blocker until that lands and this replace is removed.

Happy to take another pass once the draft is promoted, Test_50/Test_22 are green, and the storage dependency points back upstream.

matthyx:
(my review agent picked it sorry)

entlein:
> (my review agent picked it sorry)

dont mention it, I m sorry for this (what I originally thought was a quick tiny PR). Will ping you once the set of the now 4 PRs is in shape, maybe I ll stuff them all in one. We ll see :) 
Happy Monday 😎 !

entlein:
superseded by #923

PR #916, opened 2026-08-24
fix(seccomp): drop unrecognized syscall numbers instead of recording "unknown"

### LINKED ISSUE #910: decodeSyscalls misreads internal state footer byte as syscall bit, injecting bogus "unknown" into every generated profile
## Summary

`decodeSyscalls` in `pkg/containerwatcher/v2/tracers/syscall.go` decodes the raw per-mntns syscall bitmap produced by the vendored Inspektor Gadget `advise_seccomp` eBPF gadget, but it iterates one byte too far and misinterprets an internal state flag as a syscall bit. This injects a bogus `"unknown"` entry into every generated seccomp/application profile.

## Root cause

In the `advise_seccomp` gadget (`gadgets/advise_seccomp/program.bpf.c`), the eBPF map value is:

```c
#define SYSCALLS_COUNT 500
#define SYSCALLS_MAP_VALUE_FOOTER_SIZE 1
#define SYSCALLS_MAP_VALUE_SIZE (SYSCALLS_COUNT + SYSCALLS_MAP_VALUE_FOOTER_SIZE)

struct val_t {
	unsigned char syscalls[SYSCALLS_MAP_VALUE_SIZE]; // 501 bytes
};
```

Byte `syscalls[SYSCALLS_COUNT]` (index 500) is **not** a syscall bit — it's an internal "runc recording started" state flag, set once the program observes `runc`'s `prctl(PR_GET_PDEATHSIG)` call:

```c
if (is_runc) {
	if (syscall_bitmap->syscalls[SYSCALLS_COUNT] == 0) {
		if (id == __NR_prctl && PT_REGS_PARM1(&regs) == PR_GET_PDEATHSIG) {
			syscall_bitmap->syscalls[SYSCALLS_COUNT] = 1;
		}
		return 0;
	}
	...
```

Since virtually every container's `runc` init process trips this flag, byte 500 is set to `1` for essentially every observed container.

`decodeSyscalls` (`pkg/containerwatcher/v2/tracers/syscall.go:141-153`) iterates the **entire** raw buffer, including this footer byte:

```go
func decodeSyscalls(syscallsBuffer []byte) []string {
	syscallStrings := make([]string, 0)
	for i := range syscallsBuffer {
		if syscallsBuffer[i] > 0 {
			syscallName, exist := syscalls.GetSyscallNameByNumber(i)
			if !exist {
				syscallName = "unknown"
			}
			syscallStrings = append(syscallStrings, syscallName)
		}
	}
	return syscallStrings
}
```

Index `i=500` doesn't correspond to any real x86_64 syscall (the highest defined syscall number is ~461), so `syscalls.GetSyscallNameByNumber(500)` fails and the literal string `"unknown"` is appended to the decoded syscall list — and from there flows straight into the generated `SeccompProfile`/`ApplicationProfile` as a bogus allow-rule entry.

## Live reproduction

On a real cluster (kubescape-managed node-agent, backend-storage mode), the generated `SeccompProfile` CR for a workload contained:

```json
"syscalls": [{
  "names": ["accept4", "access", ..., "umask", "uname", "unknown", "unlinkat", ...],
  "action": "SCMP_ACT_ALLOW"
}]
```

`"unknown"` is not a valid syscall name on any architecture and has no meaning to the kernel/runc when the profile is compiled into a BPF filter.

## Impact

- Every generated profile that ever observed a `runc`-attributed syscall (i.e. essentially all of them) carries this spurious `"unknown"` entry.
- We confirmed via live testing that removing `"unknown"` from the profile alone does **not** by itself fix a separately-observed enforcement issue on the same workload, so on the runtimes we tested, the bogus entry appears to be tolerated/ignored by the seccomp compiler rather than corrupting the whole rule — but this is undocumented/version-dependent behavior of the downstream OCI runtime, not something to rely on. On stricter or different seccomp-profile loaders, an unresolvable syscall name could plausibly cause the whole rule (or the whole profile) to be rejected.
- More generally, this pollutes generated profiles with meaningless data, which is confusing when inspecting/auditing generated seccomp profiles and wastes an entry slot.

## Suggested fix

In `decodeSyscalls`, exclude the footer byte from iteration, e.g.:

```go
for i := 0; i < len(syscallsBuffer)-1; i++ { // skip the SYSCALLS_COUNT footer flag byte
```

or equivalently bound the loop to `SYSCALLS_COUNT` explicitly (matching the gadget's own `SYSCALLS_COUNT` constant) rather than `len(syscallsBuffer)`, so the internal state byte is never treated as syscall data.

## Environment

- `kubescape/node-agent` v0.3.179 (vendored)
- IG `advise_seccomp` gadget (`github.com/matthyx/inspektor-gadget` fork, pinned via replace directive)
- Reproduced on a DigitalOcean K8s cluster, containerd 2.2.3, kernel 6.12.96, x86_64

rotemamsa:
Closes #910

## Problem

`decodeSyscalls` names every syscall number it cannot resolve `"unknown"`:

```go
syscallName, exist := syscalls.GetSyscallNameByNumber(i)
if !exist {
    syscallName = "unknown"
}
```

That string is not a diagnostic. It is recorded as a syscall name in the container profile, merged into the application profile, and copied verbatim into the seccomp profiles we generate from that profile. A generated `SeccompProfile` CRD then contains `unknown` in its `SCMP_ACT_ALLOW` name list, where no kernel can match it.

Seen on a live POC cluster, in an applied profile:

```yaml
names:
  - times
  - umask
  - uname
  - unknown      # <- this
  - unlinkat
```

It also makes the profile permanently look wrong in the product: a syscall that is allowed but can never be observed in use keeps a workload from reaching the "optimized" seccomp state.

## Fix

Skip the entry and log the number instead. Profiles then only ever contain names the kernel can match, and the unresolved number is still visible for debugging.

## Test

`TestDecodeSyscallsSkipsUnrecognizedNumbers` picks a syscall number this build cannot resolve plus one it can, and asserts the resolvable name survives while no placeholder is emitted.

Verified on linux/amd64:

```
=== RUN   TestDecodeSyscallsSkipsUnrecognizedNumbers
--- PASS: TestDecodeSyscallsSkipsUnrecognizedNumbers (0.00s)
ok  	github.com/kubescape/node-agent/pkg/containerwatcher/v2/tracers	0.208s
```

With the fix reverted the test fails, so it covers the bug:

```
Error: "[read unknown]" should have 1 item(s), but has 2
```

AI-skills: armosec-shared-rules:docs_find | cmds: /model

## Summary by CodeRabbit

* **Bug Fixes**
  * Unrecognized system calls are no longer displayed as `"unknown"`; they are omitted from decoded results.
  * Added debug logging to identify skipped system call numbers.

### DIFF
--- a/pkg/containerwatcher/v2/tracers/syscall.go
+++ b/pkg/containerwatcher/v2/tracers/syscall.go
@@ -2,6 +2,7 @@ package tracers
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/inspektor-gadget/inspektor-gadget/pkg/datasource"
 	gadgetcontext "github.com/inspektor-gadget/inspektor-gadget/pkg/gadget-context"
@@ -140,20 +141,24 @@ func (st *SyscallTracer) callback(event *utils.DatasourceEvent) {
 
 func decodeSyscalls(syscallsBuffer []byte) []string {
 	syscallStrings := make([]string, 0)
+	// Syscall numbers this build cannot resolve to a name are skipped. Recording a
+	// placeholder name would propagate into the application profile and from there
+	// into generated seccomp profiles, where it is not a valid syscall name. The
+	// numbers are collected and logged once per decode to keep this loop cheap.
+	var skipped []int
 	for i := range syscallsBuffer {
 		if syscallsBuffer[i] > 0 {
 			syscallName, exist := syscalls.GetSyscallNameByNumber(i)
 			if !exist {
-				// The syscall number is not known to this build. Recording a placeholder
-				// name would propagate into the application profile and from there into
-				// generated seccomp profiles, where it is not a valid syscall name.
-				// Drop it and keep the number for diagnostics.
-				logger.L().Debug("decodeSyscalls - skipping unrecognized syscall number",
-					helpers.Int("syscallNumber", i))
+				skipped = append(skipped, i)
 				continue
 			}
 			syscallStrings = append(syscallStrings, syscallName)
 		}
+	}
+	if len(skipped) > 0 {
+		logger.L().Debug("decodeSyscalls - skipped syscall numbers with no known name",
+			helpers.String("syscallNumbers", fmt.Sprint(skipped)))
 	}
 	return syscallStrings
 }
--- a/pkg/containerwatcher/v2/tracers/syscall.go
+++ b/pkg/containerwatcher/v2/tracers/syscall.go
@@ -144,7 +144,13 @@ func decodeSyscalls(syscallsBuffer []byte) []string {
 		if syscallsBuffer[i] > 0 {
 			syscallName, exist := syscalls.GetSyscallNameByNumber(i)
 			if !exist {
-				syscallName = "unknown"
+				// The syscall number is not known to this build. Recording a placeholder
+				// name would propagate into the application profile and from there into
+				// generated seccomp profiles, where it is not a valid syscall name.
+				// Drop it and keep the number for diagnostics.
+				logger.L().Debug("decodeSyscalls - skipping unrecognized syscall number",
+					helpers.Int("syscallNumber", i))
+				continue
 			}
 			syscallStrings = append(syscallStrings, syscallName)
 		}
--- a/pkg/containerwatcher/v2/tracers/syscall_decode_test.go
+++ b/pkg/containerwatcher/v2/tracers/syscall_decode_test.go
@@ -1,0 +1,49 @@
+package tracers
+
+import (
+	"testing"
+
+	"github.com/inspektor-gadget/inspektor-gadget/pkg/utils/syscalls"
+	"github.com/stretchr/testify/assert"
+)
+
+// TestDecodeSyscallsSkipsUnrecognizedNumbers verifies that syscall numbers this build
+// cannot resolve to a name are dropped rather than recorded under a placeholder name.
+// A placeholder propagates into the application profile and from there into generated
+// seccomp profiles, where it is not a valid syscall name.
+func TestDecodeSyscallsSkipsUnrecognizedNumbers(t *testing.T) {
+	// Find a syscall number that this build cannot resolve, to stand in for a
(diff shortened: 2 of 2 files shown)


rotemamsa:
Root cause found while investigating a related data-loss issue, worth recording here: the `"unknown"` entry is not (only) an exotic syscall number. The `advise_seccomp` gadget's map value is `SYSCALLS_COUNT + 1` bytes — index 500 is a **footer flag** the eBPF sets when it starts recording a runc init (`syscalls[SYSCALLS_COUNT] = 1` after `prctl(PR_GET_PDEATHSIG)`). `decodeSyscalls` iterates the full buffer, hits index 500, finds no name, and emitted `"unknown"`.

Because the 30s map fetch uses `BPF_MAP_LOOKUP_AND_DELETE_BATCH`, the footer byte appears exactly once per container lifetime: in the first drained packet after `runc create`. So `"unknown"` in a profile is actually a marker that the first drain (the one carrying all container-startup syscalls) survived the pipeline.

This PR's fix handles it correctly either way: index 500 resolves to no name and is skipped. The debug log will fire once per container first-drain with `syscallNumber=500`, which is expected and harmless.

matthyx (approved):
Reviewed the change and rotemamsa's follow-up comment about the `advise_seccomp` footer byte (index 500).

- Logic is correct: unresolved syscall numbers are now skipped and logged instead of leaking a bogus `"unknown"` name into the profile/seccomp pipeline. Confirmed no other code path relies on the old `"unknown"` placeholder from `decodeSyscalls`.
- Built `pkg/containerwatcher/v2/tracers/...` and ran `TestDecodeSyscallsSkipsUnrecognizedNumbers` locally — passes, and the test meaningfully covers both the skip and the "known syscalls still survive" cases.
- CI: the only failing job (`Test_28_UserDefinedNetworkNeighborhood/unknown_domain_R0005`) fails with `write metadata: insert metadata: sqlite: step: database is locked` — a test-harness infra flake unrelated to this change, not a regression from this PR.

No blockers. LGTM, approving.

PR #364, opened 2024-09-11
Feature/private

amitschendel:
## Overview
This PR exposes some interfaces to use the node-agent as a pkg.

Issue #908, opened 2026-08-20
Wait, retry, assert vs require conditions in component tests not necessarily meaningful, investigate more robust conditions.


entlein:
# Description
CTs sometimes fail for (mostly its timing issues), and the tests are red, even though the functionality is green. This is wasteful 

# Environment
github hosted runners, see config in .github (stable for last year)
 
# Steps To Reproduce
run the CTs, see that some sometimes fail

# Expected behavior
CTs should fail when what the tests assert is broken

# Actual Behavior
CTs sometimes fail due to `flakyness` and rerunning the tests makes them green. 

# Additional context
It might be simply helpful to categorize which `types` of flakyness we currently have. A `zoology of test-devils`, so to say.

This bug is not high prio, just me being impatient with waiting for test runs

BTW: the benchmark has similar issues

ConstanzeTU:
Data point from a component-test stabilization pass: a recurring class of "rotating" single-test failures (a different test each run, green in adjacent runs) traced to the tests' own direct storage creates eating a transient 5xx — the storage backend fails fast on SQLite write-lock contention by design (kubescape/storage#365, damage contained by kubescape/storage#366), and real in-cluster writers requeue, but the test clients had no retry at all.

What made the conditions robust in practice:
- one retrying round-tripper on the shared test clientset (replay on 5xx/connection error) instead of per-call-site retry loops — creates are safe to replay because the backend's write path is atomic since #366;
- alert-presence oracles must tolerate AlertManager ingestion lag: a single-shot count comparison right after an exec attributes lag as a product failure — an Eventually/Never window around the count is the meaningful condition;
- absence oracles based on AlertManager are unreliable across test phases (resolved alerts linger in retention); asserting on the system's own state objects or agent log markers held up better.

With those three applied the matrix failure set stopped rotating.

PR #923, opened 2026-08-24
feat(network): symmetric R0011/R0012 with NetworkPolicy-style internal allowlisting and port-aware alerting

entlein:
# Ports and NW Policies - twin rules for internal/external nw

replacing the partial PRs #902, #905, #915 (storage companion: kubescape/storage#364).

##  Feature: Ports alert, Internal and External Network Slices can be selected for allowlisting

**Rules — symmetric egress/ingress, no IP-class gate**
- R0011 (egress) and new **R0012 (ingress)** are exact twins: internal and external peers treated alike. Lateral movement to an unlisted internal peer alerts; a serviceCIDR-wide allowlist entry no longer blinds detection.
- Both consume the port-aware matcher `was_address_port_protocol_in_*`: an allowed address on a violated port alerts; an entry with no ports means any port, computed per-(port,protocol).

**Allowlisting internal traffic, NetworkPolicy-style**
- Peer selectors: `podSelector` on profile neighbors, matched at event time (`was_selector_in_*`)
- Service references: `serviceRef{Name}`, `serviceSelector`, `entity: host` resolve at projection time to ClusterIP + endpoint IPs + the Service FQDN, feeding the ordinary address/DNS surfaces. Unresolvable selectors contribute nothing — never a match-all. Informers are gated behind `networkServiceResolutionEnabled` (default ON) and strip managedFields/annotations; 

measured cost MUST BE REMEASURED

## Validation
- Full component-test matrix is now green , please check Test53 specifically

## Dependency
`go.mod` temporarily replaces `kubescape/storage` with the fork branch of kubescape/storage#364 (schema fields + generated code + loss-guards in collapse/deflate/NetworkPolicy generation). After #364 merges, the replace drops for a pseudo-version pin — no storage release required.

## Blocker before merge
must switch the rules OFF by default, even though Test53 is to show how many FalsePositives a user would get if it were turned on. 

### Difficulties encoutnered
Matching the host - ip turned out to be difficult, so its not used by default (its enabelable) as its costly

### Not changed
Rule Definitions and severity (and mitre fields) need to be aligned with whenever we update the `rulelibrary`, for now this feature must be considered depended on a new rulelibrary release, as `R0012` does not exist yet.

## Summary by CodeRabbit

* **New Features**
  * Added Kubernetes-aware resolution for service references, selectors, DNS names, and host peers.
  * Added port- and protocol-aware ingress and egress network matching.
  * Added destination namespace and pod-label data for network rules.
  * Enabled network service resolution by default, with configurable host-peer alerts.
  * Added support for service-based network policies and loopback traffic learning.
  * Added the Unexpected Ingress Network Traffic alert to default rules.

* **Bug Fixes**
  * Corrected wildcard and literal port handling, including port `0`.
  * Prevented invalid or unresolved selectors from broadening access.
  * Improved profile updates as service endpoints and network topology change.


charts/kubescape-rules/templates/binding.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: kubescape.io/v1
```
entlein: chicken, please remove those here, thats the PR to rulelibrary, but add them to the tests/** folder

tests/chart/templates/node-agent/default-rules.yaml
```diff
@@ -313,7 +313,7 @@ spec:
           uniqueId: "event.dstAddr + '_' + string(event.dstPort) + '_' + event.proto"
           ruleExpression:
             - eventType: "network"
-              expression: "event.pktType == 'OUTGOING' && !net.is_private_ip(event.dstAddr) && !cp.was_address_in_egress(event.containerId, event.dstAddr)"
+              expression: "event.pktType == 'OUTGOING' && !event.dstAddr.startsWith('127.') && event.dstAddr != '::1' && !cp.was_address_port_protocol_in_egress(event.containerId, event.dstAddr, event.dstPort, event.proto) && !cp.was_selector_in_egress(event.containerId, event.dstNamespace, event.dstPodLabels)"
```
entlein: Will make the rules broad and OFF by default. I.e. the private ip and localhost filters will be removed

tests/resources/serviceref-k6.yaml
```diff
@@ -0,0 +1,43 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: k6-script
+data:
+  load.js: |
+    import http from 'k6/http';
+    import { sleep } from 'k6';
+    export const options = { vus: 2, duration: '30m' };
```
entlein: remove this

matthyx (changes requested):
Reviewed as requested. This lines up with the author's own "NOT READY FOR REVIEW" flag — the `networkpeer` resolution/fail-closed design (empty selectors → nil, `MatchExpressions` rejected, `ClusterIPNone` excluded, gateway IP verified via `ipNet.Contains`) is sound, but there are concrete blockers before this can merge:

**Critical**
- `go.mod`/`go.sum`: `replace github.com/kubescape/storage => github.com/k8sstormcenter/storage v0.0.240-...` pins the module to a fork, with the pre-replace requirement left at the placeholder `v0.0.0-00010101000000-000000000000`. This is unbuildable for any consumer until kubescape/storage#364 merges and a real pseudo-version replaces it.
- Unrelated dependency downgrades pulled in alongside the fork pin: `kubescape/backend v0.0.39 → v0.0.31` and `gotest.tools/v3 v3.5.2 → v3.5.0`. These look like tidying against the stale fork rather than an intentional change — please restore both unless there's a reason to downgrade.

**High**
- `pkg/objectcache/containerprofilecache/reconciler.go` (~L499-510): the cache-invalidation fix reads `listerGen()` *after* `Apply(...)` runs, in two separate calls. If a `Bump()` happens during resolution, the new generation gets stamped onto IPs resolved against the *old* view, so `refreshOneEntry`'s staleness check never fires and the stale result is served from the LRU indefinitely — the exact bug this fix was meant to close. `containerprofilecache.go` (~L618-620) does this correctly by reading the generation once, before resolution — please match that pattern here.
- The widened R0011 (drops `net.is_private_ip`) and new R0012 ship `enabled: true` in `default-rules.yaml`/binding, but `EnableNetworkServiceResolution` defaults to `false` in `pkg/config/config.go` — only the test chart's `values.yaml` turns it on. On upgrade, existing profiles start alerting on all internal peers with no serviceRef/serviceSelector allowlisting available to quiet them (resolution is off by default). Either default resolution on alongside the widened rules, or ship R0012/the widened R0011 disabled until resolution is on by default.
- `pkg/networkpeer/expand.go` (~L127): nil `namespaceSelector` on a `serviceSelector` resolves cluster-wide ("cluster-wide by design" per the comment), which contradicts both the PR description ("nil namespaceSelector = same namespace, as in NetworkPolicy") and the same-namespace default used elsewhere (`network.go`'s `namespaceSelectorMatches`). `resolveServices` doesn't have the profile's namespace in scope to fix this today — `serviceSelector: {app: foo}` currently allowlists that label across every namespace in the cluster, which is a much broader allowlist than the feature intends.
- RBAC (`discovery.k8s.io/endpointslices`) only lands in `tests/chart`'s ClusterRole. Per the PR description, the shipped chart lives in kubescape/helm-charts and "needs to be moved" — without a companion PR there, real deployments enabling this feature get informer 403s and silently degrade to ClusterIP-only resolution (the exact failure mode already found and fixed once during this PR's own validation).

**Medium**
- `expand.go`'s `hasServiceFields` accepts a `ServiceRefNamespace`-only neighbor, but `specFromNeighbor` rejects it — causes permanent re-projection churn for that shape.
- `pkg/utils/cel.go`'s `dstPodLabels` returns a raw `map[string]string` while sibling CEL accessors return `celtypes`-wrapped values — worth double-checking this doesn't break type coercion in CEL expressions that consume it.

No fork images found in shipped config (only a `ghcr.io/fluxcd` reference in a test fixture, which is fine). Given the go.mod fork dependency alone, this isn't mergeable yet — requesting changes rather than approving.

entlein:
thanks for catching the regression during the merge/rebase

**AI, **
please
`pkg/objectcache/containerprofilecache/reconciler.go (~L499-510): the cache-invalidation fix reads listerGen() after Apply(...) runs, in two separate calls. If a Bump() happens during resolution, the new generation gets stamped onto IPs resolved against the old view, so refreshOneEntry's staleness check never fires and the stale result is served from the LRU indefinitely — the exact bug this fix was meant to close. containerprofilecache.go (~L618-620) does this correctly by reading the generation once, before resolution — please match that pattern here.`
-> recheck and fix

Clusterrole RBAC: Thanks for bringing it up -> AI,  check how much surface this introduces to the deaomnset overall and align with how e.g. cilium minimizes that role

`EnableNetworkServiceResolution defaults to false in pkg/config/config.go ` : we assume that noone is currently using those profiles, the default `loud` is desired, else a TDD approach to find/scan for all the connections to add to a profile is difficult.
That being said: you are correct that the mechanism is inconsistent and silly ATM: better to turn it on in ...config.go and ship the rules default OFF
AI, please lets flip config to ON (or even ditch the feature-flag) and we turn the rule OFF (not remove internal IPs, that defeats the purpose, just ship the rules R0011/R0012 as OFF)

fork dependencies: for the CTs to run, AI, please patch storage-tag.sh with our fork ref, else it builds against upstream which doesnt have the change. That defeats the test

**Human readers,**
Bringing us to point of how to split the rules: the whole slew of changes here break a lot of default kubescape behavior IF people use profiles at all (again: probably noone is): things will be default loud, not default learn.

I m also adding an alert on `new Container` , such that any workload not known to kubescape will alert instead of learn - that completely flips the basic node-agent UX -> 
Possibly there should be just one global flag in the `helm values.yaml` to toogle all user-defined-profiles to on. Not sure yet.

Reason this is in Draft: 
- turning on these rules with bad profiles causes massive floods of alerts that may even DoS a cluster, if RuleCoolDown is disabled (my default setting), as every internal connection from e.g. DB to prometeus is going to alert ... yes.



Namespaces: That is correct: namespaces are not read - meaning, if there is a collision in names across namespaces they act as allowlists -> I think, that should be how it works.

Exec summary : 🙏 thanks for catching the technical bugs again (this is embarassing) while I sort out my brain

pkg/rulemanager/cel/libraries/containerprofilenetwork/selector_test.go
```diff
@@ -0,0 +1,63 @@
...
+			}
+		})
+	}
+}
+
+func TestWasSelectorInPeers_EmptySelectorAndEmptyLabels(t *testing.T) {
+	profileNs := "redis"
+	emptySelector := []objectcache.PeerSelector{{PodSelector: &metav1.LabelSelector{}}}
+
+	if !wasSelectorInPeers(emptySelector, labels.Set{}, "redis", profileNs) {
```
entlein: No chicken, it should NOT. Rework this into:
If no name is supplied : no match

Namespaces can only ever be used for disambiguation or performance reasons (to prioritize the search in the local namespace for large clusters). In doubt: the NS selector  could also be dropped entirely, if this just causes too many edge-cases.

pkg/rulemanager/cel/libraries/containerprofilenetwork/selector_test.go
```diff
@@ -0,0 +1,63 @@
...
+	if ns != nil {
+		p.NamespaceSelector = &metav1.LabelSelector{MatchLabels: ns}
+	}
+	return p
+}
+
+func TestWasSelectorInPeers(t *testing.T) {
+	// Peer identity as IG's kubeipresolver stamps it onto the event: a namespace
+	// and pod labels, resolved cluster-wide. No IP, no local pod lookup.
+	podLabels := labels.Set{"app": "redis-client"}
```
entlein: we need to cover services and hosts as well in the tests. Add all possible combinations test for the k8s-api in resolve_test.go

pkg/rulemanager/cel/libraries/containerprofilenetwork/selector_test.go
entlein: Note to reviewers: yes, this means that a foreign pod can come in with a prefab label and allowlist itself. 

This is what the signatures will then disable, but without them, yes, thats a feature

entlein:
Here s the benchmakr from this run BTW @matthyx : could have sworn that  benchmark-thingy posted to the PR for the last round, now the widget is getting 403. Possibly intentional, possibly not. 
BTW: I think, the benchmark is wrong, but here we are:

EDIT: entlein does not believe that her nonsense `improved` the perf :) 


  eBPF Dedup Benchmark Results


  Node-Agent Resource Usage
  -------------------------------------------------------
  Metric                      BEFORE       AFTER       Delta
  -------------------------------------------------------
  Avg CPU (cores)              0.218       0.213       -2.0%
  Peak CPU (cores)             0.231       0.230       -0.6%
  Avg Memory (MiB)           374.079     297.109      -20.6%
  Peak Memory (MiB)          377.723     302.402      -19.9%

  Dedup Effectiveness: no data available

matthyx:
@entlein I think the benchmark is accurate on `main`, I have fixed my IG fork to use the latest deepcopy implementation with a pool... we had a more efficient implementation that was never used for months, my bad.

entlein:
NICE!

entlein:
Note: Test20 really needs a fluke-fix, AI, be nice and check it AGAIN , please. 

53: we need a concise side-by-side of the current UX (which must be preserved in this PRs default settings) vs the new functionality 
Anticipate Test 52 must be aligned with the behavior (sign `on` AND `off`)

**Hard rule: The current default `learning->enforce` UX must be preserved if default settings are used.**

pkg/containerprofilemanager/v1/container_data.go
entlein: We need to remove this guard here, and do another pass for various aliases that localhost can masquerade as. Add to unit test please

entlein:
@coderabbit : review this PR now, please

We have two open caveats
- we are only testing the network capture on iptables CNI
- for partially learnt profiles, the loopback from host seems to FP occasionally

tests/chart/templates/node-agent/configmap.yaml
```diff
@@ -14,6 +14,7 @@ data:
         "prometheusExporterEnabled": {{ eq .Values.nodeAgent.config.prometheusExporter "enable" }},
         "runtimeDetectionEnabled": {{ eq .Values.capabilities.runtimeDetection "enable" }},
         "networkServiceEnabled": {{ eq .Values.capabilities.networkPolicyService "enable" }},
+        "networkServiceResolutionEnabled": {{ if hasKey .Values.nodeAgent.config "networkServiceResolution" }}{{ .Values.nodeAgent.config.networkServiceResolution }}{{ else }}true{{ end }},
```
entlein, line 17: we really need this feature flag? test if it even works

tests/chart/values.yaml
```diff
@@ -32,8 +32,8 @@ global:
 storage:
   name: "storage"
   image:
-    repository: quay.io/kubescape/storage
-    tag: v0.0.156
+    repository: ghcr.io/k8sstormcenter/storage
+    tag: net-v2-rc1
```
entlein, line 36: thats an ancient image

entlein:
Asked the AI to review those CTs that have been fluking in this round , maybe we can address some of https://github.com/kubescape/node-agent/issues/908 , prob its just a re-run.

entlein:
@matthyx : Happy for your thoughts on this at this point :)

tests/chart/templates/node-agent/default-rules.yaml
```diff
@@ -329,6 +329,31 @@ spec:
           - "network"
           - "anomaly"
           - "networkprofile"
+      - name: "Unexpected Ingress Network Traffic"
+        enabled: true
```
matthyx, line 333: @entlein you still say it must be OFF by default?

tests/chart/templates/node-agent/clusterrole.yaml
```diff
@@ -11,6 +11,9 @@ rules:
 - apiGroups: [""]
   resources: ["events"]
   verbs: ["list", "watch", "create"]
+- apiGroups: ["discovery.k8s.io"]
```
matthyx, line 14: please open a PR in helm-charts for the same change

matthyx (changes requested):

pkg/rulemanager/cel/libraries/containerprofilenetwork/network.go
```diff
@@ -210,12 +234,136 @@ func (l *containerProfileNetworkLibrary) wasAddressPortProtocolInIngress(contain
...
+// carries (the form these profiles use). A nil selector matches only the
+// profiled workload's own namespace: the learned generator omits the selector
+// exactly for same-namespace peers, and NetworkPolicyPeer gives an absent
+// namespaceSelector the same meaning. Selectors keyed on other namespace
+// labels are not resolved here.
+// namespaceSelectorMatches: a nil namespaceSelector does NOT consult the
+// namespace — matching is on pod labels alone, and namespace is only used to
+// disambiguate a label collision (an explicitly-set selector). profileNs is
+// unused now but kept in the signature for the collision case.
+func namespaceSelectorMatches(sel *metav1.LabelSelector, ns, profileNs string) bool {
```
matthyx: This flips the doc block right above it (and `projection_types.go`'s `Namespace` field doc, line 105) without changing the behavior — a nil `namespaceSelector` still matches *any* namespace, it just now says that's intentional.

But `getNamespaceMatchLabels` (network_helpers.go:132-139) returns nil exactly for same-namespace learned peers. So every learned same-namespace peer becomes a cluster-wide allowlist by pod labels: a pod I control in a different namespace with matching labels is now allowlisted too. That's the opposite of the lateral-movement protection described in the PR ("a serviceCIDR-wide allowlist entry no longer blinds detection").

Was this flip intentional? If so please reconcile it with `expand.go:140` and `lister.go:62-70` (also cluster-wide on nil selector) and drop the now-dead `profileNs` param. If not, this needs to resolve nil to same-namespace as the original comment (and `projection_types.go`) still says.

pkg/objectcache/containerprofilecache/containerprofilecache.go
```diff
@@ -579,9 +609,21 @@ func (c *ContainerProfileCacheImpl) buildEntry(
...
+	// Project under the current spec, resolving any serviceRef/entity network
+	// neighbors to concrete IPs first. Record whether this profile depends on
+	// the live cluster view and the lister generation it was resolved against,
+	// so the reconciler re-projects it when that view changes.
 	spec := c.snapshotSpec()
-	projected := Apply(spec, userMerged, tree)
+	// Resolution-dependence keys on the profile's OWN neighbors, before the
+	// synthetic host peer is injected (which resolves to the stable local node IP
+	// at projection time and must not force re-projection on unrelated lister bumps).
+	entry.UsesServiceResolution = networkpeer.HasServiceNeighbors(userMerged)
```
matthyx: `UsesServiceResolution` is computed from the profile's own neighbors before `WithHostPeer` injects the synthetic host peer (same pattern in reconciler.go:504). But `main.go:371` intentionally hands the lister over without blocking on cache sync.

If a profile's only cluster-dependent peer is the host one (the common case with `alertOnHostPeers=false`, the default) and projection runs before the Node lister has synced, `HostIPs()` comes back empty and the host peer never resolves. Since `UsesServiceResolution` is false, the fast-skip at reconciler.go:429 means it's never retried even once the lister fills in — R0012 keeps alerting on kubelet probes / node traffic until something unrelated forces a spec rebuild. Same gap on a node IP change or late PodCIDR assignment.

Might need `!cfg.AlertOnHostPeers || c.serviceLister != nil`-style handling folded into `UsesServiceResolution`, or resolve the host peer eagerly instead of deferring to the next reconcile.

matthyx (changes requested):
Two more from the re-review pass (leaving the go.mod/storage-fork item as-is since that's already tracked).

tests/chart/templates/node-agent/default-rules.yaml
entlein: Yes, we ll switch it off, when we actually release it. Its just hard to test if they are off :) 
I think, that ll go into the rulelibrary PR, right?

matthyx: yes

pkg/rulemanager/cel/libraries/containerprofilenetwork/network.go
entlein: So, Namespaces are `not` used in `networkpeers` unless needed for disambiguation 

Maybe, I should drop them entirely.  I actually think,  I need to:

The thing is that once profiles are signed, (and they do `not depend on ns`) , someone might move a profile with a nw-peer name to a different namespace (which is allowed), the signature is still valid, and the detection still works. 
This actually doesnt allow lateral-movement, it just allows a service/pod/host of nameA to move from namespaceA to namespaceB and be allowlisted. 

AI: `please do a side-by-side` of testing this supposed `namespace move` using both the `sign ON` and `sign OFF` charts - everything else equal. To confirm or disconfirm the following hypothesis:

- AppA is in namespace A allowlisting each type of internal selector we have in this PR `without using ns as selector` from AppB in namespace B
- Now, AppB is deployed to namespace C -> nothing should change in terms of profiles of detection 
- Then, AppEvil appears using identical selectorname as AppB  in NamespaceB
- Then, AppEvil appear using identical selectorname as AppB in NamespaceE

Desired outcome:
- the risk profile for AppEvil should be identical no matter the Namespace
- using signatures, AppEvil should not be able to fake sign the identicalselectorname as AppB

Lets add this as test (the signature chart is currently fork only, so run the sign-OFF here, and the sign-ON on the fork CT)

Then, we know if the whole namespace discussion even makes sense.

tests/resources/nsinv-client-other.yaml
```diff
@@ -0,0 +1,16 @@
...
+kind: Deployment
+metadata:
+  name: nsinv-client
+  labels: {app: nsinv-other}
+spec:
+  replicas: 1
+  selector: {matchLabels: {app: nsinv-other}}
+  template:
+    metadata:
+      labels: {app: nsinv-other}
```
entlein: you need TWO of the SAME selector, my dear. This is trivial

entlein:
**Re: namespace-invariant peer matching / can a same-selector app masquerade as an allowlisted peer**

The concern is real, but the defense doesn't live in `namespaceSelector` — it lives in the **admission-class signature**, and the two aren't orthogonal: they overlap through the RBAC/key-custody that decides *which key may sign the `admission` fragment class*. A peer is only allowlisted if an `admission`-class overlay says so, and that overlay is only honored if signed by a key the trust policy trusts for `admission`. An attacker who copies a legit peer's `podSelector` still can't get admitted, because they can't sign the overlay that would add them.

I built a truth table for exactly this masquerade and ran it end-to-end on the fork signature CT (bundle-signing is fork-only, so it can't run here on #923 — it's committed as **disabled**, with these results embedded as commentary + reproducible fixtures). One variable: the key that signs the `admission` overlay adding `app:masq-client`. Same selector every row.

| overlay signed by | composite root | admitted | R0012 on server |
|---|---|---|---|
| `ct-operator` (trusted admission key) | `dce998…` | **yes** | **0** (allowed) |
| `ct-untrusted` (not in policy) | `2bbc47…` | no | **10** (blocked) |
| `ct-vendor` (base key, wrong class) | `2bbc47…` | no | **10** (blocked) |
| unsigned | `2bbc47…` | no | **10** (blocked) |

The three rogue rows collapse to the **identical composite root** — the masquerade fragment is dropped as a never-admitted non-member (`AssembleAndVerifyPartial`: *"signer not permitted for class admission"* / *"fragment is not signed"*), leaving `base` + the benign baseline. R0012 then fires on the server for the unlisted peer and stays silent for the genuinely-allowlisted one.

**Implication for the `namespaceSelector` semantics:** signed admission fragments are inherently namespace-portable (a signed fixture can't name a runtime namespace), so the safe model is exactly the identity-based / namespace-agnostic one — the impersonation boundary is the signature + RBAC key-custody, not the namespace. That argues for dropping the `profileNs` param rather than reverting nil to same-namespace, which I'm doing in a follow-up.

*(The disabled test also has a harness-side timing race — the bundle composite's ingress projection lags the exec sentinel, so R0012 landed ~6s past the poll window; the alerts themselves are real, 10 per rogue namespace. Both reasons it stays fork-only-and-disabled are documented at the test.)*

pkg/rulemanager/cel/libraries/containerprofilenetwork/network.go
entlein: Thanks AI, great first step. but you didnt address if we keep the disambiguation or drop the (currently inconsistent) implemenation in expand/lsiter .

entlein:
@matthyx — on dropping `namespaceSelector` entirely vs keeping it for disambiguation. My recommendation: **keep it** (nil = cluster-wide default + explicit selector for disambiguation), i.e. where the follow-up commit left it after removing the dead `profileNs` param. Reasoning from the evidence:

**What the masquerade table proved, and what it didn't.** The signature gates *authoring* — a rogue peer can't be *added* to an allowlist without the trusted `admission` key (the table above). It does **not** cover the *runtime label-copy*: the matcher does pure label matching with no peer-side signature check (`network.go` `wasSelectorInPeers`), so a cluster-wide `podSelector: app:frontend` is matchable by *any* pod wearing `app:frontend`, in any namespace.

**`namespaceSelector` is the only runtime scoping for that gap.** It resolves as an equality on `kubernetes.io/metadata.name` — the auto-label a pod carries iff it lives in that namespace. To match `namespaceSelector: metadata.name=prod`, the attacker's pod must *be in* `prod`, which needs RBAC to create pods there. So it's an RBAC-backed blast-radius limit, complementary to the signature (authoring) defense — not redundant with it.

| | keep (nil=cluster-wide + explicit) | drop entirely |
|---|---|---|
| runtime label-copy | explicit selector scopes it (RBAC-backed) | widened to cluster-wide, no runtime control |
| NetworkPolicy parity (the PR's model) | preserved | diverges |
| cost when unused | zero — nil is the default the learned generator emits | — |

**Your points are valid** — it's a partial impl (metadata.name equality only), the learned generator never emits it, and signed portable overlays can't pin a namespace name. But "unused by default" ≠ "remove": dropping it deletes a real, RBAC-backed capability and adds **no** security (it removes a scoping knob, it doesn't harden anything), while breaking NP parity.

**What would flip me to drop:** add peer-side signature verification at *match* time (verify the peer pod's own identity, not just who authored the allowlist). That closes the runtime label-copy gap and makes `namespaceSelector` genuinely redundant → drop it then. Until that exists, it's the only thing between an allowlisted identity and a cluster-wide label-copier.

Disambiguation is covered and passing locally — `TestWasSelectorInPeers_NamespaceDisambiguation` (identical labels in two namespaces; nil matches both, `metadata.name=prod` matches only prod) plus the `ns-scoped peer, right/wrong ns` rows in the CEL-eval truth table.

matthyx:
@entlein — the masquerade-table work is solid evidence for the *authoring* gate, but I don't think it settles the `namespaceSelector` question, and I'd like to propose a middle option rather than nil=cluster-wide as the permanent default.

**Why the signature argument doesn't cover this.** The admission-signature system is fork-only and currently committed as disabled (per your own note on the masquerade CT) — it isn't part of #923. So for the code actually shipping here, there's no authoring-side gate at all. `namespaceSelector` is the *only* protection surface that exists in this PR against the label-copy path (a pod anywhere adopting an allowlisted peer's labels). Standardizing on nil=cluster-wide gives that up before the thing that's supposed to replace it has landed.

**Why this isn't an edge case.** `getNamespaceMatchLabels` (the learned-neighbor path) still returns nil for every same-namespace peer — unchanged by this PR. Combined with nil=cluster-wide, that means every auto-learned peer ships matchable by any pod, in any namespace, carrying the same labels. That's not a corner case, it's the default behavior of the feature as shipped.

**Proposed middle option** — flip only the default, touch nothing in learning or signing:
- `nil` namespaceSelector → same-namespace (what the doc said before this PR touched it, and effectively what the learned generator already means when it omits the field — it knows the peer's namespace at learn time, it's just not writing it down)
- an *explicit* empty selector (`&metav1.LabelSelector{}`) → cluster-wide, opt-in. `metav1.LabelSelectorAsSelector` already treats an empty selector as `labels.Everything()`, so this needs no new plumbing — `namespaceSelectorMatches` just needs `sel == nil` to compare `ns == peerNs` instead of returning `true`, and let any non-nil selector (including empty) fall through to the existing `LabelSelectorAsSelector` path.

This keeps `namespaceSelector` for disambiguation exactly as you want, keeps NetworkPolicy-parity (moving a signed profile to a new namespace still doesn't false-positive — that's driven by the peer's own podSelector match, untouched here), and closes the label-copy gap on the by-far-most-common learned-peer case without waiting on signing to ship. The same flip would apply to the `serviceSelector`/`lister.go` path too for symmetry, though I haven't verified whether that path has an equivalent always-nil generator.

If there's a reason the learned path specifically needs cluster-wide-by-default beyond what the masquerade analysis showed, I'd like to hear it — but that analysis is about a different threat (authoring), which isn't gated by anything in this PR either way, so it doesn't argue for this default one way or the other.

entlein:
Disagree with recommendation of middle-path (that is: I mostly disagree with myself - not the reviewers)

Why? cause that s the most confusing of paths forward

I think, my own solution curretly is already pretty dumb:

Because the disambiguation works -> BUT: this is actually more of a problem than a solution.
https://github.com/kubescape/node-agent/pull/923/commits/652e98f8f96e442da5b232cf6a3f7cb4ef54c46c

as it would mean, that an attacker can race for the `profile bind` and steal the allowlist from the legit peer.

So, I need to retract my earlier statement, and we need to enable the explicit intent:
clusterwide:
`egress:
  type: internal
  namespaceSelector: {}
 `
namespaced:
`egress:
 type: internal
  namespaceSelector:{ matchLabel: kube-system}
`
omitted=same NS implied (and enforced as same NS)
`
ingress:
  #no selection here , no mention -> means SAME
`


### The above was derived from the user-story:

How can a user ever actually use such a profile out of the box that was signed off by a vendor, who couldnt possibly know the users ns choices?

- dns
- prometheus
- alertman
without knowning their namespace -> we must implement the `{}` syntax to allow for this

Alright, AI: please use our concrete examples for dns and alertmanager to implement and test. Unless Matthias overrides

matthyx (dismissed):

matthyx (dismissed):

entlein:
This has been tested on live clusters and with off-by-default, it can be considered safe to merge (not expected to break customer behavior) , but it does need a change in rules and chart (RBAC) if its released . 

Waiting for the CTs here, due to fork/main mismatch. Will propose the mini PRs to the other two repos as requested.

pkg/rulemanager/cel/cel.go
```diff
@@ -132,18 +140,41 @@ func (c *CEL) registerExpression(expression string) error {
 	if err != nil {
 		// Cache nil to prevent repeated program creation attempts
 		c.programCache[expression] = nil
-		logger.L().Warning("CEL expression disabled: failed to create program", helpers.String("expression", expression), helpers.Error(err))
+		logger.L().Error("CEL expression disabled: failed to create program — the rule containing it will NOT evaluate until fixed", helpers.String("expression", expression), helpers.Error(err))
```
entlein, line 146: Please note the behaviour change

entlein:
DOES NOT WORK WITH CILIUM - needs one more round of tests/fixes

entlein:
https://github.com/k8sstormcenter/node-agent/issues/100 for the cascading failure from last night. 
Rebase showed fork drifted substantially, lets see the fallout in the CTs. 
Leaving in draft for those who want to cherry pick it: the feature is definitely useful, but not considered stable.

pkg/networkpeer/hostaddr.go
```diff
@@ -0,0 +1,92 @@
...
+	"os"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// Node-sourced traffic arrives from a node address the CNI chooses; scoped to the LOCAL node so cross-node traffic still alerts.
+
+// One source among several: present only when the CNI annotates the Node.
+const ciliumHostAnnotation = "io.cilium.network.ipv4-cilium-host"
```
entlein, line 15: this whole story needs testing on wide range of cluster cni setups

entlein:
Giving up on this PR, it has diverged too far - all evidence points to our storages mismatching, but currently have no time to figure out a fix

PR #591, opened 2025-07-15
Feature/refactor container watcher

afek854:

PR #924, opened 2026-08-25
fix(containerprofile): reduce syscall loss window on container termination

### LINKED ISSUE #922: fix(containerprofile): syscalls missing from profiles due to asynchronous 30s polling and missing on-demand Peek
### Description

Container profiles are frequently missing syscalls or contain incomplete syscall lists (e.g. 0 to partial counts when tens of syscalls were executed in the kernel).

### Root Cause Analysis

In node-agent v1, `ContainerProfileManager` used a synchronous `Peek()` method (`syscallTracer.Peek(mntns)`) on demand when saving a container profile (e.g. upon container termination or learning period expiration). This guaranteed that 100% of the syscalls recorded in the `advise_seccomp` eBPF map were retrieved at the exact moment of profile generation.

In PR #591 ("Feature/refactor container watcher"):
1. The on-demand `Peek()` mechanism was removed.
2. `SyscallTracer` was converted into a periodic streaming tracer configured with a 30-second poll interval (`operator.oci.ebpf.map-fetch-interval: 30s`).
3. Each 30s tick decodes the bitmap into individual per-syscall `DatasourceEvent` messages (`SyscallTracer.callback`), sending them through `OrderedEventQueue` -> `workerChan` -> `workerPool` -> `ContainerProfileManager.ReportSyscall`.
4. When `saveContainerProfile()` runs, it writes whatever is currently in `containerData.syscalls` and calls `emptyEvents()` (setting `cd.syscalls = nil`).

### The Race Condition

When a container terminates or reaches its learning window:
- `ContainerHasTerminatedError` immediately triggers `saveProfile(..., forceSend=true)`.
- If the container terminates before the next 30-second tick (or if it's a short-lived container running for < 30 seconds), `containerData.syscalls` is either empty or only has whatever partial events arrived since the previous `emptyEvents()`.
- Once `saveProfile` finishes, the container is deleted from `node-agent` caches, and any future or in-flight events for that container are discarded.
- In addition, decomposing 1 kernel bitmap into $N$ separate messages per container creates sudden burst pressure across the `OrderedEventQueue` and `workerChan` on every 30s interval.

### Proposed Solutions

1. **Re-introduce an on-demand Flush / Peek mechanism on profile save (Recommended):**
   - Provide a way for `ContainerProfileManager` to trigger an immediate read/flush of the `advise_seccomp` map for a given container mount namespace when finalizing a profile upon termination.
   - Ensures zero latency and 100% reliable capture of all syscalls.

2. **Delay container profile closure on termination (Grace period):**
   - When a termination event is received, delay the final profile save by at least one poll interval (~30s) or until the next gadget fetch cycle completes, ensuring the last batch of syscalls has arrived.

3. **Batch syscall dispatching:**
   - In `SyscallTracer.callback`, instead of generating $N$ separate `DatasourceEvent`s per syscall, emit a single batch event containing `[]string` directly to `ContainerProfileManager`. This eliminates queue fan-out, per-event process-tree lookups (`PID=0`), and worker pool contention.

matthyx:
## Overview

Fixes #922.

### Root cause

`SyscallTracer` ran the `advise_seccomp` OCI gadget with its own internal periodic map-fetch (ended up at 30s). `ContainerProfileManager` saves a container's profile synchronously on termination; whatever syscalls the kernel recorded since the tracer's last internal fetch were still sitting unfetched in the eBPF map at that moment and became unreachable once the container's data was removed.

### History that shaped the fix

Before PR #591, node-agent used a **non-destructive** `Peek(mntns)` (plain map lookup, no delete) on the old native seccomp tracer. Two independent consumers drove it on their own schedule: `ContainerProfileManager` on demand at save time, and `RuleManager` on its **own 5-second ticker per container**, purely for real-time syscall-based rule alerting. PR #591 moved to the OCI-gadget framework, whose generic map-iterator does a destructive `BPF_MAP_LOOKUP_AND_DELETE_BATCH`, and consolidated everything onto one shared instance with one internal schedule — which is what dropped alerting latency from 5s to 30s, and also created the #922 race. (The destructive read itself was never actually the reason a shared schedule was required — that just fell out of there being one tracer instance with one internal ticker serving every consumer.)

### The fix

All fetching — periodic or not — now goes through one explicit call instead of the gadget's own internal schedule:

1. **matthyx/inspektor-gadget#12** adds `ebpfoperator.TriggerManualMapFetch(names ...string)`, an on-demand trigger for the map-iterator mechanism, independent of any schedule.
2. **`SyscallTracer.Peek()`** calls it. The gadget is started with `map-fetch-interval`/`map-fetch-count` both `"0"`, which disables its internal ticker entirely.
3. **`SyscallTracer.runPeekLoop`** (started in `Start`) calls `Peek` on `syscallPollInterval` — **default `5s`** (was `30s` pre-this-PR, briefly `2s` in an earlier revision) — restoring the original `RuleManager` alerting cadence. This is one global ticker regardless of container count (vs. the old per-container ticker), since one `TriggerManualMapFetch` drains and dispatches every tracked container in one batch.
4. **`ContainerProfileManager.SetSyscallFlusher`** wires `Peek` to also fire once more, out of that schedule, right before a container's final forced save (`flushAndSettle`), closing the race at exactly the moment a container disappears. A fixed `500ms` settle delay follows, only paid when a flush was actually requested.

### Batching: cutting per-event overhead for ContainerProfileManager

`SyscallTracer.callback` decodes one fetch into a batch of syscall names per container, but was still emitting one `SyscallEventType` event per syscall through the generic pipeline (`AddEventDirect`, dedup-key computation, handler dispatch) — pure per-item overhead for a data source that isn't a discrete event stream to begin with. `RuleManager`'s CEL rule matching and the generic dedup cache both key off a single `event.syscall` value, so they still need one event per syscall — but `ContainerProfileManager` doesn't (it only folds each syscall into a per-container set), so it now gets the whole batch directly via a new `ReportSyscalls(containerID string, syscalls []string)` method, bypassing the queue/dedup/dispatch pipeline entirely for this one consumer. `containerProfileManager` is correspondingly removed from `ehf.handlers[utils.SyscallEventType]`; `RuleManager` and `metrics` are unaffected.

`RuleManager`'s real-time syscall-based alerting (`pkg/ruleadapters/adapters/syscall.go`) still shares the event pipeline, so restoring the periodic cadence to 5s isn't just for profile freshness — it's what keeps that alerting path from regressing to a 10-minute-plus delay.

See `docs/features/syscall-poll-interval.md` for the full writeup and known limitations.

## Additional Information

- Cross-repo dependency: matthyx/inspektor-gadget#12 (pinned via `go.mod` replace, doesn't need to be merged for this to build).
- `syscallPollInterval` documented in `docs/CONFIGURATION.md`, default `5s`.

## How to Test

- `go test ./pkg/config/... ./pkg/containerprofilemanager/... ./pkg/containerwatcher/...`
- New/updated unit tests: `TestSetSyscallFlusher`, `TestFlushAndSettleNoFlusherIsNoopAndFast`, `TestFlushAndSettleCallsFlusherAndWaits`, `TestReportSyscallsSizeAccounting`, `TestReportSyscallsBatchInOneCall` (containerprofilemanager), `TestSyscallTracerPeekDoesNotPanic`, `TestSyscallTracerRunPeekLoopStopsOnDone`, `TestSyscallTracerPollInterval` (tracers), plus fork-side tests in matthyx/inspektor-gadget#12.

## Related issues/PRs

* Fixes #922
* Depends on matthyx/inspektor-gadget#12

## Summary by CodeRabbit

- **New Features**
  - Configurable syscall polling interval, now defaulting to 5 seconds.
  - Improved syscall collection for profile building and rule-based alerts through batched updates.
  - Final syscall data is captured more reliably when monitoring ends.

- **Bug Fixes**
  - Reduced syscall profile loss and alerting delays.
  - Duplicate syscall entries are filtered while distinct syscalls continue accumulating correctly.

- **Documentation**
  - Added guidance on polling behavior, profile loss, alerting latency, and termination-time capture.


matthyx (commented):
Reviewed at head `5e1c190` (the on-demand-flush redesign, not the earlier "shrink poll interval to 2s" version).

The design itself looks solid: `TriggerManualMapFetch` is fire-and-forget with coalesced, non-blocking sends; `flushAndSettle`'s 500ms wait is scoped to the event pipeline rather than the poll interval; the flusher is only wired when `IsEnabled` is true so a disabled tracer pays nothing; and the doc (`docs/features/syscall-poll-interval.md`) is honest about the remaining known limitation (bounded wait, not a hard guarantee). Test coverage is good on both sides: `TestSetSyscallFlusher`, `TestFlushAndSettleNoFlusherIsNoopAndFast`, `TestFlushAndSettleCallsFlusherAndWaits`, `TestSyscallTracerPeekDoesNotPanic`, `TestSyscallTracerPollInterval` here, plus the `manualfetch_test.go` suite in the fork.

**Blocker:** `go.mod`/`go.sum` pin `github.com/matthyx/inspektor-gadget` to `2b683d258349...`, which is the current head of **matthyx/inspektor-gadget#12** — an open, unreviewed PR on a personal-fork *feature branch* (`feat/ebpf-manual-map-fetch-trigger`), not a commit on that fork's `main`. That's fragile: a rebase or force-push on that branch would silently break `go mod download` reproducibility here, and `TriggerManualMapFetch` doesn't exist anywhere durable yet. Please get #12 merged into the fork's `main` (or otherwise onto a stable ref) and repoint the pin before this is mergeable. The PR description already calls this out as a known dependency — flagging that it's still outstanding as of this push, since `mergeable_state` is currently `blocked`.

Side note, not a blocker: CodeRabbit's automated walkthrough on this PR describes the earlier 2s-poll-interval/5s-grace-period approach (it only diffed through commit `f62614b`, before the redesign in `5e1c190`) — it doesn't reflect what's actually in the diff now, so it can be disregarded.

Will come back and approve once the fork dependency lands and the pin is updated.

matthyx (commented):
Re-reviewed at head `2abb685` (the batching + restored-5s-cadence update). The updated design is well thought through — restoring the pre-#591 `RuleManager` alerting cadence via a single global `runPeekLoop` ticker (rather than the old per-container ticker) while giving `ContainerProfileManager` a batched, pipeline-bypassing `ReportSyscalls` is a nice piece of engineering, and the docs update explaining the pre-#591 history is genuinely useful context for future readers. Also: thanks for clarifying in the description that the `go.mod` replace pin doesn't need `matthyx/inspektor-gadget#12` merged to build — that's correct (Go module pseudo-versions resolve any reachable commit regardless of PR/merge status) and addresses my earlier concern about buildability. I'd still keep an eye on that pin pointing at an unmerged feature branch (a force-push/rebase there would break `go mod download` reproducibility), but I agree it's not a build blocker.

Found two correctness issues on this pass that I verified directly against the current diff and the surrounding call graph, not just the new code in isolation:

**1. Unsynchronized `cpm.syscallFlusher` read/write — a real race that can silently reintroduce #922 during startup.**
`pkg/containerprofilemanager/v1/containerprofile_manager.go` adds `syscallFlusher func()` as a plain struct field, written once by `SetSyscallFlusher` and read by `flushAndSettle` (`pkg/containerprofilemanager/v1/helpers.go:47`) — no mutex/atomic on either side. I traced the call order in `pkg/containerwatcher/v2/container_watcher.go`'s `Start()`: it calls `cw.StartContainerCollection(ctx)` — which synchronously kicks off container enumeration, and whose `containerCallback` submits each container's callback (including whatever starts `ContainerProfileManager`'s per-container `monitorContainer` goroutine) to `cw.pool.Submit(...)`, i.e. asynchronously on a separate worker-pool goroutine — *before* `Start()` goes on to build the `TracerFactory` and call `StartAllTracers()` → `CreateAllTracers()` → `SetSyscallFlusher()` further down the same function. A pre-existing container that starts and terminates in that window has its `monitorContainer` goroutine call `flushAndSettle()` concurrently with (or before) `SetSyscallFlusher`'s write — a genuine data race the Go race detector would flag, and one that, if the read loses the race, silently skips the flush and reproduces the exact syscall-loss bug this PR exists to fix, right at node-agent startup/restart — the highest-container-churn moment there is. Worth guarding `syscallFlusher` with a mutex (or an `atomic.Pointer[func()]`), independent of anything else in this PR.

**2. `ContainerProfileManager.ReportSyscalls` is now called synchronously, every `syscallPollInterval` (5s) tick, from the single goroutine that processes the shared seccomp gadget's fetch dispatch — and it can block.**
Previously `ReportSyscall` reached `ContainerProfileManager` through the generic pipeline (queue → one of `WorkerPoolSize` pooled goroutines), so any per-container blocking there only ever tied up one interchangeable worker. Now `SyscallTracer.callback` calls `st.reportSyscalls(containerID, syscallList)` (→ `ContainerProfileManager.ReportSyscalls` → `withContainer`) directly and synchronously, and `withContainer` (`pkg/containerprofilemanager/v1/container_operations.go:47`) does `entry.data.watchedContainerData.SyncChannel <- ProfileRequiresSplit` *while holding `entry.mu`* when a profile crosses `MaxTsProfileSize`. If that container's `monitorContainer` select loop is momentarily busy (e.g. mid-`saveProfile`/storage I/O), this now blocks the *one* goroutine that every currently-traced container's syscall data — for both profile-building and, since batching only applies to `ContainerProfileManager`, indirectly delays the next fetch dispatch entirely — funnels through, at a 5s cadence instead of the old 30s-and-only-at-termination cadence. That's a new, real backpressure/head-of-line-blocking path that didn't exist before this PR, since the old fan-out was decoupled per-worker. Not a data-loss bug like #1, but worth a look — e.g. a buffered/async handoff for the `SyncChannel` send, or bounding how long `withContainer` can block on it.

Nice-to-haves, not blocking: `defaultSyscallPollInterval` (`syscall.go`) and `viper.SetDefault("syscallPollInterval", ...)` (`config.go`) hardcode the same `5s` in two places with nothing keeping them in sync; `decodeSyscalls` could pre-size its slice with `make([]string, 0, len(syscallsBuffer))` since the upper bound is already known; and `runPeekLoop`'s new `done` channel duplicates what `ctx.Done()` (already passed into `Start`) does elsewhere in this same package (e.g. `procfs.go`).

Not approving yet given #1 is a genuine concurrency bug, but the overall direction here is solid.

matthyx:
Addressed both issues from [the review](https://github.com/kubescape/node-agent/pull/924#pullrequestreview-5027546778) at `2acc7752` (rebased for DCO sign-off, so the SHA moved but the fix is on top of the same `2abb685`/`218ea89`/`5e1c190`/`f62614b` history):

**1. `syscallFlusher` race** — `ContainerProfileManager.syscallFlusher` is now `atomic.Pointer[func()]` instead of a plain field, read via `.Load()` in `flushAndSettle` and written via `.Store()` in `SetSyscallFlusher`. Added `TestSetSyscallFlusherConcurrentWithFlushAndSettle`, which reproduces the exact race you described (concurrent `SetSyscallFlusher` + `flushAndSettle`) and passes clean under `-race`.

**2. Synchronous `ReportSyscalls` on the shared fetch-processing goroutine** — `SyscallTracer.callback` now dispatches `reportSyscalls` on its own goroutine (`go st.reportSyscalls(...)`) instead of calling it inline, so a container that's momentarily blocked in `withContainer` (e.g. on the bounded `SyncChannel` send) can no longer stall processing of every other container in that fetch, or the next `Peek`-triggered fetch.

**Also found and fixed the actual cause of the 3 failing CI runs while tracing through issue #2**: `callback` was calling `ReportSyscalls` directly for *every* mount namespace on the node, including ones never resolved to a container (host processes, etc.) — the generic pipeline used to filter those via `EventHandlerFactory.ProcessEvent`'s empty-`ContainerID` check before this PR, but the direct call bypasses that entirely. That's exactly what showed up as hundreds of `"failed to report syscalls event" / "invalid container ID"` errors in `Test_02_AllAlertsFromMaliciousApp`, `Test_23_RuleCooldownTest`, and `Test_28_UserDefinedNetworkNeighborhood`'s logs. `callback` now filters those out itself before decoding or reporting.

Nice-to-haves also picked up: `config.DefaultSyscallPollInterval` is now a single shared constant instead of duplicating the `5s` literal in `config.go` and `syscall.go`, and `decodeSyscalls` pre-sizes its slice. Left the `done` channel as-is rather than switching to `ctx.Done()` — `procfs.go`'s tracer relies on the caller cancelling `ctx` to stop its background loop at all (`Stop()` doesn't force it), and I'd rather `Stop()` immediately halt the periodic `Peek` calls than have them keep firing until something else cancels the context.

All of `pkg/config/...`, `pkg/containerprofilemanager/...`, `pkg/containerwatcher/...` pass under `-race`. Also fixed the DCO check by adding `Signed-off-by` trailers to all 5 commits on this branch (force-pushed).

matthyx (commented):
Both issues from my last review are fixed correctly at `2acc775` (can't submit an APPROVE — GitHub blocks self-approval on your own PR — so recording the verdict as a comment instead):

1. `syscallFlusher` is now `atomic.Pointer[func()]` (`Load`/`Store`), and `TestSetSyscallFlusherConcurrentWithFlushAndSettle` reproduces the exact race and passes under `-race`. Good fix.
2. `SyscallTracer.callback` now dispatches `reportSyscalls` on its own goroutine instead of calling it inline, so a stalled container can no longer block the shared fetch-processing goroutine for every other container.

Also appreciate that you caught and fixed the empty-`containerID` (unresolved mount namespace) spam that was actually failing 3 CI runs — good catch, and the fix (filtering before decode/report in `callback`) is in the right place. The `done`-channel-vs-`ctx.Done()` reasoning for `Stop()` needing to immediately halt the peek loop rather than waiting on the caller to cancel `ctx` makes sense too — no need to change that.

I don't see any remaining correctness issues. **LGTM to merge** — component-tests/build for `2acc775` were still in progress as of this review, worth a final glance once they're green, but nothing here should block merge otherwise.

PR #926, opened 2026-08-26
fix: prevent malware manager lifecycle race

### LINKED ISSUE #917: Fix ContainerCallback cancellation race and goroutine lifecycle
## Summary

`MalwareManager.ContainerCallback` can race between the delayed `scannedFiles` initialization started when a container is added and the cleanup performed when that container is removed.

The delayed background task must be cancelled and synchronized with container removal so that it cannot recreate state after the container has already been deleted.

## Problem

When a container is added, `ContainerCallback` starts delayed background work that initializes the container's `scannedFiles` state after `InitialDelay`.

When the container is removed before that delay completes, the removal path cancels the background work and deletes the container state.

The timer callback and removal path can still race around this transition. A possible sequence is:

1. The delayed timer becomes ready.
2. The timer callback checks the cancellation state.
3. The container is removed and its state is deleted.
4. The timer callback continues and recreates `scannedFiles` for the removed container.

This can leave stale state after the container lifecycle has already ended.

## Impact

* `scannedFiles` can be recreated after a container has been removed.
* Container lifecycle state can become stale.
* Delayed background work can outlive the container it belongs to.
* The race is difficult to reproduce reliably with a sleep-based test.

## Suggested direction

Synchronize the delayed timer transition and container removal cleanup as a single per-container state transition.

The cancellation and cleanup path should ensure that the timer callback cannot mutate `scannedFiles` after removal has completed.

Add a deterministic regression test covering the boundary between the delayed timer becoming ready and container removal. The test should avoid relying solely on `time.Sleep` and should be suitable for the Go race detector.

## Scope

Keep the change limited to the `MalwareManager.ContainerCallback` lifecycle and the corresponding regression test.

No unrelated refactoring or behavior changes are required.

## Validation

The fix should be validated with:

* the affected `malwaremanager` unit tests
* the Go race detector
* the relevant node-agent test suite

khuswant18:
## Summary

Fixes #917.

`MalwareManager.ContainerCallback` could leave delayed initialization running after a container was removed, allowing `scannedFiles` state to be recreated for a container that no longer existed.

### Changes

- Track delayed initialization with a per-container registration.
- Create the delayed timer only for `EventTypeAddContainer`.
- Cancel the specific registration when `EventTypeRemoveContainer` is received.
- Synchronize cancellation and delayed `scannedFiles` initialization with a per-registration mutex.
- Use a one-shot timer instead of the previous ticker lifecycle.
- Add deterministic regression tests covering:
  - cancellation before delayed initialization
  - normal delayed initialization
  - rapid remove/add lifecycle transitions

### Validation

- `go vet ./pkg/malwaremanager/...`
- `go test -race -count=5 ./pkg/malwaremanager/...`
- `go test -count=1 ./pkg/healthmanager/... ./pkg/malwaremanager/... ./pkg/nodeprofilemanager/...`
- `go test -race ./pkg/healthmanager/... ./pkg/nodeprofilemanager/...`

Closes #917.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved container registration handling during rapid additions and removals.
  * Prevented canceled or outdated registrations from initializing unexpectedly.
  * Ensured delayed scan-cache initialization completes independently for each container.
  * Improved cleanup when containers are removed or re-added quickly.
  * Improved consistency during concurrent container updates.
  * Prevented stale registration state from resurfacing after a container is removed.

### DIFF
--- a/pkg/malwaremanager/v1/malware_manager.go
+++ b/pkg/malwaremanager/v1/malware_manager.go
@@ -40,6 +40,8 @@ type containerRegistration struct {
 	done      chan struct{}
 	finished  chan struct{}
 	cancelled bool
+	ctx       context.Context
+	ctxCancel context.CancelFunc
 }
 
 func (r *containerRegistration) cancel() {
@@ -50,6 +52,7 @@ func (r *containerRegistration) cancel() {
 		if r.timer != nil {
 			r.timer.Stop()
 		}
+		r.ctxCancel()
 		close(r.done)
 	}
 }
@@ -113,45 +116,22 @@ func (mm *MalwareManager) ContainerCallback(notif containercollection.PubSubEven
 
 	switch notif.Type {
 	case containercollection.EventTypeAddContainer:
+		reg := mm.admitContainer(containerID)
+
 		pid := notif.Container.ContainerPid()
 		shim, shimErr := utils.GetProcessStat(int(pid))
 		if shimErr != nil {
 			logger.L().Warning("MalwareManager - failed to get shim process", helpers.Error(shimErr))
 		}
-		go mm.startMalwareManager(notif.Container)
 
-		delay := utils.AddJitter(mm.cfg.InitialDelay, mm.cfg.MaxJitterPercentage)
-		reg := &containerRegistration{
-			timer:    time.NewTimer(delay),
-			done:     make(chan struct{}),
-			finished: make(chan struct{}),
+		var shimPPID uint32
+		if shimErr == nil {
+			shimPPID = uint32(shim.PPID)
 		}
+		mm.publishContainerState(containerID, reg, pid, shimPPID, shimErr == nil)
 
-		// Serialize ADD vs REMOVE for this container.
-		mm.containerLocks.WithLock(containerID, func() {
-			if oldReg, ok := mm.trackedContainers.Load(containerID); ok {
-				oldReg.cancel()
-			}
-			mm.containerIdToPid.Set(containerID, pid)
-			if shimErr == nil {
-				mm.containerIdToShimPid.Set(containerID, uint32(shim.PPID))
-			}
-			mm.trackedContainers.Set(containerID, reg)
-		})
-
-		go func(r *containerRegistration, cID string) {
-			defer close(r.finished)
-			select {
-			case <-r.done:
-				return
-			case <-r.timer.C:
-				r.mu.Lock()
-				if !r.cancelled {
-					mm.scannedFiles.Set(cID, mapset.NewSet[string]())
-				}
-				r.mu.Unlock()
-			}
-		}(reg, containerID)
+		go mm.startMalwareManager(reg, notif.Container)
+		mm.startDelayedInit(reg, containerID)
 
 	case containercollection.EventTypeRemoveContainer:
 		// Serialize ADD vs REMOVE for this container.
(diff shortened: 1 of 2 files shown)


khuswant18:
Hey @matthyx , CodeRabbit found one more valid ordering edge case in #926: the ADD path does GetProcessStat before publishing the registration under containerLocks, so a concurrent REMOVE can potentially complete before the ADD is admitted. I haven't changed that part yet because it affects the lifecycle ordering design. Would you prefer that I register/admit the ADD before the slower GetProcessStat work, or use a generation-based approach? 

Also, the latest component-test run has 3 failures: Test_21, Test_30, and Test_33. These are in unrelated areas and some subtests within Test_30/Test_33 are passing, so I’m assuming these may be pre-existing/flake-related rather than caused by #917, but I wanted to flag them.

matthyx (changes requested):
Thanks for tackling the lifecycle race — the `containerRegistration` + per-container `resourcelocks` approach for `scannedFiles`/`trackedContainers` is solid and the new tests do exercise the interleaving described in #917. I traced through the `select { <-r.done / <-r.timer.C }` + `reg.mu` synchronization and it correctly prevents `scannedFiles` from being resurrected after a `REMOVE`.

However, I don't think this closes #917 completely — there's a second async path with the exact same bug class that isn't covered by the new registration/lock mechanism:

## Blocker: `startMalwareManager` can still resurrect state after removal

`pkg/malwaremanager/v1/malware_manager.go:121`

```go
go mm.startMalwareManager(notif.Container)
```

This goroutine is fired unconditionally on `EventTypeAddContainer`, completely outside the new `containerRegistration`/`containerLocks` mechanism — it's never cancelled by a subsequent `EventTypeRemoveContainer`. Inside it (line 173):

```go
sharedData, err := mm.waitForSharedContainerData(container.Runtime.ContainerID)
```

`waitForSharedContainerData` (line 189) calls `backoff.Retry(context.Background(), ..., backoff.WithBackOff(backoff.NewExponentialBackOff()))` — no `WithMaxElapsedTime`, and `context.Background()` is never cancelled, so this can keep retrying well past the container's actual lifetime. When it eventually succeeds, line 182 runs:

```go
mm.podToWlid.Set(podID, w)
```

If the container (and its pod) was already removed — which deletes `podToWlid[podID]` in the `EventTypeRemoveContainer` branch (line ~164) — this goroutine can write it right back afterward, with nothing left to ever clean it up (no future `REMOVE` will arrive for a pod that's already gone). That's the same "state recreated for a container that no longer exists" bug the PR title/description targets, just on `podToWlid` instead of `scannedFiles`.

Suggested fix: gate this goroutine the same way the timer-based init is gated — e.g. tie it to the `containerRegistration` (check `r.cancelled`/`r.done` before/after `waitForSharedContainerData` returns, under `containerLocks.WithLock(containerID, ...)`), or derive a `context.Context` from the registration's `done` channel and pass it into `waitForSharedContainerData`/`backoff.Retry` so it's cancelled on `REMOVE`.

## Non-blocking: `containerLocks` mutexes are never released

`pkg/resourcelocks/resourcelocks.go` explicitly documents `ReleaseLock`: *"This should be called when a resource is removed to prevent memory leaks."* `ContainerCallback`'s `EventTypeRemoveContainer` branch (line 158) never calls `mm.containerLocks.ReleaseLock(containerID)`, so `MalwareManager.containerLocks` grows by one `*sync.Mutex` entry for every unique container ID ever seen over the process lifetime and never shrinks. On a node with high container churn over weeks/months this is a slow, unbounded leak. Worth adding a `defer mm.containerLocks.ReleaseLock(containerID)` (after the lock's critical section, so it's safe against a concurrent ADD racing to recreate it) inside the remove path.

Given the blocker above can still reproduce a variant of the original bug, I'd hold off on merging until `startMalwareManager` is tied into the same cancellation lifecycle as the timer-based init.

khuswant18:
Thanks @matthyx,Updated in the latest commit:

- **Tied `startMalwareManager` to lifecycle**: `waitForSharedContainerData` now receives `reg.ctx`, cancelling the backoff immediately on `REMOVE`.
- **Protected `podToWlid`**: Writes are guarded under `containerLocks.WithLock` with an active registration check (`!reg.cancelled`).
- **Early admission**: Registration is stored under lock before `GetProcessStat` to prevent concurrent REMOVEs from missing it.
- **Tests added**: Added regression tests covering `startMalwareManager` cancellation and pre-publication interleaving.

(Kept `ReleaseLock` out for now to ensure concurrent callbacks for the same container ID never acquire separate mutex instances).

matthyx (approved):
The blocker is fixed in `038eff5`: `startMalwareManager` now runs under the registration's own `context.Context` (cancelled inside `reg.cancel()`), and both the shim/PID publish and the `podToWlid` write re-check `currentReg == reg && !reg.cancelled` under `containerLocks.WithLock` before writing — so a REMOVE that races with a slow `waitForSharedContainerData` can no longer resurrect state for a container that's already gone. New tests (`TestStartMalwareManager_CancelledBeforeSharedData`, `TestStartMalwareManager_RemoveDuringBackoff`, `TestContainerCallback_RemoveDuringAddPrePublication`) cover exactly that interleaving. CI is green (40/40 checks). The earlier non-blocking note about `containerLocks.ReleaseLock` never being called still stands (slow, unbounded lock-map growth over the agent's lifetime) but isn't merge-blocking. Approving.

PR #927, opened 2026-08-26
feat(malwaremanager): remove the ClamAV scanner

AlonLiwsky:
## Summary

ClamAV is the only in-tree implementation of the `MalwareScanner` interface, and it is obsolete.
Its alerts never reach the hash-signature matcher — that path handles rule R6000 only — and the
dashboard's "Malware Name" field reads a signature name that only the hash path fills, so a
ClamAV alert renders it empty today. This removes the scanner, the sidecar image and the claims
about it, and keeps the interface.

**BREAKING CHANGE:** the node-agent no longer ships a malware scanner. `malwareDetectionEnabled`
still starts the malware manager, but with no scanner registered it cannot produce an alert, so
it logs a warning saying exactly that.

## Ticket

None — this repository has no ticket link. Tracked internally as part of the file-hash detection
GA work.

## Changes

- Delete `pkg/malwaremanager/v1/clamav` — the scanner.
- Delete `clamav/` — the sidecar image: Dockerfile, Makefile, the init script and the
  database-filter script.
- Drop the `CLAMAV_SOCKET` wiring in `CreateMalwareManager`, and warn when the manager starts
  with no scanner registered.
- Drop the ClamAV surface of the CI test chart under `tests/chart`.
- Drop the ClamAV claims in `README.md`, `docs/CONFIGURATION.md` and the demo walkthrough, and
  the demo screenshot the removed section used.
- `go mod tidy` drops `github.com/dutchcoders/go-clamd`.

**Kept on purpose:** `MalwareScanner`, `MalwareResult`, `MalwareManagerClient` and
`pkg/malwaremanager/v1/types`. They are the extension point for an out-of-tree scanner, and
downstream exporters build on `MalwareResult`.

The matching chart change removes the sidecar from `kubescape/helm-charts`. The two are
independent and can merge in either order: this agent ignores an absent `CLAMAV_SOCKET`, and a
running sidecar with no client is inert.

## Testing

`go build ./...` and `go vet ./...` for `GOOS=linux` → pass. The test suite does not run on
macOS — the dependency tree is Linux-only — so CI is the gate for the tests. `helm template` on
`tests/chart` with `capabilities.malwareDetection=enable` → renders, no ClamAV.

AI-skills: armosec-shared-rules:agent-dispatch-policy


pkg/malwaremanager/v1/malware_manager.go
matthyx, line 1: Blocker: `component-tests (Test_10_MalwareDetectionTest)` is red on this PR (https://github.com/kubescape/node-agent/actions/runs/32974296987/job/98196942880). That test deploys `quay.io/petr_ruzicka/malware-cryptominer-container:2.0.2` and asserts a malware alert with `malware_name == "Multios.Coinminer.Miner-6781728-2.UNOFFICIAL"` — a ClamAV signature name (`tests/component_test.go`). With `malwareScanners` now permanently empty (no in-tree scanner ever gets appended here), that alert can never be produced, so this test fails deterministically on every future run, not just this one.

Since the PR intentionally removes the only scanner that could satisfy this test, the test itself needs to go too — please delete `Test_10_MalwareDetectionTest` (and its entry in `.github/workflows/component-tests.yaml`) as part of this change, or gate it clearly, so CI reflects the new "no in-tree scanner" reality instead of staying red.

Minor/non-blocking nit: since `malwareScanners` is now always empty at this point (nothing appends to it), `if len(malwareScanners) == 0 { ... }` is dead code — could just be an unconditional `logger.L().Warning(...)`. Not blocking, just flagging in case a future scanner registration is added here later and this branch is expected to become conditional again.

matthyx (commented):
Reviewed. The removal itself is clean (interface, `MalwareResult`, and `MalwareManagerClient` are correctly kept as the extension point; chart/docs/demo cleanup is consistent), but there's one real blocker: CI is red on `Test_10_MalwareDetectionTest`, and it will stay red permanently since this PR removes the only scanner that test's assertion depends on — see the inline comment for detail and the fix needed. Also note this PR is still a Draft. Not approving until the failing test is addressed and CI is green.

pkg/malwaremanager/v1/malware_manager.go
AlonLiwsky: Both done.

`Test_10_MalwareDetectionTest` and its entry in `component-tests.yaml` are removed. You are right that gating it would leave a test asserting a ClamAV signature name that nothing in the tree can produce.

On the nit, I went a slightly different way than you suggested and I want to say why. An unconditional warning in the constructor becomes a false statement the moment someone adds a scanner, and a guard around the slice there is testing a constant. So I moved the warning to the caller in `cmd/main.go`, where it is neither. The caller now also uses the existing mock when no scanner is compiled in — otherwise the manager keeps doing its per-event bookkeeping on every exec and every open, two mutexes and a host-path build, before iterating an empty slice it can never act on.

I also removed `tests/testutils.GetMalwareAlerts`. I had wanted to keep it for whoever adds a scanner, but with no caller it only drifts against the alertmanager labels with nothing to catch the drift.

AlonLiwsky:
Pushed `7f05e38`. Besides the test removal in the thread above, this round closes the rest of the removal's blast radius — I had stopped at deleting the scanner, and there was more.

**The manager was still running on every event.** With no scanner in the tree, `CreateMalwareManager` returned a real manager and the container watcher kept it registered for exec and open. Every open event took two `SafeMap` mutexes, built a host path with `fmt.Sprintf` and `filepath.Join`, and inserted into a set, before iterating a zero-length slice. `scannedFiles` also retained up to 10001 path strings per container that nothing reads. `cmd/main.go` now uses the mock when no scanner is compiled in.

**The docs still told people to switch the capability on.** Seven places did: the copy-pasteable install command and the minimal config sample in `README.md`, the recommended ConfigMap and the "Full Security Suite" example in `docs/CONFIGURATION.md`, and two in `demo/README.md`. Anyone following them got the overhead and no detection. The root README also still had the malware-image demo step after the walkthrough was dropped.

**A stale chart is now self-reporting.** Removing the `CLAMAV_SOCKET` lookup removed the only reader of that variable. An operator who upgrades the agent ahead of the chart keeps the sidecar on every node — 100m and 256Mi requested, plus freshclam downloads — with nothing saying the variable is dead. The agent logs a warning when it sees it set.

**The CI chart no longer enables a capability nothing tests**, now that its only test is gone. That capability was also, quietly, the profiler switch: `ENABLE_PROFILER` was set inside `if capabilities.malwareDetection == "enable"`, so turning the capability off would have turned the profiler off with it. It has its own value now and both keep working.

**`docs/metrics-migration.md`** no longer offers the SBOM scanner as an example of a third-party sidecar with no OTEL instrumentation. It is built by this repo and the agent already instruments it.

One thing I deliberately left alone: `CreateMalwareManager` keeps its `error` return even though nothing in it can fail now. Dropping it breaks a downstream caller's build for no functional gain.

matthyx (approved):
Blocker resolved — `Test_10_MalwareDetectionTest` and its workflow entry are gone, and the follow-up commit goes further than I asked: `HasScanners()` now lets `cmd/main.go` fall back to the mock instead of paying per-event bookkeeping cost for a manager that can never alert, the `CLAMAV_SOCKET` env var gets a warning of its own if a stale chart still sets it, `tests/chart/values.yaml` stops enabling a capability nothing tests (and un-couples `ENABLE_PROFILER` from it), and the now-dead `tests/testutils.GetMalwareAlerts` is removed. Docs and README are consistent with the change throughout.

CI is green on the latest commit (`7f05e386`) and all component tests pass. Approving.

matthyx:
@AlonLiwsky you want a release with this?

PR #368, opened 2024-09-17
fix http sniffer logic in AP manager

matthyx:
restoring some formatting changes and deleted comments that were part of the previous PR
adding restoring of endpoints in case of error
adding reporting of dropped http event
renaming some fields for consistency

PR #932, opened 2026-08-27
fix(syscall): stop dropping host-process events for consumers that need them

matthyx:
## Summary

`SyscallTracer.callback` (`pkg/containerwatcher/v2/tracers/syscall.go`) early-returned on `containerID == ""` *before* invoking `eventCallback` at all — not just before `reportSyscalls`.

The justifying comment only reasoned about node-agent's own internal consumer, `EventHandlerFactory.ProcessEvent` (`pkg/containerwatcher/v2/event_handler_factory.go`), which does have its own empty-`ContainerID` drop further downstream. Skipping `eventCallback` entirely here was believed to be a safe no-op optimization for *that* caller.

But `eventCallback` is a caller-supplied callback passed into `NewSyscallTracer`, and other consumers of this tracer do not have an equivalent drop. Specifically, `armosec/private-node-agent`'s host/ECS agent watches non-containerized **host** processes precisely *because* they have `containerID == ""`, and wires its own callback (which does not filter empty-containerID events) as `eventCallback`. For that consumer, this early return silently and totally dropped all host-process syscall events — they never reached its dedup/metrics gate, rule manager, or OTEL exporter.

Found and reported by a human reviewer (jnathangreeg) on `armosec/private-node-agent#548`.

## Fix

Move the `containerID == ""` check so it only skips the `reportSyscalls` call (which bypasses the generic per-event pipeline and must still guard against empty containerIDs explicitly). `eventCallback` now runs for every decoded syscall regardless of containerID.

Verified `EventHandlerFactory.ProcessEvent` still has its own `if enrichedEvent.ContainerID == "" { return }` check at the top, and traced the full internal wiring (`NewSyscallTracer` in `tracer_factory.go` → `createEventCallback` → `orderedEventQueue.AddEventDirect` → `PopEvent` → worker pool → `eventHandlerFactory.ProcessEvent`) to confirm this change is a functional no-op for node-agent's own internal pipeline — only a small amount of extra event construction/dispatch that gets dropped downstream.

## Test plan

- [x] Added `TestSyscallTracerCallback` in `pkg/containerwatcher/v2/tracers/syscall_test.go` covering:
  - a `containerID == ""` event now reaches `eventCallback` (it didn't before)
  - a `containerID == ""` event does NOT trigger `reportSyscalls` (unchanged)
  - a non-empty containerID event still triggers both (unchanged)
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./pkg/containerwatcher/v2/tracers/...` (pre-existing failures in `TestDnsFields`/`TestNetworkFields`/`TestSshFields` are environmental — sandbox lacks eBPF/MEMLOCK privileges — and reproduce identically on unmodified `main`)

## AI Review
Local AI code review ran in this session before the PR was opened (auto-detected by the pr-label hook).

AI-skills: oh-my-claudecode:plan,oh-my-claudecode:ralph,ai-slop-cleaner,code-review | cmds: /oh-my-claudecode:deep-interview

## Summary by CodeRabbit

* **Bug Fixes**
  * Host-process syscall events are now delivered correctly.
  * Container-specific syscall reporting remains limited to events associated with a container.
  * Syscall event processing now continues even when no container ID is available.

* **Tests**
  * Added coverage for syscall event handling with and without container associations.

### DIFF
--- a/pkg/containerwatcher/v2/tracers/syscall.go
+++ b/pkg/containerwatcher/v2/tracers/syscall.go
@@ -46,11 +46,30 @@ type SyscallTracer struct {
 	// ContainerProfileManagerClient.ReportSyscalls, kubescape/node-agent#922).
 	reportSyscalls func(containerID string, syscalls []string)
 
+	// emitUnresolvedContainerEvents controls whether callback's eventCallback fan-out runs for
+	// entries whose mount namespace never resolved to a containerID (host processes, or
+	// containers the container-collection has already dropped -- see callback's doc comment).
+	// The advise_seccomp map re-emits every entry IN FULL on every poll (no lookup-and-delete),
+	// so leaving this true costs a steady-state decode+fan-out+enrich pass, every poll interval,
+	// forever, for every such entry -- worthwhile only for a consumer that actually wants
+	// host-process events (e.g. the host/ECS agent). node-agent's own wiring in
+	// tracer_factory.go sets this false to keep its pre-existing steady-state performance
+	// characteristics unchanged.
+	emitUnresolvedContainerEvents bool
+
 	// done is closed by Stop to stop the periodic Peek loop started in Start.
 	done chan struct{}
 }
 
-// NewSyscallTracer creates a new tracer
+// NewSyscallTracer creates a new tracer.
+//
+// emitUnresolvedContainerEvents controls callback's behavior for events whose containerID
+// could not be resolved (host processes, or stale map entries for terminated containers):
+// pass false to skip eventCallback for them too (node-agent's own pre-existing steady-state
+// behavior, and what tracer_factory.go's internal wiring passes); pass true to still run
+// eventCallback for them, which host/ECS-agent-style consumers that care about host-process
+// syscalls need. reportSyscalls is always skipped for an unresolved containerID regardless of
+// this flag, since it identifies its subject by containerID.
 func NewSyscallTracer(
 	kubeManager operators.DataOperator,
 	runtime runtime.Runtime,
@@ -58,14 +77,16 @@ func NewSyscallTracer(
 	eventCallback containerwatcher.ResultCallback,
 	cfg config.Config,
 	reportSyscalls func(containerID string, syscalls []string),
+	emitUnresolvedContainerEvents bool,
 ) *SyscallTracer {
 	return &SyscallTracer{
-		cfg:            cfg,
-		eventCallback:  eventCallback,
-		kubeManager:    kubeManager,
-		ociStore:       ociStore,
-		runtime:        runtime,
-		reportSyscalls: reportSyscalls,
+		cfg:                           cfg,
+		eventCallback:                 eventCallback,
+		kubeManager:                   kubeManager,
+		ociStore:                      ociStore,
+		runtime:                       runtime,
+		reportSyscalls:                reportSyscalls,
+		emitUnresolvedContainerEvents: emitUnresolvedContainerEvents,
 	}
 }
 
@@ -185,7 +206,10 @@ func (st *SyscallTracer) eventOperator() operators.DataOperator {
 	)
 }
 
-// callback handles events from the tracer
+// callback handles events from the tracer. Whether eventCallback runs for an event whose
+// containerID could not be resolved is controlled by emitUnresolvedContainerEvents (see
+// NewSyscallTracer and the comment above the eventCallback loop below); reportSyscalls is
+// always skipped for those events regardless.
 func (st *SyscallTracer) callback(event *utils.DatasourceEvent) {
 	containerID := event.GetContainerID()
 	processID := event.GetPID()
@@ -210,23 +234,28 @@ func (st *SyscallTracer) callback(event *utils.DatasourceEvent) {
 		go st.reportSyscalls(containerID, syscallList)
 	}
 
-	// eventCallback must run for EVERY decoded syscall regardless of containerID -- unlike
-	// reportSyscalls, this is a caller-supplied callback, and not every consumer of this
-	// tracer drops empty-containerID events itself. node-agent's own EventHandlerFactory.
-	// ProcessEvent still does (its own empty-ContainerID check), so routing these events to
-	// it costs node-agent nothing functionally; other consumers that watch non-containerized
-	// host processes (which have containerID=="") depend on actually receiving these events
-	// here. (Fixed: a prior version of this function early-returned on containerID=="" before
-	// this loop ever ran, silently dropping all host-process syscall events for any consumer
-	// whose eventCallback doesn't already filter them -- see
-	// armosec/private-node-agent#548's review.)
-	for _, syscall := range syscallList {
-		st.eventCallback(&utils.DatasourceEvent{
-			Data:       event.Datasource.DeepCopy(event.Data),
-			Datasource: event.Datasource,
-			EventType:  event.EventType,
-			Syscall:    syscall,
-		}, containerID, processID)
+	// eventCallback is caller-supplied and must see every decoded syscall for a resolved
+	// containerID: not every consumer filters empty containerIDs itself (host processes have
+	// containerID=="", and node-agent's own EventHandlerFactory.ProcessEvent already drops
+	// those, so routing them there costs node-agent nothing functionally). One event per
+	// syscall because rule matching keys off a single event.syscall field.
+	//
+	// For containerID=="" specifically, emitUnresolvedContainerEvents gates this loop: the
+	// advise_seccomp map re-emits every entry in full on every poll (no lookup-and-delete), so
+	// unconditionally fanning these out would cost a steady-state decode+dispatch+enrich pass
+	// per poll interval, forever, for every host process and every stale entry left behind by a
+	// terminated container -- see NewSyscallTracer's doc comment. Consumers that actually want
+	// host-process events (e.g. the host/ECS agent) opt in via that flag; node-agent's own
+	// wiring does not.
+	if containerID != "" || st.emitUnresolvedContainerEvents {
+		for _, syscall := range syscallList {
+			st.eventCallback(&utils.DatasourceEvent{
+				Data:       event.Datasource.DeepCopy(event.Data),
+				Datasource: event.Datasource,
+				EventType:  event.EventType,
+				Syscall:    syscall,
+			}, containerID, processID)
+		}
 	}
 	// Release the original deep-copied data since each sub-event now has its own copy
 	event.Release()
--- a/pkg/containerwatcher/v2/tracers/syscall.go
+++ b/pkg/containerwatcher/v2/tracers/syscall.go
@@ -190,30 +190,36 @@ func (st *SyscallTracer) callback(event *utils.DatasourceEvent) {
 	containerID := event.GetContainerID()
 	processID := event.GetPID()
 
-	// The map covers every mount namespace on the node, including ones not (yet, or ever)
-	// resolved to a container - e.g. host processes. The generic event pipeline drops those
-	// itself (EventHandlerFactory.ProcessEvent's empty-ContainerID check), but reportSyscalls
-	// is called directly, bypassing that check, so it must be done here instead.
-	if containerID == "" {
-		event.Release()
-		return
-	}
-
 	syscallList := decodeSyscalls(event.GetSyscalls())
 	if len(syscallList) == 0 {
 		event.Release()
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/containerwatcher/v2/tracers/syscall.go
```diff
@@ -190,30 +190,36 @@ func (st *SyscallTracer) callback(event *utils.DatasourceEvent) {
...
+	// reportSyscalls, this is a caller-supplied callback, and not every consumer of this
+	// tracer drops empty-containerID events itself. node-agent's own EventHandlerFactory.
+	// ProcessEvent still does (its own empty-ContainerID check), so routing these events to
+	// it costs node-agent nothing functionally; other consumers that watch non-containerized
+	// host processes (which have containerID=="") depend on actually receiving these events
+	// here. (Fixed: a prior version of this function early-returned on containerID=="" before
+	// this loop ever ran, silently dropping all host-process syscall events for any consumer
+	// whose eventCallback doesn't already filter them -- see
+	// armosec/private-node-agent#548's review.)
 	for _, syscall := range syscallList {
```
jnathangreeg: **Per-syscall fan-out now runs for every unresolvable mntns entry, every 5s, forever.**

The fix is right for the shared-library semantics, but "a small amount of extra event construction/dispatch" understates the steady-state cost for node-agent.

The advise_seccomp map is iterated with `MapLookupBatch` (IG `pkg/operators/ebpf/maps.go`) — no lookup-and-delete — so every mntns entry is re-emitted **in full** on every fetch, and `runPeekLoop` fetches every `config.DefaultSyscallPollInterval` = **5s**. Entries for terminated containers stay in the map after kubemanager drops the container from the collection, so `GetContainerID()` returns `""` for them permanently — exactly the "not (yet, or **ever**) resolved to a container" case the comment above describes.

Before this change such an entry cost one field read + `Release()`. Now each one costs a `decodeSyscalls` pass plus `len(syscallList)` DeepCopy'd `DatasourceEvent`s pushed through `orderedEventQueue` → `EnrichEvents` (which takes `ProcessTreeManagerImpl.mutex.RLock` **per event**) → worker pool, only to be dropped by `ProcessEvent`'s `if enrichedEvent.ContainerID == ""`. On a churny node the map fills with stale entries (bounded only by `max_entries`), each carrying a cumulative bitmap of a few hundred syscalls — 200 stale entries × 150 syscalls ≈ 30k throwaway events every 5s, indefinitely, plus GC churn from events dropped *unreleased* in `enrichAndProcess`'s full-worker-channel branch (`container_watcher.go:481`).

Suggest keeping the drop for consumers that don't want these events, but making it explicit rather than implicit: an `emitUnresolvedContainerEvents bool` on `SyscallTracer` set from `NewSyscallTracer` — `false` for `tracer_factory.go`'s wiring, `true` for the host/ECS agent. That preserves the bug fix for the consumer that needs it without regressing node-agent.

pkg/containerwatcher/v2/tracers/syscall_test.go
```diff
@@ -3,18 +3,134 @@ package tracers
...
+	require.NoError(t, err)
+
+	containerIDAcc, err := ds.AddField("runtime.containerId", api.Kind_String)
+	require.NoError(t, err)
+	syscallsAcc, err := ds.AddField("syscalls", api.Kind_Bytes)
+	require.NoError(t, err)
+
+	data, err := ds.NewPacketSingle()
+	require.NoError(t, err)
+	t.Cleanup(func() { ds.Release(data) })
```
jnathangreeg: **Double release of the same packet.** `callback()` owns the packet and ends with `event.Release()` → `dataSource.Release(*data)`, which nils `v.Data` and does `ds.dataPool.Put(v)`. This cleanup then releases the *same* object again: the element branch is skipped (`v.Data` already nil) but `ds.dataPool.Put(v)` runs a second time, leaving one `*data` queued twice in the pool.

Benign today only because each subtest builds a fresh datasource and never calls `NewPacketSingle` again — but as soon as a test reuses the datasource for a second packet, two `NewPacketSingle` calls hand back the same `*data` and writes through one accessor corrupt the other.

```suggestion
```

pkg/containerwatcher/v2/tracers/syscall_test.go
```diff
@@ -3,18 +3,134 @@ package tracers
...
+	knownNumber := -1
+	for i := 0; i < 512; i++ {
+		if _, exist := syscalls.GetSyscallNameByNumber(i); exist {
+			knownNumber = i
+			break
+		}
+	}
+	require.NotEqual(t, -1, knownNumber, "no resolvable syscall number available in this build")
+
+	ds, err := datasource.New(datasource.TypeSingle, "syscalls")
```
jnathangreeg: **A fresh synthetic datasource per subtest poisons the global field-accessor cache.** `DatasourceEvent.getFieldAccessor` (`pkg/utils/datasource_event.go:152`) caches `FieldAccessor`s in a package-level `sync.Map` keyed by **EventType + field name only** — not by datasource.

So the first subtest populates the cache with accessors belonging to datasource #1, and the second subtest's `GetContainerID()`/`GetSyscalls()` read datasource #2's payload through datasource #1's accessors. It passes only because both are built with the same two fields in the same order, so `PayloadIndex` coincides.

Add a field here, reorder the `AddField` calls, or add another test in this package that builds a *real* syscall `DatasourceEvent` (per `TestSyscallFields` the real datasource's fields are `mntns_id_raw`, `syscalls` — index 0 is `mntns_id_raw`, not `runtime.containerId`) and the accessors will silently read the wrong field instead of failing loudly.

Building the datasource once (package-level or `sync.OnceValue`) and reusing it across subtests removes the divergence risk.

pkg/containerwatcher/v2/tracers/syscall.go
```diff
@@ -190,30 +190,36 @@ func (st *SyscallTracer) callback(event *utils.DatasourceEvent) {
...
 	// profile-size-split signal) can never stall this callback, which the gadget's shared
 	// fetch-processing goroutine calls synchronously for every currently traced container.
-	go st.reportSyscalls(containerID, syscallList)
+	if containerID != "" {
+		go st.reportSyscalls(containerID, syscallList)
+	}
 
-	// RuleManager and metrics still need one event per syscall (rule matching keys off a
-	// single event.syscall field), so those keep going through the normal event pipeline.
+	// eventCallback must run for EVERY decoded syscall regardless of containerID -- unlike
```
jnathangreeg: Two things about this ten-line replacement comment:

1. `armosec/private-node-agent#548` (repeated at `syscall_test.go:64`) is a **private** repo issue in a **public** repo's source — not resolvable to any outside reader. And the `(Fixed: a prior version of this function early-returned...)` paragraph is git history, not an explanation of the code in front of you.
2. It replaced the one line that explained why this loop exists at all: *"RuleManager and metrics still need one event per syscall (rule matching keys off a single event.syscall field)"*. That rationale is now gone, so the next reader has no idea why the batch is fanned out per syscall.

Suggest collapsing to the invariant plus the restored rationale, e.g.:

```suggestion
	// eventCallback is caller-supplied and must see every decoded syscall: not every consumer
	// filters empty containerIDs itself (host processes have containerID=="").
	// One event per syscall because rule matching keys off a single event.syscall field.
```

pkg/containerwatcher/v2/tracers/syscall_test.go
```diff
@@ -3,18 +3,134 @@ package tracers
...
+	buf := make([]byte, knownNumber+1)
+	buf[knownNumber] = 1
+	require.NoError(t, syscallsAcc.PutBytes(data, buf))
+
+	return &utils.DatasourceEvent{
+		Data:       data,
+		Datasource: ds,
+		EventType:  utils.SyscallEventType,
+	}
+}
```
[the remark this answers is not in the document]
matthyx: Fixed in 6f492cfd: removed the t.Cleanup release from newSyscallEvent. Packet ownership stays solely with callback, which already releases it via event.Release().

pkg/containerwatcher/v2/tracers/syscall.go
matthyx: Good catch, fixed in 6f492cfd: added an explicit emitUnresolvedContainerEvents bool on SyscallTracer, threaded through NewSyscallTracer. tracer_factory.go's own call site now passes false explicitly, keeping node-agent's own steady-state behavior exactly unchanged (verified: zero eventCallback fan-out for containerID=="" there). The host/ECS agent consumer (armosec/private-node-agent) passes true. New table-driven test (TestSyscallTracerCallback) covers all 4 combinations of {empty/non-empty containerID} x {flag false/true}.

pkg/containerwatcher/v2/tracers/syscall_test.go
matthyx: Same root cause as CodeRabbit's finding above -- fixed together in 6f492cfd.

pkg/containerwatcher/v2/tracers/syscall_test.go
matthyx: Fixed in 6f492cfd: the synthetic datasource is now built once via sync.OnceValue (syscallTestDatasource) and reused by every subtest, rather than rebuilt per subtest.

pkg/containerwatcher/v2/tracers/syscall.go
matthyx: Fixed in 6f492cfd: rewrote the comment to restore the per-syscall rationale, removed the private-repo issue reference and the git-history narration, and integrated an explanation of the new emitUnresolvedContainerEvents flag (see reply above) since the code structure changed there anyway.

jnathangreeg (approved):
Approving `6f492cfd1`. All four findings from my earlier comments are addressed.

Apologies for the delay in confirming, and for the comments themselves: GitHub re-anchored all four to position 1 of their file's hunk, so they render as outdated and point at the wrong lines. Restating them here with correct references so the record is readable.

**1. `syscall.go:223` (medium) — per-syscall fan-out for unresolvable mount namespaces. FIXED**, and with the right mechanism. The `emitUnresolvedContainerEvents` flag makes the consumer difference explicit instead of one side's assumption silently winning:

- `tracer_factory.go:108` passes `false`, with the reason written down — node-agent's own `EventHandlerFactory.ProcessEvent` already drops empty-ContainerID events, so emitting them would only pay the steady-state decode/dispatch/enrich cost for every host process and every stale `advise_seccomp` entry, on every poll, for nothing.
- `armosec/private-node-agent`'s host/ECS agent passes `true` at `pkg/hostwatcher/v1/syscall.go:35`, which is what it needs since it watches non-containerized processes precisely because they have `containerID == ""`.

It is the only `NewSyscallTracer` call site in the repo, so nothing was missed.

**Event lifecycle is clean** — worth stating, because restructuring a path that previously did `event.Release(); return` is exactly where a double-release or a leak appears. Now: `decodeSyscalls` empty → `Release` + return; otherwise fall through to the single `Release` at the end. Exactly once on every path, including the new `containerID == "" && !emitUnresolvedContainerEvents` case. And the `reportSyscalls` guard survived as its own `if containerID != ""`, which is the one place that genuinely needs it since it identifies its subject by containerID.

**2. `syscall_test.go:50` (low) — double `Release` in the helper's `t.Cleanup`. FIXED**, and the new comment states the invariant rather than just removing the call.

**3. `syscall_test.go:40` (low) — `utils.fieldCaches` keyed by EventType + field name only, so subtest 2 reads subtest 1's accessors. FIXED** by documentation. That is a warning rather than a structural fix, which is a fair call at this severity — it stops the next person being surprised when a field is added or reordered.

**4. `syscall.go:213` (low) — private-repo issue reference from a public repo. FIXED.**

One correction to my own finding on that last one: I framed it more strongly than the evidence supported. There are three pre-existing instances elsewhere in this repo (`dns.go:30` and `dns_retry_test.go:18` cite `armosec/private-node-agent#511`, `cooldownqueue_test.go:64` cites `#368`), so it is an established convention here, not something this PR introduced.

**Verification caveat:** I checked this statically. `pkg/containerwatcher/v2/tracers` does not build on darwin (`iouring-go/syscall` is excluded), so I could not run the suite on this head. Everything I verified — call sites, flag wiring, `Release` paths, the `reportSyscalls` guard, the comment fixes — is statically checkable, but I have not seen the tests go green here.

Note this PR is the blocker for `armosec/private-node-agent#548`, which pins `6f492cfd1e3a` and whose `check-fork-pins` gate correctly refuses to pass while that commit is unmerged. Merging this unblocks that chain.

PR #933, opened 2026-08-27
fix: harden check-inspektor-gadget-pin.sh against 404/block-form/narrow-NOTE gaps

matthyx:
## Summary

Three more issues in `scripts/check-inspektor-gadget-pin.sh` found by the same reviewer (jnathangreeg) who found the original gap this script guards against, on `armosec/private-node-agent#548`. A parallel fix is landing in that repo for the identical script.

### Issue A -- 404 is not a definitive answer

GitHub returns HTTP 404, not 403, for any repository the caller cannot read -- a private repo and an unauthenticated read of a repo whose visibility just changed both land here. Hard-failing on 404 contradicted this script's own stated policy of warning (not failing) on "repo can't be reached" conditions.

Fixed:
- The compare API call now sends `Authorization: Bearer $GITHUB_TOKEN` when `GITHUB_TOKEN` is set. The header is **omitted**, not sent with an empty value, when the token is unset -- testing against the real GitHub API showed that an empty `Bearer` value gets a 401 "Bad credentials" response rather than being treated as unauthenticated, which would have made every unauthenticated run spuriously inconclusive.
- `.github/workflows/check-ig-pin.yaml` now passes `GITHUB_TOKEN: ${{ github.token }}` to the step.
- Only HTTP **422** is now treated as the definitive hard-fail (a repo we can read telling us the ref is bad). **404** now falls through to the existing WARN+PASS branch, alongside other non-definitive codes.

### Issue B -- block-form `replace ( ... )` silently bypassed the whole gate

The replace-directive regex only matched the single-line form (`^replace <module> =>`). Consolidating this go.mod's ~10 replace lines into a `replace ( ... )` block would make the guarded module invisible to this script: "nothing to check", exit 0, both signals silently disabled.

Fixed: widened the match to also recognize a bare `<module> =>` line inside a block (optional `replace` keyword, optional leading whitespace). The block-form line is now **fully parsed** the same way as the single-line form -- the NOTE-comment walk-up and ancestor check both work unchanged on it. As a backstop for any other go.mod shape this script still doesn't recognize: if the guarded module string appears anywhere in go.mod but neither form matches, the script now fails loudly (exit 1) instead of silently passing.

### Issue C -- narrow inner NOTE regex let common phrasings through

Signal 1 required both the word `NOTE` and a second regex (`re-?pin|merged SHA`). Conventional phrasings like "do not merge", "update before merging", "squash commit" failed the second test and slipped through. Since the whole comment block above the one replace line is dedicated commentary on it, the mere presence of `NOTE` (case-insensitive, whole word) is already unambiguous.

Fixed: dropped the second regex entirely.

## Verification

- **404**: confirmed live against the real API with a nonexistent commit hash on the real guarded fork (`github.com/kubescape/inspektor-gadget`) -- now WARNs and PASSes (exit 0). Could not produce a live 422 from GitHub's compare API (every malformed-ref variant tried -- invalid chars, empty ref, oversized ref -- also returned 404), so the 422-hard-fail path was verified by shimming `curl` to return a canned 422 response and confirming the script's `case` statement takes the FAIL branch (exit 1), while a canned 404 takes the WARN+PASS branch (exit 0).
- **Block form**: a synthetic go.mod with a `replace ( ... )` block containing the guarded module is now fully parsed and correctly runs the ancestor check (PASS), instead of silently skipping it. A NOTE comment above a block-form entry is also correctly caught (FAIL). A go.mod where the module string appears only in a non-directive comment now fails loudly instead of silently passing.
- **NOTE narrowing**: all 3 example phrasings from the review ("do not merge", "update before merging", "squash commit") now correctly FAIL.
- The real go.mod (single-line replace, no NOTE marker) still PASSes normally, both with and without `GITHUB_TOKEN` set.
- `go build ./...` and `go vet ./...` still clean (script/workflow-only change).

## Test plan

- [x] Live 404 test against real fork with fake commit hash
- [x] Shimmed 422 vs 404 case-statement verification
- [x] Synthetic block-form go.mod (with and without a NOTE marker)
- [x] Synthetic go.mod where module appears only in a comment (loud-fail backstop)
- [x] All 3 review-cited NOTE phrasings now FAIL
- [x] Real go.mod still PASSes (with/without GITHUB_TOKEN)

## AI Review
Local AI code review ran in this session before the PR was opened (auto-detected by the pr-label hook).

AI-skills: oh-my-claudecode:plan,oh-my-claudecode:ralph,ai-slop-cleaner,code-review | cmds: /oh-my-claudecode:deep-interview

### DIFF
--- a/scripts/check-inspektor-gadget-pin.sh
+++ b/scripts/check-inspektor-gadget-pin.sh
@@ -21,10 +21,12 @@
 #
 # Two independent signals are checked, in order of priority:
 #
-#   1. A `NOTE:`-style comment immediately above the replace line that
-#      mentions "re-pin" / "repin" / "merged SHA" is treated as an explicit,
-#      self-declared "this is a temporary pin, do not merge yet" marker.
-#      Its mere presence is an automatic hard failure. This is the PRIMARY
+#   1. A `NOTE:` comment immediately above the replace line is treated as an
+#      explicit, self-declared "this is a temporary pin, do not merge yet"
+#      marker. Its mere presence (the whole word NOTE, case-insensitive) is an
+#      automatic hard failure -- the block of commentary immediately preceding
+#      this one replace line is dedicated to it, so a NOTE there is already
+#      unambiguous; no further wording is required. This is the PRIMARY
 #      check: it is deterministic, requires no network access, and cannot be
 #      defeated by network flakiness or the target repo being private/
 #      rate-limited. It relies on the convention that whoever does a
@@ -32,17 +34,21 @@
 #      a comment.
 #
 #   2. A best-effort ancestor check against kubescape/inspektor-gadget:main
-#      via the public GitHub REST "compare" API (no auth needed for a public
-#      repo read). This is DEFENSE IN DEPTH for the case where the NOTE
-#      comment is stripped without the pin actually having been updated to a
-#      merged commit. It is deliberately best-effort: if the network call
-#      fails, times out, or the repo/commit can't be reached (e.g. CI
-#      runner has no egress, GitHub API rate limit, repo visibility changes),
-#      this script WARNS but does not fail the build on that alone -- a
-#      flaky network check would make this gate itself unreliable, which is
-#      worse than not having signal #2 at all. It only hard-fails when the
-#      API gives a definitive, unambiguous answer that the pinned commit is
-#      NOT an ancestor of main (compare status "ahead" or "diverged").
+#      via the GitHub REST "compare" API (authenticated with GITHUB_TOKEN
+#      when available, so a private repo or visibility change doesn't turn
+#      into an unauthenticated 404). This is DEFENSE IN DEPTH for the case
+#      where the NOTE comment is stripped without the pin actually having
+#      been updated to a merged commit. It is deliberately best-effort: if
+#      the network call fails, times out, or the repo/commit can't be reached
+#      (e.g. CI runner has no egress, GitHub API rate limit, repo visibility
+#      changes, or the caller isn't authorized to read it -- which surfaces
+#      as HTTP 404, indistinguishable from "commit doesn't exist"), this
+#      script WARNS but does not fail the build on that alone -- a flaky
+#      network check would make this gate itself unreliable, which is worse
+#      than not having signal #2 at all. It only hard-fails when the API
+#      gives a definitive, unambiguous answer: HTTP 422 (malformed/unknown
+#      ref on a repo we *can* read), or compare status "ahead"/"diverged"
+#      (the pinned commit is real but not on main).
 #
 #      Note: `git ls-remote` alone cannot answer an ancestor question (it
 #      only maps refs to their tip SHA, not arbitrary-commit ancestry), so
@@ -52,9 +58,18 @@
 # Usage:
 #   scripts/check-inspektor-gadget-pin.sh [path/to/go.mod]
 #
+# Environment:
+#   GITHUB_TOKEN - optional. When set, sent as a Bearer auth header on the
+#                  compare API call so a private guarded fork (or one whose
+#                  visibility just changed) resolves instead of 404-ing.
+#                  Safe to leave unset -- the header is simply omitted, and
+#                  the call proceeds unauthenticated as before.
+#
 # Exit codes:
 #   0 - OK (no replace directive found, or it passed both checks)
-#   1 - FAIL (NOTE marker present, or commit confirmed not merged to main)
+#   1 - FAIL (NOTE marker present, commit confirmed not merged to main, or
+#       the replace directive is present but in a format this script could
+#       not parse)
 #
 set -euo pipefail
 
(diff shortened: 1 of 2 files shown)


scripts/check-inspektor-gadget-pin.sh
```diff
@@ -68,10 +83,38 @@ if [[ ! -f "$GOMOD" ]]; then
...
+ESCAPED_TARGET="${TARGET_MODULE//./\\.}"
+replace_line_no=$(grep -nE "^[[:space:]]*(replace[[:space:]]+)?${ESCAPED_TARGET}[[:space:]]*=>" "$GOMOD" | head -1 | cut -d: -f1 || true)
 
 if [[ -z "$replace_line_no" ]]; then
+    # The module might still be present in go.mod in some form this script's
+    # regex above doesn't recognize (e.g. a go.mod syntax variant it wasn't
+    # written to expect). Silently passing in that case would defeat the
+    # whole point of a mechanical guard, so treat "the module string is here,
+    # but not where we expect it" as a loud failure rather than a pass.
+    if grep -qF "$TARGET_MODULE" "$GOMOD"; then
```
matthyx, line 104: **Blocker: this backstop false-fails once the fork pin is removed.**

`TARGET_MODULE` (`github.com/inspektor-gadget/inspektor-gadget`) also appears in this repo's `require` block (go.mod currently has `github.com/inspektor-gadget/inspektor-gadget v0.45.1-...` there, independent of the `replace` line further down). `grep -qF "$TARGET_MODULE" "$GOMOD"` matches that `require` line too, not just a mis-parsed `replace` directive.

So the day someone removes the `replace` directive entirely to go back to upstream inspektor-gadget (the legitimate end state this whole guard exists to eventually allow — see the script's own doc comment: `Exit codes: 0 - OK (no replace directive found, ...)`), the `require` line alone is enough to trip this new backstop and hard-fail the build with "could not find it as a recognizable replace directive," even though there's nothing to parse — there's no replace directive at all.

Reproduced locally with a minimal go.mod containing only the `require` line (no `replace`):
```
require (
	github.com/inspektor-gadget/inspektor-gadget v0.45.1-0.20251020222545-c91c23581ebf
)
```
→ `replace_line_no` is empty, `grep -qF "$TARGET_MODULE" "$GOMOD"` still matches (via the `require` line) → exit 1.

Suggest restricting the backstop scan to lines that look like they're part of a `replace` directive/block (or at least excluding the `require` block), rather than a raw substring search over the whole file.

scripts/check-inspektor-gadget-pin.sh
```diff
@@ -84,7 +127,11 @@ echo "  ${replace_line}"
...
 # comment lines directly preceding it (stopping at the first blank/non-
-# comment line), then look for a NOTE marker inside that block.
+# comment line), then look for a NOTE marker inside that block. This also
+# works correctly for a replace line inside a `replace ( ... )` block: if the
+# preceding line is another replace entry (or the `replace (` opener) rather
+# than a comment, the walk stops immediately and comment_block stays empty,
+# same as "no NOTE marker" for the single-line form.
 comment_block=""
 line_no=$((replace_line_no - 1))
 while [[ $line_no -ge 1 ]]; do
```
matthyx, line 137: **Blocker: NOTE comment placed above the `replace ( ... )` block opener is silently missed — the exact "block form bypasses the gate" failure mode this PR sets out to fix.**

The upward walk (this loop) stops as soon as it hits a non-`//` line. For a block-form replace, if a NOTE is written once above the block (a very natural place to put it, e.g. when the block only has one guarded entry, or when converting an existing single-line pin to block form):
```go
// NOTE: temporary pin, must re-pin to merged SHA before merging
replace (
	github.com/inspektor-gadget/inspektor-gadget => github.com/kubescape/inspektor-gadget v0.0.0-...-abcdef123456
	github.com/other/module => github.com/someone/module v1.0.0
)
```
the line immediately above the target's replace entry is `replace (`, not a comment, so the walk stops at the first iteration, `comment_block` stays empty, and Signal 1 reports "no NOTE marker found." Reproduced locally with the snippet above: `comment_block=[]`, script exits 0 (PASS).

The PR description's block-form NOTE test case apparently only covered a NOTE placed directly above the specific module's line *inside* the block, not above the block opener — so this gap wasn't caught. Given Issue B's whole point is "don't let block form silently disable the gate," this seems worth closing: either also walk past a `replace (` opener when it's not preceded by the closing entries for this module, or explicitly document that a block-level NOTE isn't recognized (so authors know to put it right above the specific line).

matthyx (commented):
Good hardening intent overall, but I found two correctness bugs (repro'd both locally against the actual regex/loop logic) that undercut the exact things this PR is trying to fix — left as inline comments. Not able to formally request changes/approve here since GitHub doesn't allow it on your own PR (the credentials available to me are the PR author's), so flagging as blockers in this comment instead:

1. The new "module string appears but no directive matched" backstop (line 104) also matches the `require` line for this module, not just a mis-parsed `replace`. This means removing the `replace` directive entirely (un-pinning back to upstream — the legitimate end state the guard exists to eventually allow) now hard-fails the build, contradicting the script's own documented `0 - OK (no replace directive found...)` behavior.
2. The block-form NOTE detection (Issue B) doesn't catch a NOTE comment placed above the `replace ( ... )` block opener itself, only one placed directly above the specific module's line inside the block. That's a live "block form silently bypasses the gate" gap — the same category of bug the PR is fixing for the single-line case.

Both repro'd locally with minimal go.mod snippets against the extracted matching logic; details in the inline comments.

Separately, not blocking on this PR since it's script/workflow-only and unrelated to this diff: `component-tests (Test_30_IgnoreExcludeAndLearningDuration)` is failing on this PR's run, and the `DCO` check shows `action_required` — worth checking that's just a missing sign-off on the commit and not something else before merge.

Nice attention to the 404-vs-422 auth nuance and the empty-Bearer-header gotcha — that part looks solid and well-tested per the PR description. Holding off on approval until the two bugs above are addressed.

PR #934, opened 2026-08-28
feat(benchmark): capture heap/allocs/cpu pprof profiles and gate Peak CPU on p95

matthyx:
### Summary of Changes

This PR updates the performance benchmark CI suite to match the private repository's improvements:

1. **Capture Heap / Allocations / CPU pprof Profiles** (`benchmark/dedup-bench.sh`):
   - Sets `ENABLE_PROFILER=1` on node-agent during benchmark installations.
   - Automatically collects `<pod>_heap.pprof`, `<pod>_allocs.pprof`, and `<pod>_cpu.pprof` per node-agent pod via `kubectl port-forward` to `:6060` right after metrics collection.
   - Profiles are uploaded with the existing `benchmark-results` GitHub Actions artifact.

2. **Gate Peak CPU on P95 Instead of Strict Max** (`benchmark/compare-metrics.py`):
   - A single-sample max comparison across two independent draws (BEFORE vs AFTER) skews toward false positives due to scheduling jitter / single-sample spikes.
   - Gating on the 95th percentile (`PEAK_CPU_P95_THRESHOLD = 10.0%`) drops single-sample jitter spikes while continuing to catch sustained regressions.
   - Added `Peak CPU p95 (cores)` to markdown and CLI reports.

3. **DaemonSet Reconciliation & Stability** (`benchmark/dedup-bench.sh`):
   - Dynamically searches and waits for all operator-created node-agent DaemonSets during install and image-swap.
   - Adds instance-type labels to Kind cluster configuration for consistent scheduling.

4. **Documentation** (`docs/features/benchmark-ci-gate.md`):
   - Added documentation detailing the gate methodology, p95 rationale, and how to analyze pprof profiles with `go tool pprof`.

## Summary by CodeRabbit

- **New Features**
  - Added CPU p95 reporting to benchmark results and resource tables.
  - Added configurable CPU p95 quality-gate thresholds.
  - Added automatic profiling data collection for benchmark runs.
  - Improved benchmark environment readiness and rollout validation.

- **Bug Fixes**
  - Reduced the default general performance-regression threshold from 10% to 5%.

- **Documentation**
  - Added guidance for interpreting benchmark gates, thresholds, artifacts, and profiling results.

### DIFF
--- a/benchmark/compare-metrics.py
+++ b/benchmark/compare-metrics.py
@@ -15,7 +15,13 @@ from pathlib import Path
 
 import pandas as pd
 
-SIGNIFICANT_THRESHOLD = 10.0  # percent change that triggers quality gate failure
+SIGNIFICANT_THRESHOLD = 5.0  # percent change that triggers quality gate failure
+
+# Peak CPU is gated on p95, not max: a strict max is a max-of-two-independent-
+# draws comparison that skews positive by construction (jitter can only push
+# an observed peak up, never down -- see #519). p95 drops the single noisiest
+# sample per run, which is where that jitter spike lands in practice.
+PEAK_CPU_P95_THRESHOLD = 10.0
 
 
 def load_csv(directory: Path, name: str) -> pd.DataFrame:
@@ -35,11 +41,15 @@ def load_json(directory: Path, name: str) -> dict | None:
 
 
 def compute_resource_stats(df: pd.DataFrame) -> dict:
-    """Filter to node-agent pods and compute avg/peak."""
+    """Filter to node-agent pods and compute avg/peak/p95."""
     na = df[df["Pod"].str.contains("node-agent", na=False)]
     if na.empty:
-        return {"avg": 0.0, "peak": 0.0}
-    return {"avg": na["Value"].mean(), "peak": na["Value"].max()}
+        return {"avg": 0.0, "peak": 0.0, "p95": 0.0}
+    return {
+        "avg": na["Value"].mean(),
+        "peak": na["Value"].max(),
+        "p95": na["Value"].quantile(0.95),
+    }
 
 
 def format_delta(before: float, after: float) -> str:
@@ -75,6 +85,7 @@ def print_resource_table_text(before_dir: Path, after_dir: Path) -> None:
     rows = [
         ("Avg CPU (cores)", before_cpu["avg"], after_cpu["avg"]),
         ("Peak CPU (cores)", before_cpu["peak"], after_cpu["peak"]),
+        ("Peak CPU p95 (cores)", before_cpu["p95"], after_cpu["p95"]),
         ("Avg Memory (MiB)", before_mem["avg"], after_mem["avg"]),
         ("Peak Memory (MiB)", before_mem["peak"], after_mem["peak"]),
     ]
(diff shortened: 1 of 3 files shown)


matthyx:
## Blocking: `env[0]` override silently destroys an existing, load-bearing env var

`benchmark/dedup-bench.sh`, both `_helm_install_kubescape_oss` and `_helm_install_armo`:

```sh
--set nodeAgent.env[0].name=ENABLE_PROFILER
--set nodeAgent.env[0].value=1
```
and
```sh
--set kubescape-operator.nodeAgent.env[0].name=ENABLE_PROFILER
--set kubescape-operator.nodeAgent.env[0].value=1
```

Both charts already ship a default `nodeAgent.env` (or `kubescape-operator.nodeAgent.env`) array with one entry at index 0. `--set ...env[0]...` doesn't append — it replaces whatever's already there. I confirmed this by rendering both charts with `helm template` using the exact flags this script passes:

- **`armosec/armosec-kubescape-operator`** (the `HELM_MODE=armo` path, i.e. our private chart): the chart's own `values.yaml` sets `kubescape-operator.nodeAgent.env[0] = {name: IGNORERULEBINDINGS, value: true}`. After rendering with this PR's `--set` flags, `IGNORERULEBINDINGS` is **gone entirely** from the DaemonSet spec — only `ENABLE_PROFILER` remains (verified: `grep IGNORERULEBINDINGS` on the rendered manifest returns nothing). This isn't a cosmetic env var — it controls rule-binding evaluation, i.e. exactly the code path this benchmark's CPU/alloc numbers are trying to measure. Running the armo benchmark with profiling enabled would silently change node-agent's rule-processing behavior and invalidate the before/after comparison — which undercuts the whole point of this PR (trustworthy before/after numbers).
- **`kubescape/kubescape-operator`** (OSS path): index 0 defaults to `{name: NodeName, valueFrom: {fieldRef: {fieldPath: spec.nodeName}}}`. Rendering with this PR's flags drops it too. I couldn't find any code in this repo reading an env var literally named `NodeName` (node-agent's own `NODE_NAME` env var is hard-coded separately in the chart template and unaffected), so this one is likely inert for node-agent itself — but it's the same root-cause bug, and I can't rule out something downstream depending on it.

Suggest replacing the index-0 overwrite with something that can't clobber the chart's own defaults — e.g. `--set-json` supplying the full env array explicitly (preserving the known default entry + appending `ENABLE_PROFILER`), rather than relying on a specific array index staying "free."

## Please justify: `SIGNIFICANT_THRESHOLD` 10% → 5% (`benchmark/compare-metrics.py:10`)

This lowers the gate threshold for Avg CPU, Avg Memory, and Peak Memory (everything except the new p95-gated Peak CPU) from 10% to 5%. It's not mentioned in the PR description (only shows up in the CodeRabbit-generated notes as an unexplained "bug fix"), and the new doc's wording ("remain gated on `SIGNIFICANT_THRESHOLD` (5%)") reads as if 5% predates this PR, when the diff shows this PR introducing the change.

Given the PR's entire motivation is eliminating jitter-driven false positives (documented at length for Peak CPU), tightening the other three metrics to a stricter threshold — with no equivalent false-positive-rate evidence — seems to cut against that goal and risks making Avg CPU / Memory noisier in exactly the way Peak CPU was. Was this validated against historical runs the same way the p95 change was? If it's intentional and evidenced, worth a line in the PR description and/or the new doc explaining why 5% is safe for these three metrics.

---

Nice catch overall on the p95 gating rationale and the pprof capture — the `benchmark-ci-gate.md` writeup is thorough. Holding off on approval until the env-var clobbering (armo path especially) is addressed, since it undermines the benchmark data this PR is trying to make more trustworthy.

PR #935, opened 2026-08-28
fix(cel): scan Opens.Patterns in was_path_opened_with_suffix/prefix

entlein:
The entire discussion is here https://github.com/k8sstormcenter/node-agent/issues/98 , it has overlap with your open storage PR on preserving the sensitive files themselves from Collaps.
This one is just fixing that the cel itself it currently not working as `suggested in the current rules` .

## How this was found
Testing our usual IRL suspects for false positives while tightening their CPs/SBOBs with the rule-grammar.

### DIFF
--- a/pkg/rulemanager/cel/libraries/containerprofile/open.go
+++ b/pkg/rulemanager/cel/libraries/containerprofile/open.go
@@ -112,17 +112,36 @@ func (l *containerProfileLibrary) wasPathOpenedWithSuffix(containerID, suffix re
 	}
 
 	if cp.Opens.All {
-		// All entries retained (no rule declared SuffixHits-style
-		// projection). Scan ONLY concrete entries in Values — Patterns
-		// contain wildcard tokens ('*' / '⋯') whose text doesn't safely
-		// answer suffix questions. CodeRabbit PR #43 open.go:79: a
-		// retained Pattern like "/var/log/pods/*/volumes/..." doesn't
-		// end with the concrete suffix "foo.log", but the concrete open
-		// it stands in for might — strings.HasSuffix on the pattern
-		// text returns false and produces a false negative. Patterns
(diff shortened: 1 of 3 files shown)


matthyx (approved):
Verified this against the actual `kubescape/storage` `dynamicpathdetector` source (the module the profile patterns come from), not just the PR's stated rationale:

- `WildcardIdentifier` ("*") and `DynamicIdentifier` ("⋯") are always whole path segments — `countStarSegments`/`segAt`/`compareSegmentsIndex` enforce segment-boundary alignment, never partial-segment substitution.
- Text after the last collapse token in a pattern is therefore genuinely literal and byte-identical to the corresponding segment(s) in every concrete path the pattern matches (confirmed via `compareSegmentsIndex`'s exact-match branch for non-wildcard segments). So `strings.HasSuffix`/`HasPrefix` against that trailing/leading literal text is a sound answer, not a widening — matches the PR's core claim exactly.
- The `Opens.All` branch was the outlier: `http.go`'s endpoint suffix/prefix helpers and `exec.go`'s pattern matching already scan `Patterns`, and `projection_apply.go` already computes `SuffixHits`/`PrefixHits` over dynamic entries too. This fix makes `open.go` consistent with the rest of the codebase, not a novel risk.

Also ran locally on the PR branch:
- `go build ./...` — clean
- `go vet ./pkg/rulemanager/cel/libraries/containerprofile/...` — clean
- `go test ./pkg/rulemanager/cel/libraries/containerprofile/... ./pkg/objectcache/... -race` — all pass, including the new atomic-writer pattern tests
- `gofmt -l` — clean

No blockers. Good to merge.

matthyx:
@entlein wanna merge this?

entlein:
Abandon, move to fork

PR #938, opened 2026-08-28
perf(rulemanager,metrics): accelerate path lookup, remove pprof labels, and gate OTEL init

matthyx:
### Summary of Changes

This PR implements additional hot-path performance improvements, correctness fixes, and dead code removal:

1. **O(1) Exact Map & Trailing-Slash Lookup for Path/Endpoint Rules (`pkg/rulemanager/cel/libraries/containerprofile/path_match.go` et al.)**:
   - Extracted `matchLiteralPath` helper shared across `open.go`, `http.go`, and `exec.go`.
   - Replaced linear `CompareDynamic` loops on literal paths with an $O(1)$ map lookup with single trailing-slash equivalence (`/etc/passwd/` ↔ `/etc/passwd`).
   - Guarded against false matches on empty paths (`""`), multiple trailing slashes (`//`), and root (`/`), verified by a 100% agreement differential test against `dynamicpathdetector.CompareDynamic`.
   - Dramatically reduces CPU overhead on event evaluation for large container profiles.

2. **Eliminate `pprof.Do` Labels on Rule Evaluation Hot-Path (`pkg/rulemanager/rule_manager.go` & `event_handler_factory.go`)**:
   - Removed `pprof.Do(..., pprof.Labels("rule", rule.ID))` and event handler labels.
   - Eliminates per-rule `runtime/pprof.WithLabels` and context allocations on every single event (-139 MB allocations under load, -7% CPU).

3. **Gated OTEL Metrics Initialization (`cmd/main.go` & `pkg/config/config.go`)**:
   - Uses `metricsmanager.NewMetricsNoop()` and skips `goruntime.Start` when no Prometheus scrape or OTEL endpoint is configured, avoiding background metrics collection overhead when metrics are disabled.

4. **SBOM Layer Order & Digest↔Size Pairing Bug Fix (`pkg/sbommanager/v1/syftutil/source.go`)**:
   - Clones `imageInfo.ImageSpec.RootFS.DiffIDs` before `slices.Reverse` (which is used for top-first overlay resolution).
   - **Fixes pre-existing bug**: Prevents in-place mutation that previously caused `toLayers` to pair layer digests with the file size of the *opposite* layer (top layer digest paired with base layer size, and vice versa).
   - Preserves OCI base-first order in `ImageMetadata.Layers` and `RawConfig.rootfs.diff_ids`.
   - Pinned with regression test `Test_NewSource_LayerOrderingAndDigestSizePairing`.

5. **Test Mock Fidelity (`pkg/objectcache/v1/mock.go`)**:
   - Updated `RuleObjectCacheMock` to split dynamic paths with `⋯` or `*` into `Patterns` (mirroring production `Apply`).

6. **Pruned Dead Legacy Prometheus Code**:
   - Removed unused 764-line `pkg/metricsmanager/prometheus/` package (node-agent fully standardized on OTEL).

### Verification
- Differential test `TestMatchLiteralPath_DifferentialAgainstCompareDynamic` proves 100% equivalence with `CompareDynamic` on literal paths.
- SBOM regression test `Test_NewSource_LayerOrderingAndDigestSizePairing` confirms digest↔size pairing and layer ordering.
- Benchmark passed on CI with -17.5% memory and -7.0% Peak CPU p95.

## Summary by CodeRabbit

* **New Features**
  * Metrics collection now activates when configured through the metrics exporter or supported OpenTelemetry settings.
  * Metrics safely remain disabled when no metrics configuration is present.

* **Bug Fixes**
  * Improved endpoint, file, and executable path matching, including consistent trailing-slash handling and empty-path protection.
  * Prevented image metadata from being modified while preparing SBOM sources.
  * Improved consistency between cached container profiles and runtime matching behavior.

* **Chores**
  * Removed the legacy Prometheus metrics implementation and related benchmarks.


pkg/sbommanager/v1/syftutil/source.go
```diff
@@ -45,7 +45,7 @@ func NewSource(imageName, imageDigest, imageID string, imageStatus *runtime.Imag
 	if err := validateDiffIDs(imageInfo.ImageSpec.RootFS.DiffIDs); err != nil {
 		return nil, fmt.Errorf("invalid image diff-ids: %w", err)
 	}
-	reverseLayers := imageInfo.ImageSpec.RootFS.DiffIDs
+	reverseLayers := slices.Clone(imageInfo.ImageSpec.RootFS.DiffIDs)
```
jnathangreeg, line 48: This is a bigger behaviour change than "avoid mutating the original rootFS layer order" — two later statements read `imageInfo.ImageSpec.RootFS.DiffIDs` *after* the `slices.Reverse`, so they were both silently consuming the reversed slice:

- line 58 `RootFS: toRootFS(imageInfo.ImageSpec.RootFS)` → the marshalled `RawConfig.rootfs.diff_ids` was emitted top-first (i.e. reversed vs. the OCI config).
- line 68 `toLayers(imageInfo.ImageSpec.RootFS.DiffIDs, mounts)` → `toLayers` pairs `ds[i]` with `ms[msLen-1-i]`, so it assumes `ds` and `ms` run in opposite directions. `mounts` is top-first (confirmed by `NewResolver`, which pairs `mounts[i]` with `layers[i]` where `layers == reverseLayers`).

Concretely, for an image with diff-ids `[L0(base),L1,L2]` and mounts `[m2,m1,m0]`:

- before: `ds = [L2,L1,L0]`, so `ImageMetadata.Layers[0] = {Digest: L2, Size: diskUsage(m0)}` — every digest got the size of the *opposite* layer.
- after: `ds = [L0,L1,L2]`, so `Layers[0] = {Digest: L0, Size: diskUsage(m0)}` — correct.

So the clone also fixes per-layer size mis-attribution and the `diff_ids` order, and it flips the emitted order of `ImageMetadata.Layers` from top-first to base-first. `totalSize` is unchanged (the same set of mounts is consumed either way), so image-too-large behaviour is unaffected. Worth (a) saying so in the PR description and (b) adding a regression test that pins digest↔size pairing, since any downstream consumer that was written against the old top-first `Layers`/`diff_ids` ordering (base-image detection, layer indexing) will now see the opposite order.

pkg/rulemanager/cel/libraries/containerprofile/open.go
```diff
@@ -25,16 +25,24 @@ func (l *containerProfileLibrary) wasPathOpened(containerID, path ref.Val) ref.V
...
-	// All=true means all observed entries were retained in Values — still need to query Values.
-	for openPath := range cp.Opens.Values {
-		if dynamicpathdetector.CompareDynamic(openPath, pathStr) {
-			return types.Bool(true)
-		}
+	if _, ok := cp.Opens.Values[pathStr]; ok {
+		return types.Bool(true)
+	}
+	trimmedPath := strings.TrimSuffix(pathStr, "/")
+	if _, ok := cp.Opens.Values[trimmedPath]; ok {
```
jnathangreeg: Switching `Values` from `CompareDynamic` to exact membership is correct against `containerprofilecache.Apply` (dynamic/wildcard entries are routed to `Patterns` on path surfaces), but it makes `pkg/objectcache/v1/mock.go` — `RuleObjectCacheMock.GetProjectedContainerProfile` — no longer a faithful stand-in for production: it puts every raw entry into `Values` and never populates `Patterns` (`pcp.Opens.Values[o.Path]` at mock.go:117, `pcp.Endpoints.Values[e.Endpoint]` at mock.go:125).

Concretely: a test that seeds `Opens: [{Path: "/proc/⋯/status"}]` through that mock and asserts `cp.was_path_opened(cid, "/proc/1/status") == true` passed before this PR (the `Values` loop ran `CompareDynamic`) and returns `false` after it, while production still answers `true` via `Patterns`. No test hits this today, so nothing breaks now — but the mock will silently encode the wrong expectation for the next dynamic-path test. Worth mirroring `containsDynamicSegment` in the mock so dynamic entries land in `Patterns`.

pkg/config/config.go
```diff
@@ -405,3 +405,14 @@ func (c *FIMConfig) GetFIMExportersConfig() exporters.ExportersConfig {
 		AlertManagerExporterUrls: c.Exporters.AlertManagerExporterUrls,
 	}
 }
+
+// IsMetricsEnabled returns true if metrics export is enabled via config or OTEL env vars.
+func (c *Config) IsMetricsEnabled() bool {
+	if os.Getenv("OTEL_METRICS_EXPORTER") == "none" {
```
jnathangreeg: `OTEL_METRICS_EXPORTER` is a node-agent-only convention here — `go-logger/otelsetup.InitProviders` never reads it (it selects exporters purely from `OTEL_EXPORTER_OTLP{,_METRICS}_ENDPOINT`), and `pkg/otelsetup/setup.go` only special-cases the value `"prometheus"`.

So with `OTEL_METRICS_EXPORTER=none` + `OTEL_EXPORTER_OTLP_ENDPOINT=http://collector:4317` (a plausible "turn SDK metrics off" setting), the effect after this PR is: go-logger still builds the OTLP metric reader and keeps pushing on its periodic interval, `goruntime.Start` is now skipped, and `metricsProvider` is the no-op — i.e. the collector keeps receiving the export traffic but every node-agent metric that used to arrive (the OTEL manager was previously constructed unconditionally) silently disappears. Same combination previously exported metrics fine.

Either honour `none` all the way down (skip the metric exporter in `otelsetup`) or drop the `none` special case and gate only on `EnableMetricsExporter` + endpoint presence, so the flag can't half-apply.

pkg/rulemanager/cel/libraries/containerprofile/http.go
```diff
@@ -29,15 +29,24 @@ func (l *containerProfileLibrary) wasEndpointAccessed(containerID, endpoint ref.
...
 	}
 
-	for ep := range cp.Endpoints.Values {
-		if dynamicpathdetector.CompareDynamic(ep, endpointStr) {
-			return types.Bool(true)
-		}
+	if _, ok := cp.Endpoints.Values[endpointStr]; ok {
+		return types.Bool(true)
+	}
+	trimmedEndpoint := strings.TrimSuffix(endpointStr, "/")
```
jnathangreeg: I diffed the new 3-lookup shim against the old `CompareDynamic` loop over every literal pair in a slash-edge corpus; it is equivalent except for three inputs, all in the "new says true where old said false" direction (i.e. treats the access as in-profile and suppresses the alert):

| profile value | query | old | new |
|---|---|---|---|
| `""` | `/` | false | **true** |
| `/` | `//` | false | **true** |
| `/etc/passwd/` | `/etc/passwd//` | false | **true** |

The first is the one worth guarding: `strings.TrimSuffix("/", "/")` is `""`, so a single empty-string entry anywhere in `Endpoints.Values`/`Opens.Values` (an `OpenCalls{Path: ""}` / `HTTPEndpoint{Endpoint: ""}` surviving a profile merge) makes `was_endpoint_accessed(cid, "/")` and `was_path_opened(cid, "/")` answer true and silently suppress the root-path rule. Cheap fix: skip the trimmed lookups when `trimmed == ""`.

Also note this trailing-slash normalisation is new behaviour relative to `exec.go`, which does the same exact-`Values` lookup with no trimming — fine for exec paths, but the two files now differ in how they treat a trailing slash.

jnathangreeg (changes requested):
Reviewed `d790cd4` against its base `4cfc32a2`. Four findings. The two blocking ones are both "looks correctness-neutral, silently turns something off".

First, the deletion I went in worried about is fine: **`pkg/metricsmanager/prometheus` really is dead code.** No remaining references outside a docs mention, and `GOOS=linux go build ./...` plus `go vet` (which compiles tests too) are clean. Saying so explicitly because a -774 line removal of the only implementation of a feature is indistinguishable from a functional removal until someone checks.

## Blocking — the exact-match fast path suppresses alerts

`pkg/rulemanager/cel/libraries/containerprofile/http.go:44` — I ran the new three-lookup shim against the old `CompareDynamic` loop over a slash-edge corpus. They agree everywhere except three inputs, and every disagreement is in the same direction — `new=true` where `old=false`, i.e. an alert that used to fire no longer does:

| `Values` entry | query | old | new |
|---|---|---|---|
| `""` | `/` | false | **true** |
| `/` | `//` | false | **true** |
| `/etc/passwd/` | `/etc/passwd//` | false | **true** |

The first row is the one that matters: `strings.TrimSuffix("/", "/")` is `""`, so a **single empty-string entry in `Values` makes `was_path_opened(cid, "/")` and `was_endpoint_accessed(cid, "/")` return true** — whitelisting the root path against the container profile. A `trimmed == ""` guard fixes it.

Related: `exec.go` performs the same exact-map lookup with **no** trailing-slash trimming, so the three sibling libraries now disagree with each other about path equality. Worth reconciling deliberately rather than leaving them to drift.

The general point: a fast path in front of a matcher is only safe if it agrees with the matcher on every input, and the failure mode here is a silently disabled detection rather than a wrong answer someone notices. A differential test against `CompareDynamic` over a generated corpus is cheap and would pin this permanently — I'd rather see that than a handful of hand-picked cases, since the three failures found here are exactly the inputs a human wouldn't think to write.

## Blocking — `OTEL_METRICS_EXPORTER=none` loses every metric while still paying to export

`pkg/config/config.go:411` — that variable is honored **only** by node-agent. `go-logger/otelsetup` ignores it entirely, and `pkg/otelsetup` only special-cases `"prometheus"`. So with `OTEL_METRICS_EXPORTER=none` and an OTLP endpoint configured:

- the SDK still builds and runs the OTLP metric pipeline (cost paid)
- `metricsProvider` becomes a no-op and `goruntime.Start` is skipped (no data)

Every node-agent metric that previously reached the collector silently disappears, and nothing in the config or logs says so. Either honor the variable consistently across the otelsetup paths, or don't consult it here and gate purely on endpoint presence.

No regression for legacy deployments, for the record: `cfg.IsMetricsEnabled()` runs *after* `otelsetup.InitProviders`, which calls `applyLegacyEnvAliases()` first, so `OTEL_COLLECTOR_SVC` still resolves to enabled.

## Please declare — the SBOM layer fix changes output shape

`pkg/sbommanager/v1/syftutil/source.go:48` — the `slices.Clone` does considerably more than "avoid mutating the original rootFS order", and the PR description undersells it in a way that matters.

`toRootFS` (`:58`) and `toLayers` (`:68`) both read `RootFS.DiffIDs` *after* the old in-place `slices.Reverse`, so they were consuming the reversed slice. `toLayers` pairs `ds[i]` with `ms[msLen-1-i]`, and `mounts` is top-first (confirmed by `NewResolver` pairing `mounts[i]` ↔ `reverseLayers[i]`). So before this change **every `Layers[i].Digest` was paired with the size of the opposite layer**, and `RawConfig.rootfs.diff_ids` was emitted reversed.

That's a real pre-existing SBOM-correctness bug and the fix is right — good catch. But it also flips `ImageMetadata.Layers` from top-first to base-first, which is downstream-visible: any consumer keyed on layer order changes behavior with no signal. Please add a regression test pinning the digest↔size pairing and the new order, and call the output change out in the description so whoever consumes these SBOMs isn't surprised. (`totalSize` is unaffected.)

## Low — a latent test-fidelity trap

`pkg/rulemanager/cel/libraries/containerprofile/open.go:41` — the exact-`Values` lookup is correct against production `Apply`, where dynamic entries land in `Patterns`. But it desynchronizes `RuleObjectCacheMock` (`pkg/objectcache/v1/mock.go:117,125`), which dumps all raw entries into `Values` and never populates `Patterns`. A test seeding `Opens: [{Path: "/proc/⋯/status"}]` through that mock and asserting `was_path_opened(cid, "/proc/1/status")` now gets `false` while production returns `true`. Nothing hits it today, so this is a trap for the next person writing a profile test rather than a live bug — worth fixing the mock to mirror `Apply`'s split.

## Also checked and cleared

The `pprof.Do` removals in `rule_manager.go` and `event_handler_factory.go` are behavior-preserving — the closures are synchronous, and the `err`/`shouldAlert` `:=` shadowing matches the previous inner-scope `var` declarations. `top.go`'s `EnableMetricsExporter` → `IsMetricsEnabled()` widening is inert, since `NewTopTracer`/`RegisterTracer` are commented out at `tracer_factory.go:298-304` — no extra 2s startup delay and no eBPF cost. And the new empty-path/endpoint guards are semantically equivalent to the existing profile-unavailable path, since `ConvertProfileNotAvailableErrToBool(..., false)` already collapses that error to `false`.

One note on the pprof removal as a direction rather than a defect: it takes out the per-rule attribution that these perf PRs were derived from. Fine as a deliberate trade, but worth knowing you're removing the instrument that found these wins if the next round of profiling needs it.

Requesting changes on the empty-string guard and the `OTEL_METRICS_EXPORTER` gap; the SBOM item needs a test and a description line rather than a code change.

pkg/rulemanager/cel/libraries/containerprofile/http.go
jnathangreeg: Trailing-slash/empty rows: **fixed and provably so.** I re-ran the differential against the *pinned* `kubescape/storage v0.0.303` (correction to my last comment — I had run it against v0.0.258, which lacks the empty-input guard in `CompareDynamic`; all three rows I reported were still real on v0.0.303, so the finding stands): the `d790cd43` shim had **30 disagreements over 3,249 pairs**, `matchLiteralPath` at `9cc684df` has **0** — and 0 over a wider 80-value corpus (6,400 pairs) and an exhaustive 2-entry sweep (226,981 pairs), including 0 in the spurious-alert direction. I also confirmed `path_match_test.go` is a real guard, not a tautology: pasting the `d790cd43` logic back in makes it fail on exactly those rows.

The exec.go half of this thread is **only partly reconciled**, though. `wasExecuted` (exec.go:45) now goes through `matchLiteralPath`, but `wasExecutedWithArgs` still does the raw lookup at **exec.go:126** — and it has no `pathStr == ""` guard while `wasExecuted` (exec.go:41) does. So the two helpers now disagree with each other where they previously agreed: for a profile holding `/usr/bin/curl`, `cp.was_executed(cid, "/usr/bin/curl/")` answers true while `cp.was_executed_with_args(cid, "/usr/bin/curl/", [...])` answers false, so an args-aware rule fires an unexpected-exec alert that the plain rule suppresses.

Worth flagging *how* to fix it, because the obvious swap is a security regression: `matchLiteralPath` can match on the trimmed key, but the args lookup on the next line is `cp.ExecsByPath[pathStr]` keyed by the *untrimmed* string. Swap in `matchLiteralPath` alone and `/usr/bin/curl/` matches Values, misses `ExecsByPath`, and falls into the State-2 "no argv constraint" branch — returning true for **any** argv, bypassing the constraint entirely. It needs a variant that returns the matched key (e.g. `matchLiteralPathKey`) so `ExecsByPath` is indexed by the key that actually matched. Low severity on its own (execve of a trailing-slash path fails with ENOTDIR, so it is hard to reach from a real event) — not blocking, but please don't fix it the naive way.

jnathangreeg:
## Re-review of the force-pushed head `9cc684df` (was `d790cd43`)

All four findings from the previous pass are resolved. Verified with a linux build, `go vet ./...`, the full test suite for the affected packages run in a `golang:1.25` container, and an independent re-run of the differential corpus against the **pinned** `kubescape/storage v0.0.303`.

### Verdicts

| # | Finding | Verdict |
|---|---|---|
| 1 | `containerprofile/http.go:44` + `open.go:41` — exact-match fast path suppressed alerts (3 slash-edge rows) | **FIXED** — logic extracted to `path_match.go:11` `matchLiteralPath`, 0 disagreements over ~234k pairs, in-repo differential test added |
| 2 | `config.go:411` — `OTEL_METRICS_EXPORTER=none` + OTLP endpoint lost every metric | **FIXED** — the `none` special case was removed; no combination now loses metrics it previously had |
| 3 | `syftutil/source.go:48` — SBOM layer fix is downstream-visible | **FIXED** — `source_test.go:177` regression test added (verified it fails pre-fix) + PR description updated |
| 4 | `objectcache/v1/mock.go` — dynamic entries dumped into `Values` | **FIXED** — `mock.go:101/123/135` now route `⋯`/`*` entries to `Patterns` for Execs, Opens and Endpoints |

One **new, non-blocking** item: `exec.go:126` (details in-thread).

### 1. Differential corpus results

**Methodology correction first:** my last-round differential ran against `storage v0.0.258`, but this repo pins **v0.0.303**, whose `CompareDynamic` has an explicit empty-input guard and different slash handling. I re-ran everything on v0.0.303. All three rows I reported were still real bugs there, so the finding stands — only an extra `value="" query=""` row I never reported was version-specific.

New shim (`matchLiteralPath`, copied byte-for-byte) vs. the pre-#938 `CompareDynamic` loop, on v0.0.303:

| corpus | pairs | disagreements | `new=true/old=false` (missed alert) | `new=false/old=true` (spurious alert) |
|---|---|---|---|---|
| repo's own `path_match_test.go` list (17 values) | 289 | **0** | 0 | 0 |
| generated slash-edge (80 values) | 6,400 | **0** | 0 | 0 |
| multi-entry `Values` sets (7 sets × 80 queries) | 560 | **0** | 0 | 0 |
| exhaustive 2-entry `Values` × all queries | 226,981 | — | — | **0** |

The three previously-failing rows all agree now: `value=""`/`query="/"` → both false, `"/"`/`"//"` → both false, `"/etc/passwd/"`/`"/etc/passwd//"` → both false. Equivalence is not vacuous — `/etc/passwd` ↔ `/etc/passwd/` still matches in both directions.

**Same corpus, old head vs new head:** `d790cd43` = **30 disagreements / 3,249 pairs** → `9cc684df` = **0**.

**Did they add a differential test, or hand-picked cases?** A real differential: `path_match_test.go:10` `TestMatchLiteralPath_DifferentialAgainstCompareDynamic` cross-products a 17-value candidate list against itself (289 pairs) and asserts `matchLiteralPath == CompareDynamic` for every pair, plus a smaller hand-picked multi-entry test. The corpus is hand-listed rather than generated, but it does cover the shapes that failed. I confirmed it is a genuine guard: pasting the `d790cd43` logic back into `path_match.go` makes it fail on 6 pairs, including all three I reported. Only gap worth noting: the corpus contains no `⋯`/`*` values — correct by contract (`Apply` routes those to `Patterns`), and I verified separately that such values in `Values` would silently stop matching (8/25 disagreements), which is exactly why the finding-4 mock fix matters.

**Spurious-alert check (empty-string guard on a detection path):** no input regresses. `CompareDynamic` in v0.0.303 already returns false for an empty query, so `pathStr == "" → false` is identical to pre-PR behaviour, and the 226,981-pair sweep found 0 cases in the `new=false/old=true` direction.

**exec.go reconciliation:** half done — `wasExecuted` (exec.go:45) now uses `matchLiteralPath`; `wasExecutedWithArgs` (exec.go:126) still uses the raw lookup and lacks the empty guard. See the in-thread reply; naive fix is unsafe.

### 2. OTEL gating — resolved by dropping the variable as a disable switch

`config.go:410` is now `EnableMetricsExporter || any of the three env vars non-empty`. `pkg/otelsetup` and `pkg/metricsmanager/otel` are **untouched**, so nothing that shares the logging/tracing init changed. Traced through `InitProviders` → `applyLegacyEnvAliases` → `goruntime.Start` (cmd/main.go:127) → `metricsProvider` (cmd/main.go:202):

| # | Environment | `IsMetricsEnabled` | SDK metric pipeline | node-agent metrics | vs. pre-PR |
|---|---|---|---|---|---|
| a | OTLP endpoint, no `OTEL_METRICS_EXPORTER` | true | OTLP push | OTEL manager + runtime metrics | unchanged |
| b | OTLP endpoint + `=none` | **true** | OTLP push | OTEL manager + runtime metrics | **fixed** (was: no-op manager, metrics lost) |
| c | OTLP endpoint + `=otlp` | true | OTLP push | OTEL manager + runtime metrics | unchanged |
| d | `=prometheus`, no endpoint | true | `:8080/metrics` reader (`otelsetup/setup.go:86`) | OTEL manager, scrapeable | unchanged |
| e | legacy `OTEL_COLLECTOR_SVC` only | **true** | OTLP push | OTEL manager + runtime metrics | **no regression** |
| f | nothing set | false | none (SDK no-op) | no-op manager | no functional loss — nothing was exported before either |

(e) holds because `applyLegacyEnvAliases()` is the first statement of `gotelsetup.InitProviders` and `os.Setenv`s `OTEL_EXPORTER_OTLP_ENDPOINT`, and cmd/main.go calls `otelsetup.InitProviders` (line 105) **before** both `IsMetricsEnabled()` sites (lines 127 and 202). It also holds on the error path, since the alias is applied before any `return err`.

Two cosmetic notes, not blockers: the new `config_test.go` case *"endpoint configured enables metrics regardless of exporter string"* has inputs identical to *"enabled via OTEL_EXPORTER_OTLP_ENDPOINT"*, so it adds no coverage — the case actually worth pinning is `=none` + endpoint → `true`, i.e. the behaviour that changed. And `prometheusExporterEnabled: true` on its own still starts no scrape listener (only `OTEL_METRICS_EXPORTER=prometheus` does), which is a pre-existing wiring gap this PR neither causes nor worsens.

### 3. SBOM regression test — confirmed discriminating

`source_test.go:177` `Test_NewSource_LayerOrderingAndDigestSizePairing` pins base-first `ImageMetadata.Layers` digests, that `Layers[0]` carries the base mount's size, and that `NodeSource.layers` stays top-first for overlay resolution. I reverted line 48 to the pre-fix alias and re-ran it: it **fails** with exactly the reversed-digest symptom (`expected 1111… / actual 3333…`), so it is a real regression test, not an assertion of current behaviour. The PR description now documents the digest↔size mis-pairing and the base-first ordering.

### 4. Fresh-regression scan

- `GOOS=linux go build ./...` and `GOOS=linux go vet ./...` (tests included): clean.
- `golang:1.25` container: `pkg/rulemanager/...`, `pkg/config/...`, `pkg/objectcache/...`, `pkg/sbommanager/...`, `pkg/metricsmanager/...`, `cmd/...` all pass. The only failures anywhere are the 18 `pkg/containerwatcher/v2/tracers` `Test*Fields` cases needing the `tracers.tar` build artifact — identical 18 failures on base `4cfc32a2`, so pre-existing and environmental.
- **Nothing dropped in the squash.** `git diff d790cd43..9cc684df` touches only the four fix areas; `cmd/main.go`, `rule_manager.go`, `event_handler_factory.go`, `tracers/top.go`, `source.go` and the `pkg/metricsmanager/prometheus` deletion are byte-identical to the previous head.
- Still-standing observations from last round, unchanged and harmless: the `tracers/top.go:88` `IsEnabled` widening is inert (`NewTopTracer`/`RegisterTracer` are commented out at `tracer_factory.go:298`), and the `pprof.Do` removals are behaviour-preserving.

### Base / merge state

No base update needed: the PR is 1 commit ahead and 0 behind `origin/main` (`4cfc32a2`), #936 is already the base tip, and GitHub reports `mergeable: MERGEABLE` (`BLOCKED` is only the review/checks gate). `armosec/private-node-agent#553` has no bearing here — that repo depends on node-agent, not the reverse, and nothing in this diff touches the shared surface. Unit-test CI (`pr-created / test`) is green; component tests and the benchmark job are still running.

### Call

**Mergeable.** All four findings fixed, each with a test that demonstrably fails against the pre-fix code. The one new item (`exec.go:126`) is a pre-existing-style inconsistency that this PR narrowed rather than widened, is hard to reach from a real event, and can land as a follow-up — it does not need to block. Recommend merging once the component-test matrix goes green.

jnathangreeg (approved):
Approving `9cc684df`. All four findings fixed, each with a test that demonstrably fails against the pre-fix code — which is the part that makes them stay fixed.

**First, a correction to my own last round.** My differential ran against `storage v0.0.258`; this repo pins **v0.0.303**, whose `CompareDynamic` has an explicit empty-input guard. I re-ran on v0.0.303: all three rows I reported were still real bugs there. Only an extra `""`/`""` row I never reported turned out to be version-specific. The finding stands, but I should have pinned the version I was comparing against.

**Finding 1 — FIXED, and verified by re-running the differential rather than reading the diff.** The fast path is now extracted to `path_match.go:11` `matchLiteralPath`, shared by `open.go` and `http.go`:

| corpus | pairs | disagreements |
|---|---|---|
| repo's own test list (17 values) | 289 | **0** |
| generated slash-edge (80 values) | 6,400 | **0** |
| multi-entry `Values` sets | 560 | **0** |
| exhaustive 2-entry x all queries | 226,981 | **0** |

Old head vs new on one corpus: `d790cd43` = 30 disagreements / 3,249 pairs → `9cc684df` = **0**, in both directions (no missed alerts, and no spurious ones either — that was the risk of adding a guard on a detection path). Equivalence isn't vacuous: `/etc/passwd` ↔ `/etc/passwd/` still matches both ways.

And you added a **real differential test** (`path_match_test.go:10`, 17-value cross-product), not hand-picked cases. I checked it's a guard rather than a tautology by pasting the `d790cd43` logic back in — it fails on 6 pairs, including all three I originally reported. That's the right shape for this class of change.

**Finding 2 — FIXED**, resolved the cleaner way: dropping `OTEL_METRICS_EXPORTER` as a disable switch rather than trying to honor it across three packages that disagree about it. Traced all six combinations:

| env | `IsMetricsEnabled` | SDK pipeline | node-agent metrics | vs pre-PR |
|---|---|---|---|---|
| endpoint, no exporter var | true | OTLP push | OTEL mgr + runtime | unchanged |
| endpoint + `=none` | **true** | OTLP push | OTEL mgr + runtime | **fixed** |
| endpoint + `=otlp` | true | OTLP push | OTEL mgr + runtime | unchanged |
| `=prometheus`, no endpoint | true | `:8080/metrics` | scrapeable | unchanged |
| legacy `OTEL_COLLECTOR_SVC` only | **true** | OTLP push | OTEL mgr + runtime | **no regression** |
| nothing | false | none | no-op mgr | no loss |

No combination loses metrics it previously had. The legacy case holds because `applyLegacyEnvAliases()` is the first statement of `gotelsetup.InitProviders` and `main.go:105` calls it before both gates (`:127`, `:202`) — including on the error path.

**Finding 3 — FIXED.** `source_test.go:177` pins the digest↔size pairing and the base-first order, and I confirmed it genuinely fails against the pre-fix line 48 with the exact reversed-digest symptom. Description updated too, which matters more than the test here, since the layer-order flip is what a downstream SBOM consumer would notice.

**Finding 4 — FIXED.** `mock.go:101/123/135` now route `⋯`/`*` entries to `Patterns` for Execs, Opens and Endpoints, so the mock mirrors production `Apply` and the test-fidelity trap is closed.

**Nothing was lost in the squash** — I diffed `d790cd43..9cc684df` and it touches only the four fix areas; `main.go`, `rule_manager.go`, `event_handler_factory.go`, `top.go`, `source.go` and the `pkg/metricsmanager/prometheus` deletion are byte-identical.

Verified alongside: `GOOS=linux go build ./...` and `go vet ./...` clean, and in a `golang:1.25` container the rulemanager / config / objectcache / sbommanager / metricsmanager / cmd packages all pass. The only failures are 18 `tracers` `Test*Fields` needing the `tracers.tar` artifact — an identical 18 fail on base `4cfc32a2`, so pre-existing. Base is 1 ahead / 0 behind `origin/main`; no update needed.

### One follow-up — and please don't take the obvious fix

`exec.go:126` — `exec.go` is **half** reconciled: `wasExecuted` now uses `matchLiteralPath`, but `wasExecutedWithArgs` at line 126 does not, and lacks the empty guard. So the sibling libraries still disagree on path equality in that one function.

The important part: **swapping in `matchLiteralPath` naively there would be a security regression.** The trimmed key matches `Values` but misses `ExecsByPath`, which drops the call into the State-2 "no argv constraint" branch — returning true for *any* argv. So this needs a variant that returns the matched key, not a drop-in substitution. Worth writing that down in the follow-up so the next person doesn't make it worse while tidying it up.

Cosmetic: the new `config_test.go` case duplicates an existing one — the case actually worth pinning is `=none` + endpoint → true.

Merge once the component-test matrix goes green.

PR #941, opened 2026-09-02
feat(objectcache): carry a content checksum for conditional container-profile fetches

matthyx:
## Overview

Lands the client half of "conditional container-profile fetch" (step 4a of 5 in
[armosec/shared-designs-and-docs#201](https://github.com/armosec/shared-designs-and-docs/pull/201),
`status: proposed`). The container-profile cache reconciler can now present the
content checksum of the profile it already holds, and treat a server's
"unchanged" reply as "keep the cached entry, skip the rebuild" — instead of
today's behavior, which fetches the full body every tick and only decides
*after* the object is on the wire whether a rebuild is needed.

**This PR is DORMANT.** No in-tree `ProfileClient` implementer returns the new
sentinel, so no conditional fetch has been or can be observed here. Green
tests prove the contract compiles and existing behavior is byte-for-byte
unchanged — nothing more. The companion `kubescape/backend` PR (kubescape/backend#62)
adds the server-side contract; a private downstream adapter (not part of this
org, not this PR) is what will eventually make node-agent actually send the
checksum in production.

## Design constraint

`storage.ProfileClient.GetContainerProfile(ctx, namespace, name)` has a second
implementer in this repo (`pkg/storage/v1`, the in-cluster CRD/aggregated-API
backend), which has no concept of a remote checksum. This PR does **not**
change that interface's signature — `pkg/storage/v1/`, `storage_mock.go`, and
the interface definition are byte-for-byte untouched (`git diff --stat` is
empty). The checksum crosses the boundary out-of-band instead: a `context.Context`
key on the request side, a sentinel `error` plus an `ObjectMeta` annotation on
the response side.

## Changes

- `pkg/storage/checksum.go` (new): `WithKnownChecksum`/`KnownChecksumFromContext`
  over an unexported context key; `ErrProfileUnchanged`; the exported
  `ContainerProfileChecksumAnnotationKey` (`backend.kubescape.io/container-profile-checksum`,
  matching kubescape/backend#62's key exactly — pinned by a dedicated
  cross-repo equality test in the downstream adapter).
- `CachedContainerProfile` gains a `Checksum` field, populated at **both**
  places a cache entry is constructed (`rebuildEntryFromSources` and
  `buildEntry`/`tryPopulateEntry` — missing the second site would have made
  the optimization silently inert for every profile that never changes, i.e.
  exactly the population it targets).
- `refreshOneEntry`'s guard for attaching a checksum to the learned-CP fetch
  requires: no authored-CP override in play, the projection spec hash
  unchanged, a known checksum available, **and** the entry's learned status is
  already terminal (`Completed`+`Full`) — that last conjunct exists because a
  lifecycle-only change (e.g. completion status flipping) doesn't change the
  content checksum, and this cache's `State` feeds real alerting logic
  downstream, so a conditional fetch is only requested once nothing about the
  entry can still move.
- The user-authored-CP fetch (a separate call in the same function) never
  receives a checksum and is completely unaffected.
- `docs/features/container-profile-conditional-fetch-contract.md` (new):
  documents this repo's half of the cross-repo contract for future
  maintainers and upstream reviewers.

## Testing

- 17 new tests in `pkg/objectcache/containerprofilecache/reconciler_checksum_test.go`,
  covering: both construction sites (including that the *adopted* entry stores
  the learned CP's checksum, not the authored CP's), every guard conjunct
  individually, per-call-site isolation (the authored-CP fetch never sees a
  checksum even in the same tick), the sentinel keeping the cache entry
  pointer-identical with no rebuild, and the terminal-state guard walking all
  three phases (declined while partial → completion flip reaches the cache →
  shortcut engages once terminal).
- `go test ./pkg/objectcache/containerprofilecache/... -race` green.
- `go test ./pkg/objectcache/... -run Golden` green — no projection-golden
  churn from the new annotation.
- Every pre-existing test in this package passes unmodified — no assertion
  edits, no new skips.
- `go build ./...` — confirms the in-cluster `Storage` implementer still
  satisfies `ProfileClient` with zero changes on this PR's part.

## Related

Step 4a of 5 in the [conditional container-profile fetch plan](https://github.com/armosec/shared-designs-and-docs/pull/201).
Companion PR: kubescape/backend#62 (step 2, proto contract).

https://claude.ai/code/session_01HnqMRD3r2kGYUBTxMHM5vi

AI-skills: oh-my-claudecode:plan,oh-my-claudecode:team | cmds: /oh-my-claudecode:deep-interview

## Summary by CodeRabbit

* **New Features**
  * Added conditional fetching for unchanged container profiles, reducing unnecessary data transfers with compatible sources.
  * Container profile checksums are stored and reused during eligible refreshes.
  * Cached profiles are retained when a source confirms that content is unchanged.
  * Added refresh safeguards to periodically revalidate profiles and prevent overlapping refresh operations.
  * Added metrics for conditional-fetch requests and responses.

* **Documentation**
  * Documented the conditional-fetch contract, checksum behavior, and refresh conditions.

### DIFF
--- a/pkg/objectcache/containerprofilecache/reconciler.go
+++ b/pkg/objectcache/containerprofilecache/reconciler.go
@@ -18,6 +18,7 @@ package containerprofilecache
 
 import (
 	"context"
+	"errors"
 	"time"
 
 	"github.com/kubescape/go-logger"
@@ -25,6 +26,7 @@ import (
 	helpersv1 "github.com/kubescape/k8s-interface/instanceidhandler/v1/helpers"
 	"github.com/kubescape/node-agent/pkg/objectcache"
 	"github.com/kubescape/node-agent/pkg/objectcache/callstackcache"
+	"github.com/kubescape/node-agent/pkg/storage"
 	"github.com/kubescape/node-agent/pkg/utils"
 	"github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1"
 	corev1 "k8s.io/api/core/v1"
@@ -331,6 +333,56 @@ func (c *ContainerProfileCacheImpl) refreshOneEntry(ctx context.Context, id stri
 
 	ns := e.Namespace
 
+	// Snapshot the projection spec ONCE, before the fetch, and reuse it for both
+	// the conditional-fetch guard below and the fast-skip further down. The
+	// guard needs it pre-fetch, and reading it twice could straddle a concurrent
+	// SetProjectionSpec and let the two decisions disagree within one tick.
+	// Consequence: a spec swap landing DURING the fetch is now noticed on the
+	// next tick rather than this one. That is self-healing — the next tick's
+	// e.SpecHash != currentSpecHash comparison forces the rebuild regardless.
+	currentSpecHash := ""
+	if spec := c.snapshotSpec(); spec != nil {
+		currentSpecHash = spec.Hash
+	}
+
+	// Offer the stored checksum as a conditional-fetch validator only when an
+	// "unchanged" answer would have led to the fast-skip below anyway — i.e.
+	// when the body is genuinely not needed for anything else this tick:
+	//   - UserCPRef == nil: no authored CP to re-fetch and re-adopt.
+	//   - UserCPRV == "": no authored RV on record either. Without this,
+	//     an entry in the authoredJustDropped shape (a recorded authored RV but
+	//     no authored CP any more) could skip that handling. It mirrors
+	//     rvsMatchCP(nil, e.UserCPRV) in the fast-skip, which is true only for "".
+	//   - SpecHash == currentSpecHash: the projection would be identical.
+	//   - Checksum != "": we actually hold a validator to offer.
+	//   - State is already terminal: see below.
+	//
+	// The state conjunct is not redundant with the others. e.State is derived
+	// from the StatusMetadataKey/CompletionMetadataKey ANNOTATIONS, which sit
+	// outside the content checksum — so a lifecycle flip (partial -> full)
+	// leaves the checksum matching. Without this conjunct such an entry would
+	// answer "unchanged" on that tick AND on every later one (its checksum
+	// stays valid indefinitely), so the rebuild that refreshes e.State would
+	// never fire and the cached state would freeze permanently. That is not
+	// the bounded one-fetch staleness accepted for e.RV: it never self-corrects.
+	// It matters because e.State is not internal bookkeeping — rulemanager gates
+	// HasFinalApplicationProfile on Completed+Full and stamps FailOnProfile on
+	// every alert from it, so a frozen state keeps alerting a completed profile
+	// as partial forever.
+	//
+	// Requiring Completed+Full (rather than isTerminalCPStatus, which also
+	// admits TooLarge) is deliberate: it is the exact predicate rulemanager
+	// treats as final, and it is the only state from which no further lifecycle
+	// transition is expected. A TooLarge profile simply keeps fetching bodies.
+	//
+	// Attached per call, never to the shared ctx: the authored-CP fetch below
+	// derives from the same ctx and must never carry the learned CP's checksum.
+	cpCtx := ctx
+	if e.UserCPRef == nil && e.UserCPRV == "" && e.SpecHash == currentSpecHash && e.Checksum != "" &&
+		e.State != nil && e.State.Status == helpersv1.Completed && e.State.Completion == helpersv1.Full {
+		cpCtx = storage.WithKnownChecksum(ctx, e.Checksum)
+	}
+
 	// Re-fetch all sources. CP fetch errors (including 404) are treated as
 	// "not available right now" — mirroring tryPopulateEntry's behavior. We
 	// leave cp=nil and rely on the RV-match fast-skip below to preserve the
@@ -339,11 +391,24 @@ func (c *ContainerProfileCacheImpl) refreshOneEntry(ctx context.Context, id stri
 	// while the storage-side consolidated CP remains unpublished.
 	var cp *v1beta1.ContainerProfile
 	var cpErr error
-	_ = c.refreshRPC(ctx, func(rctx context.Context) error {
+	_ = c.refreshRPC(cpCtx, func(rctx context.Context) error {
 		cp, cpErr = c.storageClient.GetContainerProfile(rctx, ns, e.CPName)
 		return cpErr
 	})
 	if cpErr != nil {
+		// Checked before IsNotFound: an implementation is free to return a
+		// sentinel that also carries a not-found shape, and "unchanged" is the
+		// more specific claim. The source verified our checksum still matches
+		// and sent no body, so the entry we hold is known-good — leave it
+		// entirely alone. Deliberately does NOT refresh e.RV or e.Checksum:
+		// there is no fresh object to read them from, so e.RV may lag a
+		// metadata-only write until the next unconditional fetch.
+		if errors.Is(cpErr, storage.ErrProfileUnchanged) {
+			logger.L().Debug("refreshOneEntry: CP unchanged (checksum match); keeping cached entry without rebuild",
+				helpers.String("containerID", id),
+				helpers.String("cpName", e.CPName))
+			return
+		}
 		if !apierrors.IsNotFound(cpErr) {
 			logger.L().Debug("refreshOneEntry: CP fetch failed transiently; keeping cached entry",
 				helpers.String("containerID", id),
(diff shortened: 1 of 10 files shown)


pkg/objectcache/containerprofilecache/reconciler.go
```diff
@@ -339,11 +391,24 @@ func (c *ContainerProfileCacheImpl) refreshOneEntry(ctx context.Context, id stri
...
 	})
 	if cpErr != nil {
+		// Checked before IsNotFound: an implementation is free to return a
+		// sentinel that also carries a not-found shape, and "unchanged" is the
+		// more specific claim. The source verified our checksum still matches
+		// and sent no body, so the entry we hold is known-good — leave it
+		// entirely alone. Deliberately does NOT refresh e.RV or e.Checksum:
+		// there is no fresh object to read them from, so e.RV may lag a
+		// metadata-only write until the next unconditional fetch.
+		if errors.Is(cpErr, storage.ErrProfileUnchanged) {
```
jnathangreeg: **The sentinel is honoured for every entry, including the ones the guard above deliberately declined to give a validator.** `cpCtx` records that decision and is available for free, but the `errors.Is` check ignores it.

Concrete misbehaviour: an adapter that mis-maps an error class, or a server that compares `req.known_checksum == stored.checksum` and matches `"" == ""`, returns the sentinel on an *unconditional* GET. This branch then returns early for entries that are still `ready`/`partial`, for entries whose CP was deleted (which would otherwise take the `cp == nil && userDefinedCP == nil` eviction path), and for entries with an authored CP — all frozen permanently, because the same conditions recur on every tick.

The PR doc calls a sentinel-without-validator "a protocol violation" and puts the burden on the remote implementer. Worth keeping that expectation *and* not depending on it — kubescape/backend#62's client rejects `unchanged` with an empty `KnownChecksum`, but this repo shouldn't rely on a sibling repo's guard for its own cache integrity. One boolean does it:

```go
sentChecksum := cpCtx != ctx
...
if sentChecksum && errors.Is(cpErr, storage.ErrProfileUnchanged) {
```

**Related, and the reason I'd want it before step 4b rather than after:** there's no bound on consecutive "unchanged" answers and no forced revalidation. This path deliberately refreshes neither `e.RV` nor `e.Checksum`, so one incorrect "unchanged" — a server checksum computed over a stale snapshot, a row mixup (see the `agents_profiles` duplicate-row case I raised on #1628), an adapter bug — pins that container's enforced profile for the container's lifetime with no self-correction, unlike the RV path which re-validates against a real body every tick. The plan lists "periodic full resync" as remaining design work; this is the line where it lands. Forcing an unconditional fetch every N ticks is cheap insurance given `e.State` and `Projected` feed rule enforcement.

pkg/objectcache/containerprofilecache/reconciler.go
```diff
@@ -331,6 +333,56 @@ func (c *ContainerProfileCacheImpl) refreshOneEntry(ctx context.Context, id stri
...
+	// Snapshot the projection spec ONCE, before the fetch, and reuse it for both
+	// the conditional-fetch guard below and the fast-skip further down. The
+	// guard needs it pre-fetch, and reading it twice could straddle a concurrent
+	// SetProjectionSpec and let the two decisions disagree within one tick.
+	// Consequence: a spec swap landing DURING the fetch is now noticed on the
+	// next tick rather than this one. That is self-healing — the next tick's
+	// e.SpecHash != currentSpecHash comparison forces the rebuild regardless.
+	currentSpecHash := ""
+	if spec := c.snapshotSpec(); spec != nil {
+		currentSpecHash = spec.Hash
```
jnathangreeg: Hoisting `currentSpecHash` above the two fetches changes behaviour for **all** entries, not just conditional-fetch ones. If `SetProjectionSpec` lands while an entry is mid-fetch (up to 2×`rpcBudget`, 5s each by default), the fast-skip further down now compares against the pre-swap hash, matches, and skips — leaving that container projected under the *old* rule-projection spec.

The accompanying nudge can't repair it: `SetProjectionSpec` sets `refreshPending` and CASes `refreshInProgress`, but the in-flight refresh was launched by `tickLoop`'s ticker branch, whose goroutine is a bare `refreshAllEntries(ctx)` with no `for c.refreshPending.Swap(false)` drain — so the CAS fails and the flag is never consumed. Recovery waits for the next tick (30s default), during which that container evaluates rules against a stale projection.

It affects at most one entry per spec change, and reading the hash at the comparison site instead costs the guard nothing — the guard's own pre-fetch read can stay.

pkg/objectcache/containerprofilecache/reconciler.go
```diff
@@ -339,11 +391,24 @@ func (c *ContainerProfileCacheImpl) refreshOneEntry(ctx context.Context, id stri
 	// while the storage-side consolidated CP remains unpublished.
 	var cp *v1beta1.ContainerProfile
 	var cpErr error
-	_ = c.refreshRPC(ctx, func(rctx context.Context) error {
+	_ = c.refreshRPC(cpCtx, func(rctx context.Context) error {
```
jnathangreeg, line 439: No metric distinguishes "validator offered", "source answered unchanged", and "no validator held", and the sentinel is logged at Debug only. The contract doc states that a cross-repo annotation-key mismatch "fails silently rather than loudly… the cache simply never observes a checksum" — with no counter, that permanently inert state is indistinguishable in production from the mechanism working perfectly, which is exactly the failure mode the plan's step-3 canary is supposed to detect. `c.metricsManager` is already threaded through here; two counters make the silent-failure mode observable.

jnathangreeg (commented):
Reviewed against the other two steps rather than in isolation — kubescape/backend#62 (step 2, approved this morning) and armosec/postgres-connector#1628 (step 1).

Verified correct, so nobody re-checks the things most likely to be wrong in this design: the annotation key string is byte-identical to `backendv1.ContainerProfileChecksumAnnotationKey`; `checksumOfCP` is nil- and nil-map-safe and reads the annotation **verbatim** rather than rehashing the held object (which is the invariant that keeps this working at all, since the object now carries an annotation the server never hashed); the validator tracks the *learned* CP at all three construction sites, so it can never describe an object other than `e.CPName`; and the sentinel early-return is behaviourally equivalent to the pre-existing fast-skip. The contract doc is genuinely useful — the guard-conjunct table especially.

**One cross-step note that decides whether any of this can fire:** #62 now requires the wire checksum to be `sha256:`-tagged and refuses to stamp an untagged value, while step 1 (#1628) emits the stored checksum untagged. Until that's reconciled, `e.Checksum` stays empty forever and this code is inert — which is the intended state for now, but it means a step-3 canary would read 0% hits for a reason that has nothing to do with this PR.

jnathangreeg (approved):
Approving at `e3044f3`. All four points are closed, and two go further than the findings did.

- **The sentinel is now validated against the request** — `validatorOffered` gates it, and an unoffered sentinel isn't merely ignored: it's counted as `conditionalFetchOutcomeProtocolError`, logged at Warning with the container and CP names, and treated as a transient failure that leaves the entry alone. That's better than the boolean I suggested, because the misbehaving implementer becomes visible instead of silently costing one refresh.
- **Forced revalidation exists now.** `maxConsecutiveUnchangedResponses = 10` with `e.consecutiveUnchanged` incremented per unchanged answer and cleared only by a successful body — so at the default cadence no entry can go more than ~10 minutes without a real body, and a wrong "unchanged" self-corrects instead of pinning the container's enforced profile for its lifetime. This was the plan's outstanding "remaining design work", so it's worth noting it's landed here rather than deferred to 4b.
- **The projection-spec race is fixed at the root.** `preFetchSpecHash` for validator eligibility and `postFetchSpecHash` for the fast-skip is exactly the split that was needed — but the more valuable part is the trailing-edge scheduler: `refreshPending` is now actually consumed in a loop, so a nudge arriving during an in-flight refresh is no longer swallowed by a failed CAS. That was the underlying reason a mid-fetch `SetProjectionSpec` couldn't self-repair, and it's a pre-existing defect this PR now closes rather than works around.
- **The inert-vs-working ambiguity is observable.** Request-mode and response-outcome counters (body / unchanged / protocol error / not found) mean a cross-repo annotation-key mismatch or an untagged checksum shows up as a metric rather than as silence.

Verified rather than read off the commit message: each of the four in the source at this head, and `go test ./pkg/objectcache/containerprofilecache/... ./pkg/metricsmanager/... ./pkg/storage/` green in a Linux container.

Standing cross-step note, unchanged by this PR and not blocking it: this stays inert until the producer side emits an algorithm-tagged checksum. armosec/postgres-connector#1628 returns the stored value untagged by design and records that `cadashboardbe` owns conversion to `sha256:<64 lowercase hex>`, which kubescape/backend#62 (merged) requires before its client will stamp or trust a value. So if a step-3 canary reads 0% hits, look there first — the new counters here will now say which of "no validator held" or "answered body" it actually is.

PR #944, opened 2026-09-04
fix: remap Service egress ports to targetPort for NetworkPolicy generation (#943)

### LINKED ISSUE #943: Egress to a Service records the Service port, not the endpoint port the CNI enforces on
# Description

For egress to a Service whose `port` differs from its `targetPort`, the generated NetworkPolicy opens the **Service** port. CNIs that enforce on the resolved endpoint match the **targetPort**, so the rule never matches and the traffic is denied once a `default-deny` is in place.

# Environment

OS: `Amazon Linux 2023, kernel 6.12 (arm64)`
Version: `kubescape-operator chart 1.40.3, node-agent v0.3.215, storage v0.0.297`
CNI: `AWS VPC CNI 1.23, NetworkPolicy enforcement in standard mode`

# Steps To Reproduce

1. Expose a workload with a Service using `port: 80, targetPort: 8080`
2. Have a client call it through the Service, and wait for the learning window
3. `kubectl get generatednetworkpolicies -n <ns> <client> -o yaml`
4. Apply the generated policy together with a `default-deny` in the client's namespace

# Expected behavior

The rule opens `8080`, the port the CNI enforces on:

```yaml
egress:
  - to: [{podSelector: {matchLabels: {app.kubernetes.io/name: api}}}]
    ports: [{port: 8080, protocol: TCP}]
```

# Actual Behavior

The rule opens `80`, and the connection is denied once `default-deny` is in place:

```yaml
egress:
  - to: [{podSelector: {matchLabels: {app.kubernetes.io/name: api}}}]
    ports: [{port: 80, protocol: TCP}]
```

# Additional context

The agent observes the connection to the ClusterIP before DNAT, while enforcement happens on the endpoint after it. This one fails **closed** and the rule looks correct on review, so it is the most damaging of the behaviours we hit: generated policies cannot be applied unedited on such CNIs, and every `port != targetPort` Service in a cluster is affected. Resolving the Service to its endpoint port at generation time would fix it.

ANAMASGARD:
## Overview

When learning egress through a Kubernetes Service whose `port` differs from `targetPort` (e.g. Service `80` → Pod `8080`), eBPF observes the pre-DNAT Service port. Generated NetworkPolicies previously opened that Service port. CNIs that enforce on the resolved endpoint (such as AWS VPC CNI in standard mode) match the pod `targetPort`, so the rule never matches and traffic is denied once `default-deny` is applied.

**Before:** egress to Service `port: 80, targetPort: 8080` → `NetworkNeighbor.Ports` / GeneratedNetworkPolicy open **80**  
**After:** the same traffic → ports remapped to the enforcement port(s) (**8080**, or multiple resolved ports for named/heterogeneous backends)

At `createNetworkNeighbor` serialization time we now resolve:

| Service config | Result |
|---|---|
| Numeric `targetPort` (`80 → 8080`) | `[8080]` |
| Omitted `targetPort` | defaults to Service `port` |
| Named `targetPort` (`name: web`, `targetPort: http`) | resolve via EndpointSlice (match `ServicePort.Name`), Endpoints fallback |
| Heterogeneous named backends | collect, dedupe, sort all resolved ports (e.g. `[8080, 9090]`) |
| Lookup / convert failure | keep observed port (safe fallback) |
| Pod / raw destinations | unchanged |

## Additional Information

- Fix is in node-agent only. Storage/`GenerateNetworkPolicy` already passes `NetworkNeighbor.Ports` through; the wrong port was recorded upstream.
- Named-port resolution prefers EndpointSlices, then falls back to Endpoints (already in node-agent RBAC).
- **Companion PR needed in `kubescape/helm-charts`:** grant node-agent ClusterRole `get/list/watch` on `discovery.k8s.io/endpointslices`. Without it, EndpointSlice list returns 403 and Endpoints fallback is used.
- Size accounting in `networkNeighborIncrement` updated so Service-branch remapping is no longer treated as exact-from-event.

## How to Test

```bash
# Targeted regression (#943)
go test ./pkg/containerprofilemanager/v1/... \
  -run 'TestCreateNetworkNeighbor_ServiceTargetPortMatrix|TestGenerateNetworkPolicy_ServiceTargetPortRoundTrip' \
  -count=1 -v

# Affected package
go test ./pkg/containerprofilemanager/v1/... -count=1

# Repo-wide (per README)
go test ./... -count=1

# Race
go test -race ./pkg/containerprofilemanager/v1/... -count=1
```

Expected: matrix asserts `80 → 8080` (and `NotEqual` 80); round-trip asserts GeneratedNetworkPolicy egress port is **8080**.

Optional cluster check (not required for merge): Service `port: 80, targetPort: 8080`, learn egress, confirm GeneratedNetworkPolicy opens `8080`, apply with default-deny, verify client traffic succeeds under the CNI.

## Related issues/PRs

* Resolves #943
* Follow-up: companion `kubescape/helm-charts` PR for EndpointSlice RBAC on node-agent

Open the PR against **`dev`**, not `main`.

## Summary by CodeRabbit

- **New Features**
  - Network policies now resolve Kubernetes Service target ports more accurately, including numeric and named ports.
  - Added support for EndpointSlices and Endpoints, multiple matching ports, and TCP, UDP, and SCTP protocols.
  - Service traffic can map observed ports to the correct enforcement ports, including heterogeneous backend ports.

- **Bug Fixes**
  - Improved fallback handling for missing, malformed, unknown, or mismatched service port definitions.
  - Non-Service network destinations retain their existing behavior.
  - Improved port ordering and deduplication in generated network policies.

### DIFF
--- a/pkg/containerprofilemanager/v1/container_data.go
+++ b/pkg/containerprofilemanager/v1/container_data.go
@@ -6,12 +6,12 @@ import (
 	mapset "github.com/deckarep/golang-set/v2"
 	"github.com/kubescape/go-logger"
 	"github.com/kubescape/go-logger/helpers"
+	"github.com/kubescape/k8s-interface/k8sinterface"
 	"github.com/kubescape/node-agent/pkg/dnsmanager"
 	"github.com/kubescape/node-agent/pkg/k8sclient"
 	"github.com/kubescape/node-agent/pkg/utils"
 	"github.com/kubescape/storage/pkg/apis/softwarecomposition/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/utils/ptr"
 )
 
 // emptyEvents clears all event data
@@ -202,12 +202,8 @@ func (cd *containerData) getEgressNetworkNeighbors(containerID string, namespace
 func (cd *containerData) createNetworkNeighbor(containerID string, networkEvent NetworkEvent, namespace string, k8sClient k8sclient.K8sClientInterface, dnsResolverClient dnsmanager.DNSResolver) *v1beta1.NetworkNeighbor {
 	var neighborEntry v1beta1.NetworkNeighbor
 
-	portIdentifier := generatePortIdentifierFromEvent(networkEvent)
-	neighborEntry.Ports = []v1beta1.NetworkPort{{
-		Protocol: v1beta1.Protocol(networkEvent.Protocol),
-		Port:     ptr.To(int32(networkEvent.Port)),
-		Name:     portIdentifier,
-	}}
+	enforcementPorts := []uint16{networkEvent.Port}
+	var serviceWorkload k8sinterface.IWorkload
 
 	if networkEvent.Destination.Kind == EndpointKindPod {
 		// For Pods, we need to remove the default labels
@@ -230,6 +226,7 @@ func (cd *containerData) createNetworkNeighbor(containerID string, networkEvent 
 				helpers.String("service name", networkEvent.Destination.Name))
 			return nil
 		}
+		serviceWorkload = svc
 
 		var selector map[string]string
 		if svc.GetName() == "kubernetes" && svc.GetNamespace() == "default" {
@@ -269,6 +266,18 @@ func (cd *containerData) createNetworkNeighbor(containerID string, networkEvent 
 			}
 		}
 	}
+
+	if networkEvent.Destination.Kind == EndpointKindService && serviceWorkload != nil && k8sClient != nil {
+		enforcementPorts = resolveServiceEnforcementPorts(
+			k8sClient,
+			networkEvent.Destination.Namespace,
+			networkEvent.Destination.Name,
+			serviceWorkload,
+			networkEvent.Port,
+			networkEvent.Protocol,
+		)
+	}
+	neighborEntry.Ports = buildNetworkPorts(networkEvent.Protocol, enforcementPorts)
 
 	neighborEntry.Type = InternalTrafficType
 	if neighborEntry.NamespaceSelector == nil && neighborEntry.PodSelector == nil {
--- a/pkg/containerprofilemanager/v1/container_data.go
+++ b/pkg/containerprofilemanager/v1/container_data.go
@@ -25,6 +25,7 @@ func (cd *containerData) emptyEvents() {
 	cd.rulePolicies = nil
 	cd.callStacks = nil
 	cd.networks = nil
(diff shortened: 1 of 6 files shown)


matthyx (changes requested):
Requesting changes for two correctness issues that are still present at `02e90171`:

1. `collectEndpointsPorts` dereferences `endpointsObj` without checking it. After an empty EndpointSlice lookup, a normal `(nil, nil)` result from `GetWorkload("Endpoints", ...)` will panic instead of falling back to the observed Service port. Please guard the nil object and add a regression test for that result.

2. `networkNeighborIncrement` budgets the original port plus only one additional `NetworkPort`, while named-port resolution can return any number of distinct ports across EndpointSlices. Three or more resolved ports therefore undercount `entry.data.size` and can delay the `MaxTsProfileSize` split. Please make the accounting cover the actual resolved multiplicity or a defensible upper bound, with a test using at least three distinct endpoint ports.

The targeted tests and the full `pkg/containerprofilemanager/v1/...` package suite pass locally, but neither test currently covers these failure modes.

ANAMASGARD:
@matthyx I have resolved the merge conflict and implemented the two changes you asked for please review this PR once again and give your feedback , thanks .

matthyx (changes requested):
The two previously requested fixes are present at `b10f215e`: the nil Endpoints fallback is guarded and the full resolved-port multiplicity is snapshotted, charged, and covered by regression tests.

This revision still cannot be approved because the merge from `main` dropped the closing brace for `TestReportNetworkEventServicePortMultiplicity` before `TestCreateNetworkNeighbor_StatefulSetPeerStripsPodIdentityLabels` in `pkg/containerprofilemanager/v1/event_reporting_test.go`. As a result, the package does not compile:

```
pkg/containerprofilemanager/v1/event_reporting_test.go:321:6: expected '(', found TestCreateNetworkNeighbor_StatefulSetPeerStripsPodIdentityLabels
```

Please restore the function-closing `}` after the final `require.Less(...)` assertion and rerun the package suite.

matthyx (approved):
Re-reviewed at `ae5c7fc0`. The merge-induced syntax error is fixed, and the two original correctness issues remain resolved with regression coverage:

- nil Endpoints fallback no longer panics;
- all resolved Service endpoint ports are consistently snapshotted, serialized, and included in size accounting.

`go test ./pkg/containerprofilemanager/v1/... -count=1` passes locally. The component workflow completed successfully overall; its advisory failing matrix jobs show Kubernetes API/server timeouts and do not point to this change. No remaining blocker found.

PR #946, opened 2026-09-05
  fix: make SBOM failure service discovery opt-in

### LINKED ISSUE #899: Avoid remote service discovery by default
## Description

Currently, node-agent defaults `API_URL` to `api.armosec.io` and calls `LoadServiceURLs()`:

https://github.com/kubescape/node-agent/blob/300b59e25b73bc2a697ab3fe7af09b8bc2aa92e1/cmd/main.go#L421-L424

In [LoadServiceURLs()](https://github.com/kubescape/node-agent/blob/300b59e25b73bc2a697ab3fe7af09b8bc2aa92e1/pkg/config/config.go#L333), if services.json does not exist, node-agent performs remote service discovery:

As a result, node-agent may access a public endpoint even when the user has not explicitly configured remote service discovery.

It would be better to perform remote service discovery only when it is explicitly enabled or `API_URL` is configured.

This would make network behavior more predictable, especially in restricted-egress and air-gapped environments.

### Expected behavior

Remote service discovery should be explicitly configurable.

For example:

```json
{
  "sbomGenerationEnabled": true,
  "sbomFailureReportingEnabled": false
}
```

When `sbomFailureReportingEnabled` is false, node-agent should not call `LoadServiceURLs()` for SBOM failure reporting.

When it is `true`, node-agent may load service information from services.json or perform remote service discovery if necessary.

ANAMASGARD:
## Overview

  Prevent node-agent from performing remote service discovery for SBOM failure reporting unless it is explicitly enabled.

  This adds the default-disabled `sbomFailureReportingEnabled` configuration option. Service discovery now runs only when both:

  - `sbomGenerationEnabled` is `true`
  - `sbomFailureReportingEnabled` is `true`

  When enabled, the existing discovery behavior is preserved:

  1. Use `SERVICES` or `/etc/config/services.json` when available.
  2. Otherwise use `API_URL`.
  3. Fall back to `api.armosec.io` when `API_URL` is unset.

  Setting `API_URL` or mounting `services.json` alone does not enable failure reporting. Discovery errors and empty receiver URLs remain
  non-fatal.

  ## Additional Information

  This changes SBOM failure reporting from implicit behavior to explicit opt-in behavior. Existing deployments that require failure reporting must add:

  ```json
  {
    "sbomGenerationEnabled": true,
    "sbomFailureReportingEnabled": true
  }
  ```

  ## How to Test

  Focused tests:

```
  go test -count=1 ./pkg/config ./cmd
  go test ./pkg/sbommanager/v1
```

  These checks pass locally.

  A serial non-e2e repository test sweep was also performed. The remaining unrelated environment-dependent failures were:

  - Tracer field tests require a generated root-level tracers.tar.
  - pkg/validator requires permission to raise the memlock limit.

  Socket-dependent exporter, node-profile, and SBOM-scanner tests passed when run outside the restricted sandbox.

  ## Related issues/PRs

  - Resolves #899

  - [x] I have commented on my code where necessary
  - [x] I have added thorough regression tests
  - [x] Focused unit tests pass locally with my changes

## Summary by CodeRabbit

* **New Features**
  * Added an opt-in setting for reporting SBOM generation failures.
  * SBOM failure reporting remains disabled by default and is activated only when both SBOM generation and failure reporting are enabled.
  * Failure report destinations can be resolved from local service configuration or the configured API endpoint.

* **Documentation**
  * Documented the new configuration option, activation requirements, and service discovery behavior.
  * Added a link to the Kubescape documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,10 +262,13 @@ See [docs/CONFIGURATION.md](docs/CONFIGURATION.md) for the complete configuratio
 | Malware Detection | `malwareDetectionEnabled` | `false` | Start the malware manager (needs an out-of-tree scanner) |
 | Network Tracing | `networkServiceEnabled` | `false` | Track network connections |
 | SBOM Generation | `sbomGenerationEnabled` | `false` | Generate SBOMs |
+| SBOM Failure Reporting | `sbomFailureReportingEnabled` | `false` | Report SBOM generation failures; service discovery runs only when this and SBOM generation are enabled |
 | File Integrity | `fimEnabled` | `false` | Monitor file changes |
 | Seccomp Profiles | `seccompServiceEnabled` | `false` | Generate seccomp profiles |
 | HTTP Detection | `httpDetectionEnabled` | `false` | Parse HTTP traffic |
 | Network Streaming | `networkStreamingEnabled` | `false` | Stream network events |
+
+SBOM failure reporting is opt-in. Setting `API_URL` or mounting `services.json` does not enable it; those sources are consulted only when both `sbomGenerationEnabled` and `sbomFailureReportingEnabled` are `true`.
 
 ## 🔌 Image-Based Gadgets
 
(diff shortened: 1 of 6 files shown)


matthyx (approved):
Reviewed the opt-in gating and service-discovery fallback behavior. The implementation matches #899, avoids discovery unless both SBOM generation and failure reporting are enabled, and preserves the existing local-file/API fallback when opted in. Focused tests and go vet pass; no blockers found.

PR #948, opened 2026-09-07
[EXPERIMENT] Validate kubescape/storage#397 (sharded single-writer) fix

matthyx:
## Purpose

**This is a validation experiment, not a proposed fix — do not merge.**

Points the test storage deployment at a manually-built image (`quay.io/kubescape/storge:test-sharded-writer-397` — note: pre-existing typo in kubescape/storage's `build-image` workflow pushes to `storge`, not `storage`), built from [kubescape/storage#397](https://github.com/kubescape/storage/pull/397) (`fix/single-writer-sharded-commits`), to empirically confirm the sharded single-writer fix resolves this repo's `component-tests` regression (18-19 of ~30 matrix jobs failing since storage's `singleWriterEnabled` default flipped to `true`, see #397's description for the full root-cause writeup) before that PR merges upstream.

Expected result: **near-zero failures**, matching the pre-regression baseline, since the fix restores per-key write parallelism (default 8 shards) while keeping every existing single-writer correctness guarantee.

## Test plan
- [ ] component-tests run on this PR — compare failure count against the current baseline (18-19/30 failing) and the unmodified-storage-image baseline (0-1/30)

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01LLN3CgGftcnAikV13qD6yJ

AI-skills: none | cmds: /oh-my-claudecode:autopilot

PR #949, opened 2026-09-07
[EXPERIMENT] Validate kubescape/storage#397 (sharded single-writer) fix, take 2

matthyx:
## Purpose

**This is a validation experiment, not a proposed fix — do not merge.**

Retry of #948, which was a false negative: that attempt's test image landed in `quay.io/kubescape/storge` (typo'd repo name from a pre-existing bug in kubescape/storage's manual `build-image` workflow), which turned out to be a brand-new, private-by-default repo. The Kind cluster couldn't pull it at all — the storage deployment never became ready, so all 31 test jobs failed on infrastructure grounds, not on anything related to the write-throughput regression.

This retry uses a corrected build (`quay.io/kubescape/storage:test-sharded-writer-397b` — confirmed public via `curl https://quay.io/api/v1/repository/kubescape/storage/tag/` with no auth), built from [kubescape/storage#397](https://github.com/kubescape/storage/pull/397)'s `fix/single-writer-sharded-commits` via a throwaway build-workflow branch that fixes the repo-name typo and drops the `:latest` tag push so the real published tag is never touched.

Expected result: **near-zero failures**, matching the pre-regression baseline, since the fix restores per-key write parallelism (default 8 shards) while keeping every existing single-writer correctness guarantee.

## Test plan
- [ ] component-tests run on this PR — compare failure count against the current baseline (18-19/30 failing)

Claude-Session: https://claude.ai/code/session_01LLN3CgGftcnAikV13qD6yJ

AI-skills: none | cmds: /oh-my-claudecode:autopilot

## Summary by CodeRabbit

* **Tests**
  * Component tests now use a fixed experimental storage image version.
  * Updated test output to reflect the selected experimental version.

### DIFF
--- a/.github/workflows/component-tests.yaml
+++ b/.github/workflows/component-tests.yaml
@@ -153,8 +153,11 @@ jobs:
           kubectl wait --for=condition=Ready pod -l app.kubernetes.io/name=prometheus -n monitoring --timeout=300s
       - name: Install Node Agent Chart
         run: |
-          STORAGE_TAG=$(./tests/scripts/storage-tag.sh)
-          echo "Storage tag that will be used: ${STORAGE_TAG}"
+          # EXPERIMENT: validating kubescape/storage#397 (sharded single-writer
+          # fix) before merge. Pins storage to a manually-built test image
+          # instead of the usual dynamic tag. Do not merge this override.
+          STORAGE_TAG=test-sharded-writer-397b
+          echo "Storage tag that will be used: ${STORAGE_TAG} (EXPERIMENT override)"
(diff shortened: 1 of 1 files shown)


matthyx:
Result: 15/30 failed (build-and-push-image succeeded, so 15/30 test jobs, 19/32 total including infra jobs). Real, substantial improvement over the 18-19/30 baseline — no more sustained 20-min hangs, storage deployment comes up cleanly, individual container-profile writes went from minutes-long timeouts to 5-9 seconds for a 16-item burst. But not the hoped-for near-zero.

Traced the residual failures: still occasional `context deadline exceeded` retries plus a new error variant (`sqlite: step: interrupted`) under concurrent write bursts, self-healing within ~1-2 minutes rather than hanging for the full test timeout. Also observed the test pod's containers getting recreated multiple times within about a minute, adding extra write churn independent of storage.

Follow-up experiment testing higher shard/pool headroom (16 shards / 24 pool connections) in #950 before concluding whether this needs a higher default or has a separate contributing cause.

PR #952, opened 2026-09-07
[EXPERIMENT] Test storage pod CPU starvation hypothesis for residual component-tests flakiness

matthyx:
## Purpose

**Diagnostic experiment, not a proposed fix — do not merge.**

Follow-up to the storage single-writer investigation (kubescape/storage#397, released v0.0.333). That fix eliminated the sustained 20-minute hangs, but component-tests still shows an elevated failure rate: 15-17/30 jobs failing (confirmed fresh today on #951: 17/31), vs a historical pre-regression baseline of 0-1/30.

#950 tested whether the residual failures were a shard/pool headroom problem (singleWriterShards 8→16, sqlitePoolSize 10→24) — result was 21/32 failed, *worse*, not better. That rules out shard/pool count as the lever and points at the storage pod's CPU quota instead: `tests/chart/values.yaml`'s `storage.resources` has stayed at `requests.cpu: 100m` / `limits.cpu: 500m` across every experiment so far. Under a starved CPU quota, more concurrent write-path goroutines (the 8 fixed shards) add scheduling overhead without proportional throughput gain, which would explain why more shards didn't help.

This PR bumps `storage.resources.requests.cpu` 100m→500m and `limits.cpu` 500m→2000m, with **no shard-count or other change**, to isolate this variable. Looking for:
- Overall failure count vs the 17/31 baseline (#951, same day)
- Whether storage-side `context deadline exceeded` / `sqlite: step: interrupted` errors specifically disappear from logs (vs the heterogeneous non-storage assertion failures, which this change should NOT affect)

## Test plan
- [ ] component-tests run on this PR — compare failure count and error signature against #951's 17/31 baseline

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01LCMGT6Po2tSr1VEDVrbbYd

https://claude.ai/code/session_01LCMGT6Po2tSr1VEDVrbbYd

AI-skills: none

## Summary by CodeRabbit

- **Chores**
  - Increased the CPU capacity allocated to the storage component.
  - Memory allocation remains unchanged.

PR #951, opened 2026-09-07
[EXPERIMENT] Validate real kubescape/storage v0.0.333 release

matthyx:
## Purpose

**Throwaway validation PR — do not merge.**

Prior validations ([#949](https://github.com/kubescape/node-agent/pull/949), [#950](https://github.com/kubescape/node-agent/pull/950)) used a manually-built test image from the [kubescape/storage#397](https://github.com/kubescape/storage/pull/397) fix branch. That PR has since merged and released as [v0.0.333](https://github.com/kubescape/storage/releases/tag/v0.0.333), and `kubescape/helm-charts` main now points `storage.image.tag` at it.

This PR makes no functional change — `tests/scripts/storage-tag.sh`'s normal dynamic resolution will pick up `v0.0.333` on its own. Opening this purely to trigger `component-tests` and confirm the real, merged, released fix behaves the same as the manually-built validation image did (expect ~15/30 failures, down from the 18-19/30 baseline, no sustained hangs).

## Test plan
- [ ] component-tests run on this PR — compare against #949's 15/30 result

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01LLN3CgGftcnAikV13qD6yJ

AI-skills: oh-my-claudecode:cancel | cmds: /oh-my-claudecode:autopilot

## Summary by CodeRabbit

* **Tests**
  * Added a validation-only update to component test workflow configuration. No user-facing product behavior has changed.

PR #953, opened 2026-09-07
[EXPERIMENT] Instrument component-tests for pod restarts and node resource pressure

matthyx:
## Purpose

**Diagnostic experiment, not a proposed fix — do not merge.**

Follow-up to the CPU-headroom experiment (#952, refuted): bumping the storage pod's own CPU limit 4x (500m→2000m) made no measurable difference to the residual 17-18/31 component-tests failure rate. Tracing one failing job (Test_01_BasicAlertTest) end-to-end showed the container-profile *write* path timing out repeatedly (`failed to create container profile, requeuing` / `context deadline exceeded`), with the internal write queue backing up — independent of the storage pod's own cgroup quota.

Two candidates remain, neither confirmed with hard data yet:
1. **Node-level CPU oversubscription** on the shared GitHub-hosted runner — the storage pod's own limit can't manufacture free cycles if the whole node (Kind control-plane + storage + node-agent + kube-prometheus-stack + kubelet/containerd + `go test`) is already saturated.
2. **CI pod churn/restarts** (liveness-probe failures under CI resource pressure) generating extra write traffic independent of storage, per the original investigation's (#949) observation of multiple container-instance IDs for the same logical container within ~1 minute.

This PR makes no behavioral change — it only adds an always-run diagnostic step after the existing log dump, capturing:
- `kubectl describe nodes` (allocatable vs. requested resources, to check for node-level saturation)
- `kubectl get pods -A -o wide` (restart counts across the whole cluster)
- `kubectl describe pod` for any pod with a non-zero restart count
- `kubectl get events -A --sort-by=.lastTimestamp` (liveness/readiness probe failures, OOMKilled, etc.)

...across the full ~30-job matrix, so failing vs. passing jobs can be correlated against real restart/pressure evidence instead of one manually-traced job.

## Test plan
- [ ] component-tests run on this PR — correlate failing jobs against restart counts / node pressure evidence in the new diagnostic step

AI-skills: none

## Summary by CodeRabbit

* **Chores**
  * Added always-run diagnostic reporting to component test workflows, including node capacity, pod status and restart details, and time-sorted cluster events.

### DIFF
--- a/.github/workflows/component-tests.yaml
+++ b/.github/workflows/component-tests.yaml
@@ -182,3 +182,22 @@ jobs:
           echo "-----------------------------------------"
           echo "Storage logs"
           kubectl logs $(kubectl get pods -n kubescape -o name | grep storage) -n kubescape
+      - name: Diagnose pod restarts and node resource pressure
+        if: always()
+        run: |
+          echo "Node describe (allocatable/capacity vs requests)"
+          kubectl describe nodes
+          echo "-----------------------------------------"
+          echo "Pods across all namespaces, with restart counts"
+          kubectl get pods -A -o wide
+          echo "-----------------------------------------"
+          echo "Pods with non-zero restarts (detail)"
+          for p in $(kubectl get pods -A --no-headers -o custom-columns=":metadata.namespace,:metadata.name,:status.containerStatuses[0].restartCount" | awk '$3!="" && $3!="<none>" && $3+0>0 {print $1"/"$2}'); do
+            ns=$(echo "$p" | cut -d/ -f1)
+            name=$(echo "$p" | cut -d/ -f2)
+            echo "=== $ns/$name ==="
+            kubectl describe pod "$name" -n "$ns"
+          done
+          echo "-----------------------------------------"
+          echo "Cluster events, sorted by time (liveness/readiness probe failures, OOMKilled, etc.)"
+          kubectl get events -A --sort-by=.lastTimestamp


matthyx:
Result: 20/31 failed this run (noisy vs. #951's 17/31 and #952's 18/31, but same band — no systematic shift).

**Both infrastructure hypotheses are now refuted with direct evidence, checked across 4 failing jobs (Test_01, Test_02, Test_14, Test_27):**

1. **Node-level CPU oversubscription: refuted.** `kubectl describe nodes` shows the Kind node has **4 CPUs** capacity/allocatable, with only **1150m requested (28%) / 1100m limited (27%)** across the entire cluster (Kind control-plane, storage, node-agent, full kube-prometheus-stack, test workloads). There's no scheduling-level saturation — 72% of the node's CPU reservation is unused. (This doesn't rule out *actual* usage spikes beyond requests, which we can't see without a metrics-server, but the reservation-level picture doesn't support the "node has no spare capacity" theory.)

2. **Pod restarts / CI churn: refuted.** Across all 4 failing jobs checked, `storage`, `node-agent`, and every test workload pod show `RESTARTS: 0`. Cluster-wide `kubectl get events` shows **zero** `Unhealthy`/`BackOff`/`OOMKilled`/`Killing` events in any of them. The multiple-container-instance-ID pattern observed in #949 was not reproduced here — no evidence of pod churn in this data.

So neither of the two leading hypotheses from the original investigation explains the residual ~17-20/31 failure rate. Per the original task's step 4, this is now a fresh investigation. The concrete lead worth pursuing next: node-agent's own container-profile write path is timing out and requeuing (`failed to create container profile, requeuing` / `context deadline exceeded`, write queue backing up to `queueSize: 5`+) against a storage pod that is itself neither CPU-starved (by request/limit) nor restarting — meaning the bottleneck is likely *inside* the request path itself (e.g., actual transient CPU/IO burst not visible in the requests view, lock contention in a specific SQLite shard, the aggregated apiserver's per-request overhead, or the client-side context timeout being too short for legitimate but bounded processing delay) rather than anything captured by infra-level restarts or cgroup limits.

Not pursuing further blind parameter experiments. Closing this diagnostic PR; the kept diagnostic step (node describe + restart table + events) is a useful pattern worth re-adding permanently or reusing case-by-case, but not merging into main as-is since it's throwaway.

PR #977, opened 2026-09-18
fix(tracers): detect TTY on /dev/pts/0

### LINKED ISSUE #975: `hasTty` is false on /dev/pts/0: bump trace_exec gadget to v0.55.0+
# Description
Hi 👋, and thanks for the TTY fields and the very clear `docs/features/exec-tty-field.md`!

`event.hasTty` is false for a shell on `/dev/pts/0`, which is where the first `kubectl exec -it` into a container lands. node-agent pins `ghcr.io/inspektor-gadget/gadget/trace_exec:v0.48.1` (`pkg/containerwatcher/v2/tracers/exec.go`), which only emits the tty index, so `has(event.ttyMajor)` is never true and `hasTty` falls back to an index where `0` means both "no terminal" and `/dev/pts/0`.

Good news: inspektor-gadget/inspektor-gadget#5714 landed in v0.55.0 and emits the terminal device number. Since node-agent already reads `tty_major` when it's present, a gadget bump might be all it takes.

# Environment
OS: `Amazon Linux 2023 (EKS 1.34), kernel 6.12, arm64, containerd 2.2`
Version: `node-agent v0.3.219 (kubescape-operator chart 1.40.4)`

# Steps To Reproduce
1. Bind these three rules to a namespace running a `busybox` pod (`sleep` as its command):
   ```yaml
   - id: TLX901 # control
     expressions:
       message: "'control: ' + event.comm + ' tty=' + string(event.tty)"
       uniqueId: "event.comm + '_' + string(event.pid)"
       ruleExpression:
         - eventType: exec
           expression: 'event.comm == "sh"'
   - id: TLX902
     # same, with expression: 'event.comm == "sh" && event.hasTty'
   - id: TLX903
     # same, with expression: 'event.comm == "sh" && has(event.ttyMajor)'
   ```
2. `kubectl exec -it <pod> -- sh -c 'tty; sleep 3'` → prints `/dev/pts/0`
3. `kubectl exec -it <pod> -- sh -c 'tty; sleep 5'` while a pts index is still held → prints `/dev/pts/1`

# Expected behavior
All three rules fire for both shells: each has a terminal (major 136).

# Actual Behavior
| Terminal | TLX901 control | TLX902 `hasTty` | TLX903 `has(ttyMajor)` | `tty` |
|---|---|---|---|---|
| `/dev/pts/0` | fired | not fired | not fired | 0 |
| `/dev/pts/1` | fired | fired | not fired | 1 |

# Additional context
This matches the phase-1 expectations in `Test_35_ExecTTYFieldTest` (`c-pts0=0`), and v0.3.231 still pins the same gadget. Our use case is a rule reporting interactive shells that stays quiet for containers whose start command is a shell (`sh -c ...` in CronJobs). For now we keep a per-workload exception list, which we'd happily drop. Happy to help test a bump. Thanks!

ANAMASGARD:
## Related issues/PRs

- Fixes #975
- Related upstream fix: inspektor-gadget/inspektor-gadget#5714

## Overview

`trace_exec:v0.48.1` reports only the driver-local TTY index. This makes `/dev/pts/0` indistinguishable from a process without a terminal because both report `tty=0`, causing `event.hasTty` to incorrectly evaluate to `false`.

This change pins only the `trace_exec` gadget to `v0.55.0`, which emits the terminal device major and minor numbers.

After this change:

- `event.hasTty` is `false` for processes without a TTY.
- `event.hasTty` is `true` for `/dev/pts/0`.
- `event.hasTty` is `true` for nonzero PTY indices.
- `has(event.ttyMajor)` is true for all emitted exec events, including no-TTY events where the major value is zero.

The strict exec schema test, component rules, acceptance assertions, and documentation were updated accordingly. No public CEL fields or Go module dependencies were changed.

## Additional Information

- `advise_seccomp`, `trace_capabilities`, `trace_dns`, and `trace_open` remain on their existing versions.
- The schema test now includes all five fields introduced by `trace_exec:v0.55.0`:
  `ctime`, `fctime`, `pctime`, `tty_major`, and `tty_minor`.
- Mixed-version guidance remains documented for agents bundling older gadgets.
- `go.mod`, `go.sum`, and workflows are unchanged.

## How to Test

```bash
make gadgets

GOTOOLCHAIN=auto GOMAXPROCS=2 \
go test -count=1 -p=1 ./pkg/containerwatcher/v2/tracers \
-run '^TestExecFields$'

GOTOOLCHAIN=auto GOMAXPROCS=2 \
go test -count=1 -p=1 \
./pkg/utils \
./pkg/rulemanager/cel \
./pkg/rulemanager/types/v1

GOTOOLCHAIN=auto GOMAXPROCS=2 \
go test -race -count=1 -p=1 \
./pkg/utils \
./pkg/rulemanager/cel \
./pkg/rulemanager/types/v1

GOOS=linux GOARCH=arm64 CGO_ENABLED=0 \
go build -o /tmp/node-agent-975-arm64 ./cmd/main.go

make docker-build \
  IMAGE=quay.io/kubescape/node-agent \
  TAG=test-975
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved exec TTY detection, including `/dev/pts/0` and other pseudo-terminals.
  * TTY-related alerts now correctly distinguish events with and without terminal information.

* **Enhancements**
  * Updated exec monitoring to include creation-time and TTY detail fields.
  * Updated the bundled exec gadget to version 0.55.0 for improved metadata reporting.

* **Documentation**
  * Updated TTY behavior and validation guidance to reflect the latest exec monitoring capabilities.


pkg/containerwatcher/v2/tracers/exec.go
```diff
@@ -18,7 +18,7 @@ import (
 )
 
 const (
-	execImageName = "ghcr.io/inspektor-gadget/gadget/trace_exec:v0.48.1"
+	execImageName = "ghcr.io/inspektor-gadget/gadget/trace_exec:v0.55.0"
```
matthyx, line 21: **Blocker.** `trace_exec` v0.55.0 also includes upstream `06323f6d`/`c759b4e1`: `args[0]` changes from the `execve` pathname (`/bin/sh`) to the committed `argv[0]` (`sh`). node-agent records the full `args` vector in profiles and R0040 (`cp.was_executed_with_args`) compares it verbatim, so existing profiles stop matching — `Test_32_UnexpectedProcessArguments` already fails on this PR (job 105671995732) while passing on other current PRs. `GetExecPathFromEvent`/malware-manager `/proc/<pid>/root/<args[0]>` are affected too. Suggest vendoring `trace_exec` at v0.48.1 + only the tty commit (`c6fc831a`) via `BUILT_GADGETS` like `trace_open`, or explicitly restoring/handling the `args[0]` convention and its profile migration before bumping.

matthyx (changes requested):
## Review: request changes

Reviewed head `489d5e26a5fa81826b3f58afa2c8b3fd678726a5` against `main` (`5acac56a`).

**The TTY fix itself works** — `Test_35_ExecTTYFieldTest` passes on CI with the new gadget, and I confirmed locally that `ig` 0.48.1 can pull/export `trace_exec:v0.55.0` and that `TestExecFields` passes against the pinned IG runtime (`kubescape/inspektor-gadget@06b0d12`, based on upstream ~v0.48.0). Bumping the gadget is the right way to close #975.

**Blocker: the bump silently changes `args` semantics and breaks R0040 / existing profiles.**

Between v0.48.1 and v0.55.0, `trace_exec` also picked up inspektor-gadget commits `06323f6d` ("read args from the new process address space"), `c759b4e1` ("report argv[0] instead of the exec path for failed execs") and `03c0f590` (lazy arg parsing). Net effect: `args[0]` used to be the `pathname` passed to `execve` (e.g. `/bin/sh`); it is now the committed `argv[0]` (e.g. `sh`). For anything launched through a shell or via PATH lookup those differ.

node-agent stores the full `args` vector (including `args[0]`) in container profiles and compares it verbatim in `cp.was_executed_with_args` (rule R0040), and `utils.GetExecPathFromEvent` returns `args[0]`. Consequences:

- **CI already shows it:** `component-tests (Test_32_UnexpectedProcessArguments)` **fails** on this PR (run 35366394715, job 105671995732), while it passes on other recent PRs (e.g. runs 35364356926, 34937021951). Example: `sh -c "echo hi"` → runtime argv `[sh, -c, echo hi]` no longer matches profile `[/bin/sh, -c, ⋯⋯]`, so R0040 fires ("must stay silent", expected 2 got 3). The workflow is green only because `component-tests` has `continue-on-error: true`.
- **Fleet impact:** every already-learned/authored profile encodes the old `args[0]` form, so after upgrade R0040 would alert on shell-launched execs across the board, and freshly learned profiles would be in a different format than existing ones (mixed-version fleets get both).
- `pkg/malwaremanager/v1` builds `/proc/<pid>/root/<args[0]>` via `GetHostFilePathFromEvent`; with a relative `argv[0]` that path no longer resolves. Alert `CommandLine`/`exec` fields in `ruleadapters/adapters/exec.go` change shape too.

Suggested remedies (either is fine; the first is the smaller blast radius):
1. Keep the gadget at v0.48.1 and vendor `trace_exec` under `pkg/ebpf/gadgets/` with only upstream `c6fc831a` (the `tty_major`/`tty_minor` change) applied, built via the existing `BUILT_GADGETS` mechanism exactly like `trace_open` already is.
2. Stay on v0.55.0 but make the args change explicit: restore path-style `args[0]` (e.g. from `exepath`) in `execoperator.go` or adjust profile recording/matching to be robust to both forms, update `Test_32`/fixtures, and document the profile-compatibility implication. This needs a real migration story for existing profiles, so it's more than a follow-up.

Non-blocking:
- `docs/features/exec-tty-field.md`: "the regression fixed by #975" — #975 is the issue; this PR is the fix.
- `Makefile`: the trailing `sudo ig image pull` for `trace_exec` is fine, but consider folding it into the existing `$(foreach …)` with a per-gadget version map if more gadgets end up pinned separately.

Validation performed: `go build ./...` OK; `go test ./pkg/utils ./pkg/rulemanager/cel/... ./pkg/rulemanager/types/v1 ./pkg/containerprofilemanager/...` OK; `TestExecFields` OK against a locally exported `trace_exec:v0.55.0`; CI Test_35 pass / Test_32 fail as above. I did not run the component suite locally (needs kind).

Search notes: related history is #868 (introduced the two-phase TTY fields and Test_35), #695 (IG v0.48.1 upgrade), #800 (exec path recording symmetry — relevant to the `args[0]` convention). No prior PR attempts a `trace_exec` bump; nothing superseding this one.

ANAMASGARD:
Thanks @matthyx for catching the argument compatibility change. `Test_32_UnexpectedProcessArguments` was a real regression: the v0.55.0 gadget changed the pathname-based `args[0]` expected by existing profiles.

Pushed `00e1429ed96533603946505ae1ccb5b746e25c99` with the smaller remedy: vendored v0.48.1 `trace_exec` plus only the `c6fc831a` TTY device-number backport, built through `BUILT_GADGETS` and packaged/loaded as `v0.48.1-tty`. The strict schema now expects only `tty_major` and `tty_minor` beyond v0.48.1, and the docs refer to #975 as the reported issue.

Local Kind validation against the rebuilt image and storage v0.0.342 passed:

- `Test_32_UnexpectedProcessArguments`: PASS, all ten subtests (552.10s).
- `Test_35_ExecTTYFieldTest`: PASS (88.95s). R9901: no-TTY=0, pts/0=1, nonzero PTY=1; R9903=3; R9904=0.
- Gadget build, strict `TestExecFields`, focused Go tests, race tests, Linux ARM64 build, Docker build, and diff checks also passed.

The empty-image SBOM debug messages still occur with the same storage image while these tests pass. Storage continues saving those profiles, so that diagnostic is separate from this compatibility fix. I also removed `continue-on-error` from the component job and changed the memlock step to raise the actual test shell's limits. Fresh CI is pending; I have not requested re-review yet.

PR #984, opened 2026-09-23
fix: support host sensor exclusions for kube-proxy-free clusters

### LINKED ISSUE #962: Host sensor warns continuously when kube-proxy is intentionally absent
### What happened?

On a Kubernetes cluster that deliberately runs without kube-proxy (Cilium kube-proxy replacement), enabling the node-agent host sensor emits this warning on every host-sensor interval:

```text
sensor failed kind=KubeProxyInfo error="failed to sense data: failed to locate kube-proxy process: ..."
```

This is an expected cluster state rather than a sensor failure. The warning creates recurring operational noise and makes warning-based health views less actionable.

### Why it happens

`NewHostSensorManager` unconditionally includes `NewKubeProxyInfoSensor` whenever host sensing is enabled. `KubeProxyInfoSensor.Sense` returns an error when it cannot locate the kube-proxy process. The same behavior is present in v0.3.142 and current `main`.

There is currently one switch for the complete host-sensor manager, so an operator cannot suppress only this inapplicable sensor without also losing OS, kernel, kubelet, open-port, hardening, and CNI data.

### Requested behavior

Please support kube-proxy-free clusters without a recurring warning. Either of these contracts would work:

1. Treat an absent kube-proxy process as a supported empty/not-applicable result; or
2. Add a per-sensor exclusion list so operators can disable `KubeProxyInfo` while retaining the other host sensors.

The first option seems preferable when kube-proxy replacement can be detected reliably. A test should cover a host with no kube-proxy process and prove the periodic sensing loop remains warning-free while the other sensors still run.

### Versions checked

- node-agent v0.3.142
- current `main` as of 2026-09-08

ANAMASGARD:
## Overview

Host sensing currently runs `KubeProxyInfo` unconditionally, producing recurring warnings on clusters that intentionally replace kube-proxy.

This PR adds `hostSensorExcludedSensors`, allowing operators to exclude `KubeProxyInfo` while retaining other host sensors.

- Keep all sensors enabled by default.
- Validate exact, case-sensitive sensor names before Kubernetes client initialization.
- Preserve sensor order, tolerate duplicates, and allow all sensors to be excluded.
- Document JSON, environment-variable, and Helm configuration.
- Add regression tests covering exclusions, repeated sensing, genuine failure reporting, and kube-proxy process discovery.

## Additional Information

Configure kube-proxy-free clusters with:

```json
{
  "hostSensorEnabled": true,
  "hostSensorExcludedSensors": ["KubeProxyInfo"]
}
```

Restart node-agent after changing exclusions. Existing host-data objects remain untouched and may contain stale data or error status.

Without an explicit exclusion, missing kube-proxy processes continue to produce warnings. The Helm example requires a chart version supporting `nodeAgent.config.extra`.

## How to Test

Using Go 1.27, run:

```bash
go test -mod=readonly -count=1 ./pkg/hostsensormanager ./pkg/config ./cmd
go test -mod=readonly -race -count=1 ./pkg/hostsensormanager ./pkg/config
go vet -mod=readonly ./pkg/hostsensormanager ./pkg/config ./cmd
go build -mod=readonly -o /tmp/node-agent-962 ./cmd
```

These focused checks passed locally. Helm rendering was also verified. The repeated-sensing regression test fails when exclusion filtering is bypassed and passes with filtering enabled.

The broader suite could not pass completely because tracer and validator tests require unavailable eBPF/memlock privileges and a gadget dependency. Live Cilium-cluster verification has not been performed.

## Related issues/PRs

- Resolved #962

Focused tests pass; the full-suite checkbox remains unchecked because of the environment limitations described above.
```

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* **New Features**
  * Host sensors can now be excluded by name through the `hostSensorExcludedSensors` configuration setting. Excluded sensors are skipped while other enabled sensors continue to run.
  * On kube-proxy-free clusters, excluding `KubeProxyInfo` can prevent recurring kube-proxy warnings while keeping other host sensors enabled.
* **Documentation**
  * Added configuration guidance covering sensor exclusions, supported configuration methods, validation, and restart requirements.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->


matthyx (changes requested):
Request changes

This is a needed, accepted solution for #962: the per-sensor filter preserves registration order, supports an empty resulting set, and leaves the default behavior unchanged. The reviewed head is 06039f349fe3dbcac146f9460e47a97af5be96a7 against main@43e0f1397c8da4a65e4cf098c15495f204716524.

Blocking findings:
- P2 — `cmd/main.go:107-250` validates exclusions only when `NewHostSensorManager` is constructed, after `NewKubernetesApi()` and `GetClusterUID()` have already run. A typo therefore performs Kubernetes startup work (and can produce its own failure/noise) before returning the actionable configuration error, contrary to the PR’s stated validation-before-client-initialization contract. Extract/reuse the validation and invoke it immediately after config loading when host sensing is enabled; retain constructor validation for other callers. This is also covered by the outstanding CodeRabbit inline thread.
- P2 — `docs/CONFIGURATION.md:192` says the environment variable overrides JSON, but an empty `HOSTSENSOREXCLUDEDSENSORS` is treated by Viper as unset, so it does not clear a JSON exclusion list. Document the behavior (and ideally add the file-plus-empty-env regression case) or explicitly support clearing through the environment. This is the other outstanding CodeRabbit thread.

Validation: `go test -mod=readonly -count=1 ./pkg/hostsensormanager ./pkg/config ./cmd`; `go test -race -mod=readonly -count=1 ./pkg/hostsensormanager ./pkg/config`; `go vet -mod=readonly ./pkg/hostsensormanager ./pkg/config ./cmd`; and `go build -mod=readonly -o /tmp/node-agent-984 ./cmd` all passed. I also checked `git diff --check`. CI currently has failures in the broad cross-platform job and component Test_16; the latter is an alert-timing failure in unrelated component-test code, and neither was reproduced by the focused checks.

History: #962 explicitly accepts an exclusion-list solution; no duplicate/superseding PR was found. PR #681 contains only bot advice favoring an empty result, not a maintainer rejection of exclusions. Recheck and rerun affected checks after updating the head.

matthyx (changes requested):
Recheck of 925dd3c found one new blocker.

P1 — `pkg/hostsensormanager/sensor_kubeproxy_test.go:18-25` does not compile in the PR merge result. The target branch defines `hostFSPrefix` as `atomic.Pointer[string]`, while this new test assigns a string and passes the pointer directly to `filepath.Join`. The required cross-platform CI job fails with these exact type errors. Please update the test to use the target branch’s `Store`/`Load` API (and restore the prior pointer value safely), then rerun CI.

The prior two requested changes are correctly addressed: validation now runs immediately after config loading and is covered by an entrypoint test; the empty-environment behavior is documented and regression-tested. Local focused checks passed on the PR head, but they do not reproduce the merged target state. I cannot approve while the merge-result build and component CI remain failing.

Issue #963, opened 2026-09-08
Garbage-collect host-data resources after Kubernetes Nodes are deleted


devantler:
### What happened?

Host-data custom resources created by the node agent remain after their Kubernetes Node is deleted. On clusters with node autoscaling, each scale-up creates another set of cluster-scoped resources and the stale objects accumulate indefinitely.

### Why it happens

`CRDClient.CreateOrUpdateHostData` names each resource after the node but sets no owner reference. The host-sensor manager has no node-deletion watcher or shutdown cleanup, so nothing removes resources when a Node disappears.

This affects the resources in `hostdata.kubescape.cloud`, including OS release, kernel, kubelet, CNI, open-port, security-hardening, control-plane, cloud-provider, and kernel-variable data.

### Requested behavior

Please garbage-collect host-data resources whose named Kubernetes Node no longer exists. Cleanup should fail closed when the Node list cannot be read or is empty, and should not remove data for a present Node.

A controller-side Node deletion watcher is likely more reliable than pod shutdown cleanup because an autoscaled node can disappear without giving its node-agent pod a clean shutdown window. Tests should cover node deletion, transient Node-list failure, and a present node.

### Versions checked

- node-agent v0.3.142
- current `main` as of 2026-09-08

PR #985, opened 2026-09-23
fix: garbage collect orphaned host data with leader election

ANAMASGARD:
- Addresses #963: one Lease-elected Node Agent collects orphaned host data at leadership acquisition and each host-sensor interval, while ordinary sensing remains per-node.
- Require complete, nonempty Node inventories; recheck each missing Node and delete only with UID/resourceVersion preconditions. Cancel and join cleanup on leadership loss/shutdown; API or election failures do not stop sensing.
- Deployment prerequisite: companion RBAC change https://github.com/kubescape/helm-charts/pull/936 must be installed before or alongside the updated agent. Election coordinates one installation namespace; conditional deletion remains necessary because Lease election is not strict fencing.
- Validation: focused race tests, vet, command builds, and real Kind integration with production CRDs/RBAC passed. Integration covers follower inactivity, takeover, all ten resource kinds, live-node preservation, API-server preconditions, and denied-Lease behavior.
- Broad Go suite: 64 packages passed; two packages were environment-blocked by eBPF/memlock permissions and a missing `trace_exec:v0.48.1` gadget image.


matthyx (commented):
## Needs clarification

The implementation is needed: #963 documents cluster-scoped host-data records surviving Node deletion. I found no source-level correctness blocker in the complete-inventory, recheck, UID/resourceVersion-precondition, and leadership-cancellation paths.

### Release/deployment gate

`NewHostSensorManager` starts the collector whenever host sensors have a namespace (`pkg/hostsensormanager/manager.go:46-60`), and the election loop retries each term (`pkg/hostsensormanager/garbage_collector_leader.go:43-53`). Existing deployments do not receive the required Node/host-data delete and Lease RBAC from this PR; those permissions are only in the still-open, review-required companion [helm-charts#936](https://github.com/kubescape/helm-charts/pull/936). Until that PR is merged/released and the rollout ordering is confirmed, upgraded agents will repeatedly fail Lease acquisition and the requested cleanup will never run.

Please confirm the merge/release coordination (or land the companion RBAC change first) before approval. This is a deployment blocker, not a request to weaken the fail-closed behavior.

Validation: `go test -mod=readonly ./pkg/hostsensormanager -run '^(TestGC|TestCollectHostData|TestListGCObjects)' -count=1` passed (12.041s) in an isolated environment; `git diff --check` passed. The PR has no attached CI checks. I did not run the Kind integration script because it creates a cluster and applies RBAC; static inspection found it targets a new disposable cluster.

PR #986, opened 2026-09-23
feat: reload namespace filters without restarting node-agent

ANAMASGARD:
## Overview

- Add opt-in `namespaceFilterFile` support so namespace include/exclude lists can change without restarting node-agent. Poll the mounted JSON file every five seconds after Kubernetes propagates the ConfigMap update; invalid updates retain the last valid settings.
- Reconcile running containers, preserve cleanup and readmission ordering, and keep metadata/rule bindings available for newly included namespaces. Apply the active filter to node status reports as well.
- Preserve static filtering by default, include-over-exclude precedence, host monitoring, and pod-label exclusions. Document the file format and directory-volume mount requirement.

Partially addresses https://github.com/kubescape/helm-charts/issues/664. This PR only implements the node-agent portion; Helm integration and changes to other components remain separate.

## How to Test

- Passed package tests across 10 relevant packages, targeted namespace-filter/lifecycle race tests, `go vet` for changed packages and `cmd`, the command build, and `git diff --check`.
- Tests cover projected-volume symlink replacement, config copies, invalid updates, discovery retries, running-container transitions, rapid profile readmission, metadata initialization, and reporting.
- Live Kubernetes/eBPF validation has not been run. For cluster validation, mount the documented ConfigMap without `subPath`, set `namespaceFilterFile`, and toggle a running workload's namespace in the file. Verify monitoring stops/resumes while node-agent and workload pods remain running.


matthyx (changes requested):
Thanks for the thorough opt-in design and coverage. I found three blockers before this can merge:\n\n1. **High — a stuck CRI status request freezes future filter changes and shutdown.**  performs  synchronously on its only polling goroutine (), and discovery calls  without a context/deadline (). The pinned runtime path reaches , so a connected-but-unresponsive runtime can block forever. Later exclusions are never published, and  waits forever for the reload goroutine (). Please make discovery cancellable/bounded and decouple publication/exclusion processing from slow admission; add a blocked-runtime test that verifies a later exclusion and shutdown still complete.\n\n2. **High — cache cleanup serializes all containers behind one potentially 10-minute wait.** The new timer holds  while calling  (). That can wait on a per-container add path doing the documented 10-minute shared-data wait. While it is held, every unrelated add/remove callback blocks on the same global mutex, including static-mode workloads. Decide the pending generation under the lock, release it, then perform deletion (with per-container synchronization if required). Add a concurrent unrelated-container regression test.\n\n3. **Medium — late admission incorrectly records a full profile.** A container that began after node-agent startup while excluded, then is admitted after a filter edit (), receives  only from start time () and is marked  () despite its earlier events never being monitored. Carry late-admission state into profile creation or explicitly preserve partial semantics, with a regression test.\n\nPublished CI is green, but it does not exercise these blocked-I/O and cross-container paths.

matthyx:
Supplementing my requested-changes review with the exact references (the first review lost inline-code formatting during submission):

1. **High — an unresponsive CRI request freezes further updates and shutdown.** `pkg/containerwatcher/v2/namespace_filter.go:33-49` runs `reconcileNamespaceFilter` synchronously in its sole polling goroutine. `runningContainersForNamespace` invokes `igK8sClient.GetRunningContainers` at `:90-107` without a context/deadline; the pinned runtime path reaches `ContainerStatus(context.Background())`. A connected-but-unresponsive runtime therefore prevents later exclusions from being read and makes `Stop()` wait indefinitely at `pkg/containerwatcher/v2/container_watcher.go:408-410`. Make the runtime-discovery boundary cancellable/bounded, and let publication/exclusion processing continue while slow admission is pending. Add a blocked-runtime regression test covering a later exclusion and shutdown.

2. **High — one stalled profile-cache registration blocks all containers' lifecycle callbacks.** The timer holds global `removalMu` across `deleteContainer` in `pkg/objectcache/containerprofilecache/containerprofilecache.go:265-271`. That deletion can wait on a per-container registration that waits up to ten minutes for metadata. Until then, unrelated additions/removals block on `removalMu`, including in static mode. Keep only generation validation under the global lock; perform deletion outside it with per-container synchronization. Add a two-container regression test.

3. **Medium — a container admitted after being excluded is labelled as fully profiled.** The reconciliation path adds it at `pkg/containerwatcher/v2/namespace_filter.go:80`; `PreRunningContainer` is based only on whether it predates agent startup at `pkg/containerwatcher/v2/containercallback.go:244-245`, and the profile manager consequently marks it full at `pkg/containerprofilemanager/v1/lifecycle.go:447-450`. It missed all activity before admission. Carry late-admission state through shared metadata and preserve partial semantics, with coverage for a container started while excluded.

matthyx (changes requested):
Rechecked current head `5fbea7a1f7710c3a0fd1263eea1c3d5e513e0ef9`.

The prior runtime-discovery, cache-lock, and late-admission blockers are addressed. However, the existing CodeRabbit finding is confirmed: `pkg/containerprofilemanager/v1/lifecycle.go:385-394` changes an actively monitored excluded container to `Completed` without changing its existing `Full` completion status. A container excluded during learning is therefore persisted and consumed as a final profile despite missing the remainder of its session.

Please set the completion status to `Partial` for an active, non-host container removed because of namespace exclusion, before termination is signalled/saved; retain host and already-completed behavior. Add coverage for ordinary admission, exclusion during learning, and persisted `Completed + Partial`.

